PR #21, opened 2017-06-09
syntax error corrected

prateek-patel:
there is a syntax error in the code after "Also we can use the method Object.assign for that." line.
The comma is added after the square bracket so which is wron. I have corrected in it.


iliakan:
Usually, a square bracket encloses optional arguments. Here the first comma is optional too.

PR #22, opened 2017-06-09
Edit article.md

aadithpm:
Added an explanation with one of the errors that strict mode catches


iliakan:
There are many typos in your text and it's out of the flow now.
I mean: Summary is always the last, it summarizes. Please read the chapter and try to fit the explanations better, if they are needed.

aadithpm:
Hey, thanks for your response. Will work on it.

PR #28, opened 2017-06-10
Edit article.md

aadithpm:
Extrapolated the "reasons for 'use strict'" list with another point. If this is okay, I'll add the corresponding notes with regard to strict mode in the respective sections for variables and object attributes. However, I understand that sidenote after sidenote for sections would just confuse the readers. I completely understand if you feel that this extra information is not needed at this point.


iliakan:
Added that info, thank you!

PR #31, opened 2017-06-10
Edit article.md. Small section on truthy and falsy

aadithpm:
The way that truthy and falsy works, especially how it evaluates against all falsy values, is a bit important so I thought it could be added to this section. Pretty much most of us turn to using truthy and falsy in conditional statements in place of traditional checks, and I felt that illustrating on it a bit more would be informative.

### DIFF
--- a/1-js/02-first-steps/10-ifelse/article.md
+++ b/1-js/02-first-steps/10-ifelse/article.md
@@ -67,6 +67,63 @@ if (cond) {
   ...
 }
 ```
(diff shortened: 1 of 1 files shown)


iliakan:
Added some on that topic into the article.
The text you write cannot be accepted, because:
- Uses `var` (old JS)
- Badly formatted code.
- Please read what you write, make it easier to read and understand :)
Thank you!

PR #32, opened 2017-06-10
Code typo in prompt example

aadithpm:
It initially had (title[, default]), but I'm not sure why you had the array notation there since default is usually a string. I decided to leave it on there since I'm sure you had your reasons. However, I think it would be more appropriate without the array notation.


iliakan:
square brackets mean "optional"

PR #49, opened 2017-06-18
Edit article.md

aadithpm:
Added a way to generate random numbers within a range with Math.random()


aadithpm:
Never mind, looked at the practise problems. Proposed change is redundant.

PR #66, opened 2017-06-28
Update task.md

brajabi:


iliakan:
Here we expect a number greater than zero :)

PR #108, opened 2017-07-22
Fix typo on a figure image

ajhsu:
Hi @iliakan,
I have to say, this is the best JavaScript tutorial I ever seen.
And I just found a trivial typo from a figure image in "Closure" section.


iliakan:
Thank you, applied the fix locally.

PR #139, opened 2017-08-15
Update article.md

herambgadekar:
falsy updated to false

### DIFF
--- a/1-js/02-first-steps/08-comparison/article.md
+++ b/1-js/02-first-steps/08-comparison/article.md
@@ -174,7 +174,7 @@ alert( null == 0 ); // (2) false
 alert( null >= 0 ); // (3) *!*true*/!*
 ```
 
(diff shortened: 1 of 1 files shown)


iliakan:
@jmbothe what you think, is that a correct change to do?

jmbothe:
@iliakan @herambgadekar 

So this is a little bit outside of general English usage, because "falsy" is technical vocabulary from the field of logic. Please correct me if I am wrong, but "falsy" is usually used to refer to a value BEFORE it is evaluated. For example, in JS we say that '' , 0 , null, undefined, etc. are 'falsy' because LATER they will evaluate to false. But we say "false" if we are talking about a strictly Boolean value, or the result of an evaluation that has already occurred, right? I am not entirely sure, as I am fairly new to JS, but this is the impression I get. I could very well be wrong.

Anyway, if I am right, then this change is correct because you are talking about the value of an expression after it has been evaluated.

Or, if you wanted to avoid any confusion, you could say "...but they both evaluate to false", or something like that.

PR #141, opened 2017-08-18
Safari does not return an empty string when a prompt is cancelled

nhooyr:
title


nhooyr:
At least not anymore.

PR #148, opened 2017-08-27
Minor text edits

gwooly:


1-js/05-data-types/02-number/article.md
```diff
@@ -1,20 +1,20 @@
...
 
-Imagine, we need to write a billion. The obvious way is:
+Imagine we need to write 1 billion. The obvious way is:
 
 ```js
 let billion = 1000000000;
 ```
 
-But in real life we usually dislike writing many zeroes. It's easy to mistype. Also we are lazy. We will usually write something like `"1bn"` for a billion or `"7.3bn"` for 7 billions 300 millions. The similar is true for other big numbers.
+But in real life we usually avoid writing the full number as it's easy to mistype. Also, we are lazy. We will usually write something like `"1bn"` for a billion or `"7.3bn"` for 7 billion 300 million. The same is true for most large numbers.
```
iliakan: The idea is that it's easy to type wrong many **zeroes**, not just full numbers. What's wrong about explicitly saying it here?

iliakan: P.S. Maybe the last sentence here should be removed?

1-js/05-data-types/02-number/article.md
```diff
@@ -125,7 +125,7 @@ There are following built-in functions for rounding:
 : Rounds to the nearest integer: `3.1` becomes `3`, `3.6` becomes `4` and `-1.1` becomes `-1`.
 
 `Math.trunc` (not supported by Internet Explorer)
-: Removes the decimal part: `3.1` becomes `3`, `-1.1` becomes `-1`.
+: Removes anything after the decimal without rounding: `3.1` becomes `3`, `-1.1` becomes `-1`.
```
iliakan: "Removes anything after the decimal" - are you sure about it? Maybe you mean "after the decimal point"? 

Are you sure the original text is wrong?

1-js/05-data-types/02-number/article.md
```diff
@@ -137,15 +137,15 @@ Here's the table to summarize the differences between them:
...
 
 For instance, we have `1.2345` and want to round it to 2 digits, getting only `1.23`.
 
-There are two ways to do so.
+There are two ways to do so:
 
 1. Multiply-and-divide.
 
-    For instance, to round the number to the 2nd digit after the point, we can multiply the number by `100`, call the rounding function and then divide back.
+    For example, to round the number to the 2nd digit after the decimal, we can multiply the number by `100`, call the rounding function, and then divide back.
```
iliakan: Same question: "just decimal", not "decimal point"?

Also, is the comma after "function" really required?

1-js/05-data-types/02-number/article.md
```diff
@@ -159,14 +159,14 @@ There are two ways to do so.
     alert( num.toFixed(1) ); // "12.3"
     ```
 
-    The rounding goes to the nearest value, similar to `Math.round`:
+    This round up or down to the nearest value, similar to `Math.round`:
```
iliakan: Maybe "This rounds"? The original wording is wrong?

1-js/05-data-types/02-number/article.md
```diff
@@ -203,24 +203,24 @@ alert( 0.1 + 0.2 ); // 0.30000000000000004
 
 Ouch! There are more consequences than an incorrect comparison here. Imagine you're making an e-shopping site and the visitor puts `$0.10` and `$0.20` goods into his chart. The order total will be `$0.30000000000000004`. That would surprise anyone.
 
-Why does it work like that?
+But why does this happen?
 
-A number is stored in memory in it's binary form, as a sequence of ones and zeroes. But fractions like `0.1`, `0.2` that look simple in the decimal numeric system are actually unending fractions in their binary form.
+A number is stored in memory in it's binary form, a sequence of ones and zeroes. But fractions like `0.1`, `0.2` that look simple in the decimal numeric system are actually unending fractions in their binary form.
```
iliakan: Maybe "it's" should be replaced by "its"? Or just "the binary form"?

1-js/05-data-types/02-number/article.md
```diff
@@ -239,22 +239,22 @@ Can we work around the problem? Sure, there's a number of ways:
...
 
     ```js run
     alert( (0.1*10 + 0.2*10) / 10 ); // 0.3
     ```
 
-    It works, because when we get `0.1*10 = 1` and `0.2 * 10 = 2` then both numbers are integers, there's no precision loss for them. 
+    This works because when we do `0.1*10 = 1` and `0.2 * 10 = 2` then both numbers become integers, and there's no precision loss. 
 
-3. If it's a shop, then the most radical solution would be to store all prices in cents. No fractions at all. But what if we apply a discount of 30%? In practice, totally evading fractions is rarely feasible, so the solutions listed above are here to help.
+3. If it's we were dealing with a shop, then the most radical solution would be to store all prices in cents and use no fractions at all. But what if we apply a discount of 30%? In practice, totally evading fractions is rarely feasible, so the solutions above help avoid this pitfall.
```
iliakan: "If it's we were dealing with..." -- can't understand this.

1-js/05-data-types/02-number/article.md
```diff
@@ -320,7 +320,7 @@ let num = +prompt("Enter a number", '');
 alert( isFinite(num) );
 ```
 
-Please note that an empty or a space-only string is treated as `0` in all numeric functions including `isFinite`. 
+Please note that spaces in a string are treated as a `0` in all numeric functions, including `isFinite`.
```
iliakan: That's wrong: the spaces are not treated as `0`.

An empty string is `0`.
A space-only string is `0`.

But spaces are generally not `0`, like here: "1 2" -- invalid number (not 102).

1-js/05-data-types/02-number/article.md
```diff
@@ -331,35 +331,35 @@ There is a special built-in method [Object.is](mdn:js/Object/is) that compares v
 
 In all other cases, `Object.is(a, b)` is the same as `a === b`. 
 
-This way of comparison is often used in JavaScript specification. When an internal algorithm needs to compare two values for being exactly the same, it uses `Object.is` (internally called [SameValue](https://tc39.github.io/ecma262/#sec-samevalue)).
+This method of comparison is often used in JavaScript. When an internal algorithm needs to compare two values for being exactly the same, it uses `Object.is` (internally called [SameValue](https://tc39.github.io/ecma262/#sec-samevalue)).
```
iliakan: Not in Javascript, but in Javascript spec.

There are many internal things used in the specification of the language, but rarely in programming on that language.

It's one of them.

1-js/05-data-types/02-number/article.md
```diff
@@ -331,35 +331,35 @@ There is a special built-in method [Object.is](mdn:js/Object/is) that compares v
...
 
-The numeric conversion using a plus `+` or `Number()` is strict. If a value is not exactly a number, it fails:
+Numeric conversion using a plus `+` or `Number()` is strict. If a value is not exactly a number, it fails:
 
 ```js run
 alert( +"100px" ); // NaN
 ```
 
-The sole exception is spaces before and after the line, they are ignored.
+The sole exception is spaces before and after the line, as they are ignored.
```
iliakan: Maybe: "The sole exception is spaces at the beginning or at the end of the string..." ?

1-js/05-data-types/02-number/article.md
```diff
@@ -331,35 +331,35 @@ There is a special built-in method [Object.is](mdn:js/Object/is) that compares v
...
 
-But in real life we often have values in units, like `"100px"` or `"12pt"` in CSS. Also in many countries the currency symbol goes after the amount, so we have `"19€"` and would like to extract a numeric value out of that.
+But in real life we often have values in units, like `"100px"` or `"12pt"` in CSS. Also in many countries, the currency symbol goes after the amount, so we have `"19€"`, but would still like to extract a numeric value out of that.
```
iliakan: Why "but"? That's "and"...

1-js/05-data-types/02-number/article.md
```diff
@@ -331,35 +331,35 @@ There is a special built-in method [Object.is](mdn:js/Object/is) that compares v
...
 
 That's what `parseInt` and `parseFloat` are for.
 
-They "read" a number from a string until they can. In case of an error, the gathered number is returned. Function `parseInt` reads an integer number, `parseFloat` reads any number:
+They "read" numbers within a string and unless there's an error (for example, if the first character of a string is not a number), the first valid number is returned. The function `parseInt` returns an integer, whilst `parseFloat` will return a floating-point number:
```
iliakan: They read only *one* "number", not "numbers". The idea is that they aread from the start to the first error and then return what they've got.
Is there a better way to say that?

1-js/05-data-types/02-number/article.md
```diff
@@ -331,35 +331,35 @@ There is a special built-in method [Object.is](mdn:js/Object/is) that compares v
...
 
 ```js run
 alert( parseInt('100px') ); // 100
 alert( parseFloat('12.5em') ); // 12.5
 
-alert( parseInt('12.3') ); // 12, only integer part
+alert( parseInt('12.3') ); // 12, only the integer is returned
```
iliakan: I mean "the integer part" as defined here: https://oeis.org/wiki/Integer_part

1-js/05-data-types/02-number/article.md
```diff
@@ -331,35 +331,35 @@ There is a special built-in method [Object.is](mdn:js/Object/is) that compares v
...
 alert( parseFloat('12.3.4') ); // 12.3, the second point stops the reading
 ```
 
-Of course, there are situations when `parseInt/parseFloat` return `NaN`. It happens when no digits could be read:
+There are situations when `parseInt/parseFloat` will return `NaN`. It happens when no digits could be read:
```
iliakan: Why "will" ?

1-js/05-data-types/02-number/article.md
```diff
@@ -423,12 +423,12 @@ For different numeral systems:
 
 For converting values like `12pt` and `100px` to a number:
 
-- Use `parseInt/parseFloat` for the "soft" conversion, which reads a number from a string until it can. 
+- Use `parseInt/parseFloat` for the "soft" conversion, which reads a number from a string until broken by something that is not a number.
```
iliakan: "...and then return the value they could read before the error?"

gwooly:
I'll wait to see comments from @jmbothe - if this is not accurate then apologies - don't want to create more work for you!

iliakan:
@gwooly thank you. Please note that you create work not for me, but for many people across the world.

I'm merely trying to make sure that PRs are good. That's very hard if you remember that English is not my native.

gwooly:
True. This is a great resource so I'd still like to contribute - it might be better if I stick to the grammar/language rather than to modify references to the JS, since it's vital those are 100% accurate (and I am no expert in JS, which is why I also use this guide!).

1-js/05-data-types/02-number/article.md
jmbothe: @iliakan  "writing many zeroes" is a somewhat awkward phrase, but you are right that it is more explicit than "the full number". In @gwooly edit, the reader must infer that "the full number" refers to the billion variable value, but to me it makes sense. alternate suggestion "we usually avoid writing a long string of zeroes as it's easy to mistype".

1-js/05-data-types/02-number/article.md
jmbothe: I had to look this one up, but I have a feeling @iliakan already knew :) The word '**decimal**' refers to **either** any numeral in the decimal system (aka base 10 system) **OR** any digit following the **decimal separator**. '**Decimal point**' refers to the separator itself. So, yea @gwooly should have said "anything after the decimal point'...

1-js/05-data-types/02-number/article.md
jmbothe: see above for the decimal point question. as for the comma, this is a long standing point of contention in English grammar. Google "Oxford comma" for more info. I think the main goal is just to remain consistent. I noticed early on that the articles consistently DO NOT use the Oxford comma, so I avoided added them.

1-js/05-data-types/02-number/article.md
jmbothe: should be "This rounds..."

1-js/05-data-types/02-number/article.md
jmbothe: Yes, "it's" should be changed to "its". Removing "as" is appropriate.

1-js/05-data-types/02-number/article.md
jmbothe: Looks like a typo. Maybe @gwooly meant to remove "it's"? That would make sense.

1-js/05-data-types/02-number/article.md
jmbothe: Agree with @iliakan . Original content is more accurate.

1-js/05-data-types/02-number/article.md
jmbothe: Agree with @iliakan . Original content is more accurate.

1-js/05-data-types/02-number/article.md
jmbothe: @iliakan that would definitely be more specific

1-js/05-data-types/02-number/article.md
jmbothe: Original and edited content seem equivalent to me. I understand both well.

1-js/05-data-types/02-number/article.md
jmbothe: Hahaha. Yea, this one is difficult to describe. I'll give it a shot, but I'm not sure if it will be adequate:

"They 'read' a string character by character, beginning with index 0. When they encounter the first non-numeral character, they stop reading and return the beginning portion of the string: the parsed number. If the first character is not a number, they return NaN."

That's a good start anyway :)

1-js/05-data-types/02-number/article.md
jmbothe: agree with @iliakan

1-js/05-data-types/02-number/article.md
jmbothe: Both of these make sense to me. They seem equivalent.

1-js/05-data-types/02-number/article.md
jmbothe: see my comments above

1-js/05-data-types/02-number/article.md
iliakan: Well, I did not know about the Oxford comma, now I do know. So funny different languages have different comma rules for precisely the same syntax constructs.

1-js/05-data-types/02-number/article.md
iliakan: And the original wording is wrong, @jmbothe ?

1-js/05-data-types/02-number/article.md
iliakan: @gwooly could you leave the original content here please? :) Looks like it's not wrong =)

1-js/05-data-types/02-number/article.md
iliakan: Uhm yes, but `parseFloat` accepts the first point `.` and stops at the second one:
`parseFloat('1.2.3') == '1.2'`.

1-js/05-data-types/02-number/article.md
jmbothe: No, the original wording is not wrong, but its awkward to use a verbal noun when a verb would suffice. But from a purely grammatical standpoint, the original wording is acceptable.

1-js/05-data-types/02-number/article.md
jmbothe: Ahhhh. good _point_. I think I might have to recuse myself from this one. The topic is beyond my expertise.

1-js/05-data-types/02-number/article.md
gwooly: It is difficult, but I agree if it's not present in the other sections as such, then it is OK to remove to keep consistency. I'll retract the comma but add 'it': and then divide **it** back.

1-js/05-data-types/02-number/article.md
gwooly: Yes correct, removed 'it's'.

1-js/05-data-types/02-number/article.md
gwooly: It could say 'will return' or 'returns'. Just 'return' reads wrong when read.

1-js/05-data-types/02-number/article.md
gwooly: So: 'Use `parseInt/parseFloat` for the "soft" conversion, which reads a number from a string and then returns the value they could read before the error.'

PR #151, opened 2017-08-30
querySelectorAll

mvp-Parfenov:
querySelectorAll


msim-way:
Нет) я имел в виду поиск только у первых детей ul списка

jonknapp:
I believe the point of the PR was to signify that `querySelector` returns a single element while `querySelectorAll` returns an array which is what it looks like the code is expecting.

iliakan:
closed due to inactivity

PR #153, opened 2017-09-01
Minor grammar/spelling edits

gwooly:
Double checked, should be good. Just pay attention to line 296 as I'm not 100% sure it's the correct edit in this context.

### DIFF
--- a/1-js/05-data-types/03-string/article.md
+++ b/1-js/05-data-types/03-string/article.md
@@ -624,7 +624,7 @@ For example:
 alert( 'S\u0307\u0323' ); // Ṩ
 ```
 
-This provides great flexibility, but also an interesting problem: the same symbol can be visually represented with different unicode compositions.
+This provides great flexibility, but also an interesting problem: two characters may visually look the same, but be represented with different unicode compositions.
 
 For instance:
 
--- a/1-js/05-data-types/03-string/article.md
+++ b/1-js/05-data-types/03-string/article.md
@@ -8,9 +8,9 @@ The internal format for strings is always [UTF-16](https://en.wikipedia.org/wiki
 
 ## Quotes
 
-Let's remember the kinds of quotes.
+Let's recall the kinds of quotes.
 
-Strings can be enclosed either with the single, double quotes or in backticks:
+Strings can be enclosed within either single quotes, double quotes or backticks:
 
 ```js
 let single = 'single-quoted';
@@ -19,7 +19,7 @@ let double = "double-quoted";
 let backticks = `backticks`;
 ```
 
-Single and double quotes are essentially the same. Backticks allow to embed any expression into the string, including function calls:
+Single and double quotes are essentially the same. Backticks however allow us to embed any expression into the string, including function calls:
 
 ```js run
 function sum(a, b) {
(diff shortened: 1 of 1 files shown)


1-js/05-data-types/03-string/article.md
```diff
@@ -598,33 +598,33 @@ alert( '𝒳'.charCodeAt(0).toString(16) ); // d835, between 0xd800 and 0xdbff
...
 
-The example:
+For example:
 
 ```js run
 alert( 'S\u0307\u0323' ); // Ṩ
 ```
 
-This gives great flexibility, but also an interesting problem: the same symbol visually can be represented with different unicode compositions.
+This provides great flexibility, but also an interesting problem: the same symbol can be visually represented with different unicode compositions.
```
iliakan: Sorry for my bad English, actually the idea was that different unicode compositions may represent the same visual character (see the example below).

The word "visually" referred to "the same symbol", not to "unicode composition".

1-js/05-data-types/03-string/article.md
```diff
@@ -635,42 +635,42 @@ alert( 'S\u0323\u0307' ); // Ṩ, S + dot below + dot above
...
+It's funny that in our situation `normalize()` actually brings together a sequence of 3 characters to one: `\u1e68` (S with two dots).
 
 ```js run
 alert( "S\u0307\u0323".normalize().length ); // 1
 
 alert( "S\u0307\u0323".normalize() == "\u1e68" ); // true
 ```
 
-In real, that is not always so. The reason is that symbol `Ṩ` is "common enough", so UTF-16 creators included it into the main table and gave it the code.
+In reality, this is not always the case. The reason being that the symbol `Ṩ` is "common enough", so UTF-16 creators included it in the main table and gave it to the code.
```
iliakan: "gave it to the code" - typo?

1-js/05-data-types/03-string/article.md
gwooly: If the word 'visually' is needed to let the reader understand that the symbol _looks_ the same but in the unicode it's actually _different_, then maybe use:

'this same symbol can be represented with different unicode compositions, but still visually look the same'.

1-js/05-data-types/03-string/article.md
gwooly: This is a difficult statement to interpret (from my beginner point-of-view).

Are you saying here that:

"S\u0307\u0323".normalize() == "\u1e68" // true 

because UTF-16 includes the `Ṩ` symbol in the spec, but not all symbols are included in this spec., so .normalize will not always translate to 'true'?

If I misunderstand this part, then I will leave your original wording so that it is technically correct.

1-js/05-data-types/03-string/article.md
iliakan: "This provides great flexibility, but also an interesting problem: two characters may visually look the same, but be represented with different unicode compositions." - ok?

1-js/05-data-types/03-string/article.md
iliakan: I mean that "S\u0307\u0323" gets normalized to a single character, because the symbol "s with two dots" is very common and has a separate code.

1-js/05-data-types/03-string/article.md
gwooly: Even better!

iliakan: please update the PR then, so that I can accept it along with other changes :)

1-js/05-data-types/03-string/article.md
gwooly: Hmm I think the wording should stay then, and just remove 'to'.

PR #154, opened 2017-09-01
replace "progress" with "timeFraction" for clarity

Annis-Monadjem:
I guess you meant "timeFraction" and not "progress", as it is more consistent with the meaning you have been conveying before and later in this chapter, and despite having no impact on final result, i still think its worth fixing. As usual, thanks Ilya for this site!


iliakan:
Thank you!

PR #178, opened 2017-09-15
Update article.md

totolouis:
A little imprecision that the continue/break is not allowed at the right side but also at the left side of the '?'


iliakan:
Yeah, but I never ever saw anyone to put it before `?`.

PR #217, opened 2017-09-27
Moar spaces in 07-operators.

Thespawnexe:
Added moar spaces to 07-operators.


iliakan:
Will that indeed look better?

PR #229, opened 2017-09-29
04 > 03-symbol - missing alert in example

MikkelSandbag:
Adding an alert to show change to object.


iliakan:
There's no value in showing a value here. It's obvious that it has changed.

PR #230, opened 2017-10-01
Fixed spy-decorator assignment test

viatsko:
Hi!

We should be checking here original function not spy to ensure that we are wrapping transparently.

The test wasn't working and probably that was just a "typo" to begin with.

This makes it work properly.


iliakan:
Sure thanks!

PR #240, opened 2017-10-13
Update solution.md

X-5B:
As described later, the answer should be 1. No more "undefined".


iliakan:
```
alert( alert(1) && alert(2) );
```

That actually shows 1 and then undefined.

X-5B:
You're right. after alert(1) shows 1, && stops, and then the outer alert()
shows undefined. I missed the outer alert somehow. I closed the PR.

PR #241, opened 2017-10-13
fix syntax problem

lilifan:
`in inside` seems to be a typo.


iliakan:
indeed a problem, but fix looks wrong :/

PR #292, opened 2017-11-05
In the task it is <=, not <.

Bemerauld:
Integer greater than zero.


iliakan:
Maybe it works ok with zero?

Bemerauld:
In the exercise, it is <=, then it becomes < in the solution. It is not a corrected mistake from the exercise since it's about code quality. Not big issue, but I like it to be perfect :ok_hand: 

http://puu.sh/yIZ7K/5d8b389bbf.png

PR #298, opened 2017-11-22
tc39/babel updates

hzoo:
(I work on Babel, and have attended TC39)

### DIFF
--- a/1-js/03-code-quality/06-polyfills/article.md
+++ b/1-js/03-code-quality/06-polyfills/article.md
@@ -24,7 +24,7 @@ Actually, there are two parts in Babel:
     There's a term "polyfill" for scripts that "fill in" the gap and add missing implementations.
 
     Two interesting polyfills are:
-    - [babel polyfill](https://babeljs.io/docs/usage/polyfill/) that supports a lot, but is big.
+    - [babel polyfill](https://babeljs.io/docs/usage/polyfill/).
     - [polyfill.io](http://polyfill.io) service that allows to load/construct polyfills on-demand, depending on the features we need.
 
 So, we need to setup the transpiler and add the polyfill for old engines to support modern features.


1-js/03-code-quality/06-polyfills/article.md
```diff
@@ -1,25 +1,21 @@
 
 # Polyfills
 
-The JavaScript language steadily evolves. New proposals to the language appear regularly, they are analyzed and, if considered worthy, are appended to the list at <https://tc39.github.io/ecma262/> and then progress to the [specification](http://www.ecma-international.org/publications/standards/Ecma-262.htm).
+The JavaScript language steadily evolves. New features requests are regularly [proposed](https://github.com/tc39/proposals) to the language and are vetted through a [4 part staging process](https://tc39.github.io/process-document/). If considered worthy, these are added to the evergreen spec at <https://tc39.github.io/ecma262/> and then get standardized to the [yearly release specification](http://www.ecma-international.org/publications/standards/Ecma-262.htm).
 
-Teams behind JavaScript engines have their own ideas about what to implement first. They may decide to implement proposals that are in draft and postpone things that are already in the spec, because they are less interesting or just harder to do.
-
-So it's quite common for an engine to implement only the part of the standard.
```
hzoo: I don't think this is true, or at least in the way you are describing. All engines implement the spec, they just take time to do so. The reason why they don't implement something is because they decide it's too slow to do (and usually something that hasn't gone through the Staging process). Now TC39 uses Stage 3 to make sure that 2 implementations have shipped so that there isn't this problem

1-js/03-code-quality/06-polyfills/article.md
```diff
@@ -28,7 +24,7 @@ Actually, there are two parts in Babel:
     There's a term "polyfill" for scripts that "fill in" the gap and add missing implementations.
 
     Two interesting polyfills are:
-    - [babel polyfill](https://babeljs.io/docs/usage/polyfill/) that supports a lot, but is big.
```
hzoo: This is only true if you import the whole thing. I don't know where you put put this in this tutorial but you should mention https://github.com/babel/babel-preset-env/tree/v1.6.1 which shows how Babel can let you compile only the parts that you need to (including polyfills)

1-js/03-code-quality/06-polyfills/article.md
iliakan: Sure, I'm aware of the proposals flow. With your update the description is better.

1-js/03-code-quality/06-polyfills/article.md
```diff
@@ -1,25 +1,21 @@
...
-
-So it's quite common for an engine to implement only the part of the standard.
-
-A good page to see the current state of support for language features is <https://kangax.github.io/compat-table/es6/> (it's big, we have a lot to study yet).
+A way to see the current state of support for language features is <https://kangax.github.io/compat-table/es6/> (it's big, we have a lot to study yet).
 
 ## Babel
 
-When we use modern features of the language, some engines may fail to support such code. Just as said, not all features are implemented everywhere.
+When we use modern features of the language, some engines may fail to support such code. Engines will not usually implement any proposal before Stage 3 since the feature is in flux, and shipping a feature in a browser takes a lot of work.
```
iliakan: This change requires the reader to get familiar with the proposals evaluation process. Like "what is stage 3?" We shouldn't require it.

The original phrase is correct.

1-js/03-code-quality/06-polyfills/article.md
iliakan: Sure, could you elaborate that in the text a little bit, just as you said that here? 👍

iliakan:
closed due to inactivity

PR #310, opened 2017-11-28
Added a href link.

harishekhar:
I believe this would be helpful for the readers, who are unaware of this topic.


iliakan:
That's a generic programming term. It would be more relevant to provide a Wiki link OR a link to ECMA standard, not 3 years ago article.

PR #316, opened 2017-11-30
Update index.html

usernamehw:


tomByrer:
Maybe add better descriptions, & why the change is important?

usernamehw:
Because
> TypeError: document.querySelector(...) is not iterable

PR #328, opened 2017-12-08
labelled -> labeled

julieeeeeee:
Labeled and labelled are both correct spellings, and they mean the same thing. How you spell the word depends on your audience. If you are writing for American readers, labeled is the preferred spelling. In other places, such as Great Britain and Canada, labelled is a more common spelling than labeled.
(Taken from https://www.grammarly.com/blog/labeled-labelled)


iliakan:
Thanks, guess it's the US variant.

PR #335, opened 2017-12-16
Update article.md

Annis-Monadjem:
Remove or alter test for integer property!


iliakan:
`Number.isInteger("1") == false`, naturally when you pass a string to it, it returns false.

Please read the statement again. You convert a string to integer and back to string, and it stays the same (string!).

PR #337, opened 2017-12-16
Update article.md

Annis-Monadjem:
Hi Iliakan: 
I'm reading the text very carefully but still I find (as in my #335 submission) the same problem. 
What i'm trying to say is that you cannot test "integer property with this formula:
                 String(Math.trunc(Number("X")))
because:
alert(String(Math.trunc(Number("+49"))) === String(Math.trunc(Number("49")))) 
So there is a strict equality between Line 1 and Line 2, and so the above formula is not a good test for "integer property"!

### DIFF
--- a/1-js/04-object-basics/01-object/article.md
+++ b/1-js/04-object-basics/01-object/article.md
@@ -414,7 +414,8 @@ alert( String(Math.trunc(Number("49"))) ); // <- Line 1: "49", same, integer pro
 alert( String(Math.trunc(Number("+49"))) ); // <- Line 2: "49", not same ⇒ not integer property
 
 // Above is Wrong, because:
-alert(String(Math.trunc(Number("+49"))) === String(Math.trunc(Number("49")))) // Strict equality between Line 1 and Line 2.
+alert(String(Math.trunc(Number("+49"))) === String(Math.trunc(Number("49")))) // true -> Strict equality between Line 1 and Line 2.
+alert(Math.trunc(Number("+49")) === Math.trunc(Number("49"))) // true -> Strict equality between Line 1 and Line 2.
 // and so the above is not a good test for "integer property"!!!
 
 alert( String(Math.trunc(Number("1.2"))) ); // "1", not same ⇒ not integer property
--- a/1-js/04-object-basics/01-object/article.md
+++ b/1-js/04-object-basics/01-object/article.md
@@ -410,8 +410,13 @@ So, "49" is an integer property name, because when it's transformed to an intege
 
 ```js run
 // Math.trunc is a built-in function that removes the decimal part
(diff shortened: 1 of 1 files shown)


iliakan:
Sorry I'm confused, what's wrong?

Anything in the lines below is wrong?
```
String(Math.trunc(Number("49"))) => "49" (same)
String(Math.trunc(Number("+49"))) => "49" (not same)
```

Annis-Monadjem:
Well, I really think you can improve on presentation by following my suggestions below.

In the manual you write:
```
The “integer property” term ... means a string that can be converted to-and-from an integer without a change.
So, “49” is an integer property name, because when it’s transformed to an integer number and back, it’s still the same. But “+49” and “1.2” are not:

alert( String(Math.trunc(Number("1.2"))) ); // "1", not same ⇒ not integer property 
alert( String(Math.trunc(Number("49"))) ); // "49", same, integer property   
alert( String(Math.trunc(Number("+49"))) ); // "49", not same ⇒ not integer property 
```
I'm suggesting that you remove the last 3 lines of code and replace them with far more elegant lines below (notice that my comments and results to first and third lines are consistent with definition of integer property, while that is not true with your lines of code above). The 3 lines of code below are easy to understand and can be verified, while your lines are doubtful:
String(Math.trunc(Number("1.2"))) == "1.2";   // false => a non-integer property
String(Math.trunc(Number("49"))) == "49";     // true   => an integer property
String(Math.trunc(Number("+49"))) == "+49"; // false => a non-integer property

Regards,

iliakan:
I added a few words there. The original code shows real value "1", and your code hides it, makes things less obvious. Hope better now.

Annis-Monadjem:
I disagree Iliakan! You are contradicting your own definition:

You have defined “INTEGER PROPERTY” as a **string** that can be converted to-and-from an integer without a change.

The above definition is best presented as (My solution):
String(Math.trunc(Number("1.2"))) == "1.2"; // false => a non-integer property (Line 1)
String(Math.trunc(Number("49"))) == "49"; // **true** => an integer property  (**Line 2**)
String(Math.trunc(Number("+49"))) == "+49"; // **false** => a non-integer property (**Line 3**)

and NOT as (Your solution):
alert( String(Math.trunc(Number("1.2"))) ); // "1", not same ⇒ not integer property (Line 1)
alert( String(Math.trunc(Number("49"))) ); // **"49"**, same, integer property (**Line 2**)
alert( String(Math.trunc(Number("+49"))) ); // **"49"**, not same ⇒ not integer property (**Line 3**)

Look carefully and compare Line 2 and Line 3 results and comments in Your lines (results are same "49" but comments are contradicting, because different), while in Me lines (results are different _true_ vs. _false_ while my comments are properly corresponding).

Well, I hope you see what I have been trying to say all this time.

Good Day!

PR #349, opened 2018-01-09
Added Atom to IDE list

mittalyashu:
Added Atom to list of IDE.


iliakan:
Sorry no much use among my friends. Removed Komodo though.

mittalyashu:
I have added atom instead of Komodo

iliakan:
There is atom there already =). Not atom IDE.

mittalyashu:
But I think Atom and Atom IDE is different.

PR #357, opened 2018-01-20
Return explanation

brentguf:
Added an explanation for writing a return statement spread over multiple lines. 

This is often used in React, so I thought it would be cool to expose new developers to this idea a first time.

### DIFF
--- a/1-js/02-first-steps/14-function-basics/article.md
+++ b/1-js/02-first-steps/14-function-basics/article.md
@@ -332,6 +332,26 @@ return*!*;*/!*
  (some + long + expression + or + whatever * f(a) + f(b))
 ```
 So, it effectively becomes an empty return. We should put the value on the same line instead.
+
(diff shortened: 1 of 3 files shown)


1-js/02-first-steps/14-function-basics/article.md
```diff
@@ -332,6 +332,26 @@ return*!*;*/!*
  (some + long + expression + or + whatever * f(a) + f(b))
 ```
 So, it effectively becomes an empty return. We should put the value on the same line instead.
+
```
iliakan: For spreading an expression into multiple lines - a good idea to say that, but need to add rules when it works. 

Not only "+" operator can do that.

Probably, more details about that should go to another part of the tutorial where semicolons are discussed.

brentguf:
Yes, you could add it somewhere in the Code Structure chapter and then refer to that chapter from this chapter (the Functions chapter). Or what exactly do you want me to do? Could you be clearer?

iliakan:
Back to JS :) Almost all changes were merged already, and optional semicolons are discussed are http://javascript.info/structure.

PR #360, opened 2018-01-22
Fixed incorrect it's/its

andrej-repisky:


andrej-repisky:
@iliakan Although this PR is merged, the changes are still not propagated to web: http://javascript.info/object. Why is that?

iliakan:
Right, I had some other pending changes so I just flushed them as a whole.

iliakan:
should be visible now.

PR #361, opened 2018-01-27
Date exercises

brentguf:
Alternative solutions for the last two exercises of the Date chapter. Feel free to add if you like them. :)


iliakan:
Are you sure that alternative solutions benefit the reader?

brentguf:
Well, in my opinion, they provide two important things:

1. Show a different way of solving a problem, which means someone will have more tools later to solve a different problem. "To someone with a hammer, every problem looks like a nail." 

2. Show that there is (most often) not one way to solve a problem. When I just started learning programming, I always felt gutted when I couldn't solve an exercise in exactly the same way as the instructor. So it could be a confidence booster.

But feel free to do what you feel is right.

PR #385, opened 2018-02-19
Instance of

brentguf:
Error fix + added a bit more explanation.


iliakan:
Sorry don't understand the idea of the change.

brentguf:
Well, you're showing here how we run through the prototype chain. We need to check the prototypes of the same constructor function (`Animal` in this case) for this looping to make sense. So if you're first check is for Rabbit.__proto__ we already have a match, so it doesn't make sense for what you're trying to explain. Right?

iliakan:
Here, `rabbit.__proto__` is the same as `Rabbit.prototype`, and not the same as `Animal.prototype`.

If I'm gettting the idea right,  you suggest to add the line `rabbit.__proto__ === Animal.prototype` ?
That's a wrong statement, or maybe I'm just not getting something?

PR #393, opened 2018-02-27
Coordinates chapter

brentguf:
Error fix in the code.


iliakan:
What's the error?

brentguf:
There's no error. `window` is the global object, so doesn't need to be specified. My mistake.

PR #395, opened 2018-03-01
Keyboard keydown keyup

brentguf:
I think we need `event.key` and not `event.code` for this task. In any case, the solution you proposed wasn't working on my computer (Azerty keyboard), because when I press Q and W, the key codes are different from yours. When I press Q, the corresponding key code is KeyA. When I press W, the key code is KeyZ.


iliakan:
There's another problem if we switch to "key".

The hotkey will *stop working* when the visitor changes language. 

For instance, in Russian the same key will have `event.key="й"`.

PR #396, opened 2018-03-02
Sorting table exercise

brentguf:
The ternary operator looks redundant here.


iliakan:
Wrong change.

brentguf:
`sort`'s default sort is based on the character's Unicode values, so it works exactly the same with or without the ternary operator on my computer. 

Would you mind to shortly elaborate?

iliakan:
As described in the [spec](https://tc39.github.io/ecma262/#sec-array.prototype.sort), the comparison function should return 1 for "greater" and -1 for "less". Not `true`/`false` like in the suggested change.

PR #415, opened 2018-03-26
Simple Typo Error

vishugosain:


upupming:
This isn't a typo, it means `default` is optional.

PR #418, opened 2018-03-27
Rewordings and corrections to Promise Basics

davegregg:
- Corrected a few things that seem to be leftovers from inaccurate translation.
- Reworded several things to more natural-sounding English.
- Added some more clarification in a few places.
- Nearly replaced the "Handlers of `.then`/`.catch` are always asynchronous" section in an effort to make it more comprehensible for people who may be new to asynchronous operations.

### DIFF
--- a/6-async/02-promise-basics/article.md
+++ b/6-async/02-promise-basics/article.md
@@ -1,18 +1,18 @@
 # Promise
 
-Imagine that you're a top singer, and fans ask for your next upcoming single day and night.
+Imagine that you're a top singer, and fans ask day and night for your upcoming single.
 
-To get a relief, you promise to send it to them when it's published. You give your fans a list. They can fill in their coordinates, so that when the song becomes available, all subscribed parties instantly get it. And if something goes very wrong, so that the song won't be published ever, then they are also to be notified.
+To get some relief, you promise to send it to them when it's published. You give your fans a list to which they can subscribe for updates. They can fill in their email addresses, so that when the song becomes available, all subscribed parties instantly receive it. And even if something goes very wrong, say, if plans to publish the song are cancelled, they will still be notified.
 
 Everyone is happy: you, because the people don't crowd you any more, and fans, because they won't miss the single.
 
-That was a real-life analogy for things we often have in programming:
+This is a real-life analogy for things we often have in programming:
 
-1. A "producing code" that does something and needs time. For instance, it loads a remote script. That's a "singer".
-2. A "consuming code" wants the result when it's ready. Many functions may need that result. These are  "fans".
-3. A *promise* is a special JavaScript object that links them together. That's a "list". The producing code creates it and gives to everyone, so that they can subscribe for the result.
+1. A "producing code" that does something and takes time. For instance, the code loads a remote script. That's a "singer".
+2. A "consuming code" that wants the result of the "producing code" once it's ready. Many functions in your "consuming code" may need that result. These are "fans".
+3. A *promise* is a special JavaScript object that links the "producing code" and the "consuming code" together. In terms of our analogy: this is the "subscription list". The "producing code" takes the time it needs to produce the promised result, and the "promise" makes it available to all of the subscribed code when it's ready.
 
-The analogy isn't very accurate, because JavaScript promises are more complex than a simple list: they have additional features and limitations. But still they are alike.
+The analogy isn't terribly accurate, because JavaScript promises are more complex than a simple subscription list: they have additional features and limitations. But this will serve for an introduction.
 
 The constructor syntax for a promise object is:
 
@@ -22,25 +22,25 @@ let promise = new Promise(function(resolve, reject) {
 });
 ```
 
-The function passed to `new Promise` is called *executor*. When the promise is created, it's called automatically. It contains the producing code, that should eventually finish with a result. In terms of the analogy above, the executor is a "singer".
+The function passed to `new Promise` is called the *executor*. When the promise is created, this executor function is called (run) automatically. It contains the producing code, that should eventually produce a result. In terms of the analogy above: the executor is the "singer".
 
 The resulting `promise` object has internal properties:
 
-- `state` -- initially is "pending", then changes to "fulfilled" or "rejected",
-- `result` -- an arbitrary value, initially `undefined`.
+- `state` — initially "pending", then changes to either "fulfilled" or "rejected",
+- `result` — an arbitrary value of your choosing, initially `undefined`.
 
-When the executor finishes the job, it should call one of:
+When the executor finishes the job, it should call one of the following functions:
 
-- `resolve(value)` -- to indicate that the job finished successfully:
+- `resolve(value)` — to indicate that the job finished successfully:
     - sets `state` to `"fulfilled"`,
     - sets `result` to `value`.
-- `reject(error)` -- to indicate that an error occurred:
+- `reject(error)` — to indicate that an error occurred:
     - sets `state` to `"rejected"`,
     - sets `result` to `error`.
 
 ![](promise-resolve-reject.png)
 
-Here's a simple executor, to gather that all together:
+To gather all of that together: here's an example of a Promise constructor and a simple executor function with its "producing code" (the `setTimeout`):
 
 ```js run
 let promise = new Promise(function(resolve, reject) {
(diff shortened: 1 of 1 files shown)


6-async/02-promise-basics/article.md
```diff
@@ -1,18 +1,18 @@
...
 
-That was a real-life analogy for things we often have in programming:
+This is a real-life analogy for things we often have in programming:
 
-1. A "producing code" that does something and needs time. For instance, it loads a remote script. That's a "singer".
-2. A "consuming code" wants the result when it's ready. Many functions may need that result. These are  "fans".
-3. A *promise* is a special JavaScript object that links them together. That's a "list". The producing code creates it and gives to everyone, so that they can subscribe for the result.
+1. A "producing code" that does something and takes time. For instance, the code loads a remote script. That's a "singer".
+2. A "consuming code" that wants the result of the "producing code" once it's ready. Many functions in your "consuming code" may need that result. These are "fans".
+3. A *promise* is a special JavaScript object that links the "producing code" and the "consuming code" together. In terms of our analogy: this is the "subscription list". The "producing code" takes the time it needs to produce the promised result, and the "promise" makes it available to all of the subscribed code when it's ready.
```
iliakan: Please reread the last sentence. Is it ok?

iliakan: Also, maybe:

2. A "consuming code" that wants the result of the "producing code" once it's ready. Many functions may need the result. These are "fans".

6-async/02-promise-basics/article.md
```diff
@@ -22,25 +22,25 @@ let promise = new Promise(function(resolve, reject) {
 });
 ```
 
-The function passed to `new Promise` is called *executor*. When the promise is created, it's called automatically. It contains the producing code, that should eventually finish with a result. In terms of the analogy above, the executor is a "singer".
+The function passed to `new Promise` is called the *executor*. When the promise is created, this executor function is called (run) automatically. It contains the producing code, that should eventually produce a result. In terms of the analogy above: the executor is the "singer".
```
iliakan: Maybe remove `(run)`?

6-async/02-promise-basics/article.md
```diff
@@ -56,16 +56,16 @@ let promise = new Promise(function(resolve, reject) {
 
 We can see two things by running the code above:
 
-1. The executor is called automatically and immediately (by `new Promise`).
-2. The executor receives two arguments: `resolve` and `reject` -- these functions come from JavaScript engine. We don't need to create them. Instead the executor should call them when ready.
+1. The executor is called automatically and immediately (by the `new Promise`).
+2. The executor receives two arguments: `resolve` and `reject` — these functions are pre-defined. We don't need to create them. Instead, we should write the executor to call them when ready.
```
iliakan: The word "pre-defined" is not as obvious as "come from JavaScript engine", how you think? I mean, the reader should understand that they are come from by Javascript itself.

6-async/02-promise-basics/article.md
```diff
@@ -94,46 +94,53 @@ let promise = new Promise(function(resolve, reject) {
 });
 ```
 
-The idea is that a job done by the executor may have only one result or an error. In programming, there exist other data structures that allow many "flowing" results, for instance streams and queues. They have their own advantages and disadvantages versus promises. They are not supported by JavaScript core and lack certain language features that promises provide, we don't cover them here to concentrate on promises.
+The idea is that a job done by the executor may have only one result or an error. (If you are familiar with other languages, you might be aware of data structures which allow many "flowing" results, like streams and queues. They have their own advantages and disadvantages as compared with Promises. But as they are not supported by the JavaScript core, we won't be covering them.)
```
iliakan: Words about "familiarity with other languages" here are not terribly precise. Such structures exist in JS as well, and are quite well-known (just maybe not know to the reader yet). No need to mention other languages here.

6-async/02-promise-basics/article.md
```diff
@@ -94,46 +94,53 @@ let promise = new Promise(function(resolve, reject) {
...
+In practice, an executor usually does something asynchronously and calls `resolve`/`reject` after some time, but it doesn't have to. We can call `resolve` or `reject` immediately, like this:
 
 ```js
 let promise = new Promise(function(resolve, reject) {
   resolve(123); // immediately give the result: 123
 });
 ```
 
-For instance, it happens when we start to do a job and then see that everything has already been done. Technically that's fine: we have a resolved promise right now.
+For instance, it might happen when we start to do a job but then see that everything has already been completed. That is fine: we have a resolved Promise immediately.
```
iliakan: Maybe "it might happen *that* we start to do a job..."?

6-async/02-promise-basics/article.md
```diff
@@ -167,7 +174,7 @@ promise.then(
 );
 ```
 
-If we're interested only in successful completions, then we can provide only one argument to `.then`:
+If we're interested only in successful Promises, then we can provide only one function argument to `.then`:
```
iliakan: Why the change `completions` -> `Promises`? Here we're talking about promise *results*.

6-async/02-promise-basics/article.md
```diff
@@ -179,7 +186,7 @@ promise.then(alert); // shows "done!" after 1 second
 */!*
 ```
 
-If we're interested only in errors, then we can use `.then(null, function)` or an "alias" to it: `.catch(function)`
+If we're interested only in errors, then we can use `.catch(errorHandlingFunction)`, which is just like doing `.then(null, errorHandlingFunction)`.
```
iliakan: `...which is exactly the same`, not "just like".

6-async/02-promise-basics/article.md
```diff
@@ -208,32 +215,38 @@ promise.then(alert); // done! (shows up right now)
 That's handy for jobs that may sometimes require time and sometimes finish immediately. The handler is guaranteed to run in both cases.
 ````
 
-````smart header="Handlers of `.then/catch` are always asynchronous"
-To be even more precise, when `.then/catch` handler should execute, it first gets into an internal queue. The JavaScript engine takes handlers from the queue and executes when the current code finishes, similar to `setTimeout(..., 0)`.
+````smart header="Handlers of `.then`/`.catch` are always asynchronous"
+Even when the Promise is immediately resolved, code which occurs on lines below your `.then`/`.catch` can still run first. When the JavaScript engine sees your Promise, it will not only run the executor (the "producing code"), it will also go ahead and start executing code which occurs after your Promise and its `.then`/`.catch` calls.
+
+So, when it is time for the function passed to `.then`/`.catch` to execute (nearly immediately in the case of an immediately-resolved Promise), the JavaScript engine puts it into an internal execution queue which will then already have items to be run.
```
iliakan: The new explanation is hard to understand for the person not familiar with the topic already.

6-async/02-promise-basics/article.md
```diff
@@ -208,32 +215,38 @@ promise.then(alert); // done! (shows up right now)
...
 
-In other words, when `.then(handler)` is going to trigger, it does something like `setTimeout(handler, 0)` instead.
+The JavaScript engine is trying to do as many things at virtually the same time as possible. Everything is racing to get done. So, in the example code below, behind the scenes, you can imagine that the events might occur in this order:
```
iliakan: "The JavaScript engine is trying to do as many things at virtually the same time as possible" - that's confusing. May seem like we're talking about multithreading, but we're not.

6-async/02-promise-basics/article.md
davegregg: Sorry, just getting back to this PR. Slightly reworded.

6-async/02-promise-basics/article.md
davegregg: Removed the parenthetical. Thanks!

6-async/02-promise-basics/article.md
davegregg: Either works in this case. I rephrased this way: `For instance, this might happen when we start to do a job...`

6-async/02-promise-basics/article.md
davegregg: Reverted to "completions".

6-async/02-promise-basics/article.md
davegregg: Clarified with your wording

davegregg:
Sorry, I've had to suddenly write a lot more curriculum for my code school, so I put this off, but I'm mostly done with the revision. Just have to finish the last big part, which I will do this weekend.

davegregg:
Alright, I have attempted to address all of your concerns. Please see the latest state of the PR.

iliakan:
Great stuff, @davegregg 
By the way, are you interested to learn git?

I have a great screencast about it, much better than all I've seen (ok, I hope so, people can judge), it is being be translated to English, maybe you could review the text or even voice the English version?

davegregg:
Yeah, I'd be up for helping with that. I'm pretty comfortable with git, so it won't be an issue.

iliakan:
I added you as a collaborator.
As of now, the 1st issue text is ready: https://github.com/iliakan/git-screencast/blob/master/s1/e1-en.md

Please fix wherever you think is better. Also if you see any mistakes about git (hope there won't be any), let me know, so I can fix those in the English version.

P.S. Do you have Mac?

iliakan:
Hi, 

Thank you very much for your wish to participate. 

There are three parts here:
1) Review the translation to ensure good English (needs only native English).
2) Check the text to ensure correctness, just in case (needs good Git knowledge, but a student may be ok too, they can signal if an explanation is not clear and learn Git in the process)
3) Voice the English version (may require a Mac, cause the screencast is written using Mac-only software, and normal/good English diction).

All that can be done by different people or by the same person.

davegregg:
Alright, let me loop in Jake (@jakerjohnson94) -- can you add him to the git screencast project? -- as I think he could do a good job with helping to ensure the correctness and naturalness of the English text. When that's done, I'll decide who will help record the English version. I'm assuming the English version will be *just* audio narration that will be overlayed on top of the video you've already taken?

iliakan:
Here's the repo: https://github.com/iliakan/git-screencast

I added Readme about the translation process.

davegregg:
I've started my student on the process of proofreading and we'll get back with you

PR #433, opened 2018-05-31
1-js/07-object-oriented-programming/08-class-patterns/

IveHD:


iliakan:
That's translation to Chinese?

iliakan:
Guess, we should close this PR as it doesn't belong here?

PR #437, opened 2018-06-05
Modified argument on the basis of comment

kunalmhatre:
According to the comment, the argument should've been NaN, but was "str".


iliakan:
converts to NaN

kunalmhatre:
Oh, sorry.

PR #438, opened 2018-06-07
uzbek language first commit

aruzikulov:
Hi, I loved the project, Would like to translate it into Uzbek language


iliakan:
UZ translatio is added.

PR #444, opened 2018-06-18
Edit due to an error

eng-ahmedsaleh:
forEach has a callback function which is invoked here by the global object (Window)


iliakan:
this=undefined unless forgot 'use strict'

PR #449, opened 2018-06-29
Updated article about Object Methods

davidbludlow:
When I came across part of the code examples on https://javascript.info/object-methods#internals-reference-type that said that the code line (user.name == "John" ? user.hi : user.bye)(); didn't work, it bugged me that I couldn't figure out what code changes I would need to make to make it work. The article just explained why it didn't work, but offered no hint as to how to make it work. I therefore wasted time trying to figure it out. After I gave up and continued with the tutorial, I figured out that a solution was written over 20 pages later on https://javascript.info/bind#solution-2-bind . I offer this change so that others won't waste time the same way I did.


iliakan:
Merged.

PR #455, opened 2018-07-02
Update article.md

meo1719:
I propose that line 250 be updated from "...he can use 'const' to guarantee it..." to "he or she can use 'const' to guarantee it," since women are programmers as well.


iliakan:
Usually people use "they".

PR #472, opened 2018-08-02
Use gender neutral pronouns

benhjames:
Because it's an important thing to do ❤️


iliakan:
Thank you.
Hopefully no one got offended :)

PR #479, opened 2018-08-19
Don't need arrow function here

friendlyanon:
The signature of `setTimeout` is:
```js
setTimeout(function[, delay, param1, param2, ...]);
```
https://devdocs.io/dom/windoworworkerglobalscope/settimeout

### DIFF
--- a/6-async/05-async-await/article.md
+++ b/6-async/05-async-await/article.md
@@ -144,7 +144,7 @@ class Thenable {
   then(resolve, reject) {
     alert(resolve); // function() { native code }
     // resolve with this.num*2 after 1000ms
-    setTimeout(() => resolve(this.num * 2), 1000); // (*)
+    setTimeout(resolve, 1000, this.num * 2); // (*)
   }
 };
 


welkin25:
maybe fix the settimeout-setinterval page as well for the same issue? Or should that be a separate PR?

friendlyanon:
Actually, I went thru a lot of code and there are an unreasonable amount of arrow functions used in places where it is absolutely unnecessary, if not a waste.

iliakan:
That setTimeout notation was created prior to arrow functions.

It's terrible actually. Not obvious at all.

iliakan:
Think yourself, that's crazy - put function parameters aside from the function.

friendlyanon:
List of builtins using this kind of function signature:

- https://devdocs.io/dom/windoworworkerglobalscope/settimeout
- https://devdocs.io/dom/windoworworkerglobalscope/setinterval
- https://devdocs.io/node/timers#timers_setimmediate_callback_args
- https://devdocs.io/node/process#process_process_nexttick_callback_args

It'd be best for a community curated resource for learning to highlight these kinds of things.
I've inherited too many codebases to work on where the previous person used
```js
process.nextTick(() => someFunc(arg1, arg2, arg3));
```
instead of properly making use of the builtin and writing
```js
process.nextTick(someFunc, arg1, arg2, arg3);
```

It's also a very obvious vararg pattern. Clarity in the proper usecases for the builtins is maintained.

iliakan:
Nah, the latter notation was created only for performance.

I mean `process.nextTick(someFunc, arg1, arg2, arg3);`.
Not for clarity, for performance only.

iliakan:
And also please re-read my message again, I was talking about setInterval/setTimeout.

Don't mess in process.nextTick please.

PR #497, opened 2018-09-05
To be more concise

Ghost-017:


iliakan:
Is it not clear without these words?

Ghost-017:
> Is it not clear without these words?

It is, but I just want to say that it is not required anymore to include those compared to the past.

PR #511, opened 2018-09-19
Typo changes

Ghost-017:
- HTML is not really a language
- I'd prefer to use the word 'parse' because it means, "The process of analyzing text made of a sequence of tokens to determine its grammatical structure ... . The parser then builds a data structure ... for compilers, interpreters or translators to create an executable program or library." Instead of just reading them.

https://stackoverflow.com/questions/2933192/what-is-parsing-in-terms-that-a-new-programmer-would-understand


iliakan:
I merged it, but there's a problem.

Not everyone knows the word parse. Maybe an explanation in parens?

Ghost-017:
Yeah definitely, do you have a simple explanation in mind? Or provide the link above.

PR #514, opened 2018-09-21
Update warning on Ninja Code page

catmclough:
I think sarcasm is the word to use here rather than irony. I think it's important to be clear here because you don't want anyone mistaking this page for real advice! 😬


iliakan:
I like the word "irony" more. Sarcasm is an evil thing. I mean no evil ;)

PR #515, opened 2018-09-22
Update article.md

nirgeier:
- Added ES6 initialize array method
- Replace alert with console.log


iliakan:
ES6 method you suggesting is just bad. `new Array(10)` is enough.
Don't remove `alerts`, they are used everywhere for convenience.

PR #522, opened 2018-09-29
Added the features about Visual Studio Code editor

Oscar2Explorr:


iliakan:
VS Code exists for quite some time already and is cross-platform.  I'll add some words about it.

PR #529, opened 2018-10-01
Cleared a misunderstanding

Ghost-017:
Style is not a property, so I think you meant the display?


Ghost-017:
Also, this confused me a lot, until I realized that I think you meant the display property.

PR #532, opened 2018-10-02
An addition to solution

Ghost-017:


iliakan:
The task mentions no borders, the addition purpose is unclear. 
I understand what you mean, but that's actually confusing.

Ghost-017:
Oh ok, I thought I was going to find the scrollbar width in an element because the examples were always revolved around an element. So the scrollbar in an entire document completely slipped my mind. Perhaps changing the title to, "What is the scrollbar width of the document?".

iliakan:
The task is to find the scrollbar width of an element.

Not the "scrollWidth".

Ghost-017:
I know, width not height.

PR #534, opened 2018-10-02
Remove Visual Studio Code from IDE

Tchoupinax:
Hello, 

I disagree to call Visual Studio Code as an IDE. As i understand IDE versus text editor and as i have listened around me, VSCode is just a very good text editor, with many good plugins, like Atom, Sublime Text or Brackets.

According me, there are all text editor. IDE are Visual Studio, PHPStorm, Xcode...

We can speak about how change this text, I briefly change it.

Regards,

### DIFF
--- a/1-js/01-getting-started/2-code-editors/article.md
+++ b/1-js/01-getting-started/2-code-editors/article.md
@@ -13,12 +13,11 @@ An IDE loads the project (can be many files), allows navigation between files, p
 If you haven't considered selecting an IDE yet, look at the following variants:
 
 - [WebStorm](http://www.jetbrains.com/webstorm/) for frontend development and other editors of the same company if you need additional languages (paid).
-- [Visual Studio Code](https://code.visualstudio.com/) (free).
 - [Netbeans](http://netbeans.org/) (paid).
 
 All of the IDEs except cross-platform.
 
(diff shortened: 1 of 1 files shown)


iliakan:
I'd request more input on that ;)

iliakan:
Quite probably you're right, I'm not VS Code user though.

Ghost-017:
Text editors open quickly. IDE's take a while to open. VS code opens quickly.

iliakan:
IDEs allow project-wide features, like autocompletion.

VS Code allows project-wide autocompletion, jumping to declarations?

cpxPratik:
VS Code comes into gray area with supporting many features that an IDE usually have such as in debugging, refactoring etc. but it is still source code editor at its core, not full-blown IDE. 
The main issue that i have with PR diff is, It says Visual Studio explicitly windows-only, but Visual studio is also available on macOS though not the same software. So it cannot be said as windows-only.

dagolinuxoid:
'Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS.' So technically yeah - VS Code is defined as a Source Code Editor and not an IDE. At the other hand the line between text editor and IDE  becomes more and more blur nowadays so I think it's not that crucial.

AngusMacrae:
I agree VSCode blurs the line between the two, but if we have to classify it as one or the other, I'd definitely still put it in the "editor" category rather than IDE.

iliakan:
Well, not sure if it matters at all =)

PR #539, opened 2018-10-04
Fixed: code desc & task desc

upupming:
We found two small mistakes while translating this article into Chinese:

1. There are only **10** runs for `requestAnimationFrame` in the code, not **20**.
2. In task 2, we want to move the ball from **left** to **right**, not the reverse direction.


iliakan:
Right, thanks.

PR #557, opened 2018-10-17
a tiny correction

dagolinuxoid:
I guess 'find return' looks a bit awkward, maybe you wanted to write 'find and return' but dropped 'and' or something like this. Just 'return' seems fine.
PS. I don't believe that I didn't solve the simple task yet. Few years ago I was able to handle much more complex stuff ... feelsWierdAndBadMan.


tendermario:
Hey, super minor, but I was just curious about the branch naming - I notice a bunch of people are using branch name `patch-1` (and this one using `patch-2` since I imagine this is your second commit to the repo).

I'm wondering is this some sort of standard pattern that I haven't heard about? It looks almost like an automated workflow how people are using this similar pattern.

dagolinuxoid:
@tendermario you can give whatever name you want to your branch explicitly, but if you're lazy you will get something less descriptive like `patch-somenumber` by default.

PR #562, opened 2018-10-22
fixSentence

dagolinuxoid:
[-] What’s the benefit over using Symbol("id") over a string "id"?
[+] What’s the benefit of using Symbol("id") over a string "id"?


dagolinuxoid:
GitHub got nuts, so I'm going to delete this duplicate.

PR #581, opened 2018-10-28
placed a semicolon after this function expression

Mrchazaaa:


Mrchazaaa:
just thought that this should be added

PR #582, opened 2018-10-29
Update wording and grammar 

evazorro:
As I was following the tutorial, I noticed some areas where the language could be made to sound more natural, as well as a few grammar errors. Hope this is helpful. :) Thanks for the tutorial; I learned a lot.


iliakan:
Thanks!

PR #616, opened 2018-11-09
Improved phrasing

jayhybrid:
Again, a matter of taste. I changed **but** to **and** because I didn't really see the need for a *but*. I also think it sounds better.


jayhybrid:
Sorry about that, should be corrected now.

iliakan:
Thanks!

PR #617, opened 2018-11-09
Obvious variable names

jayhybrid:
Being the JS newbie I am, it was very misleading for me when I read **form** and **forms**. At first, I didn't know what the variable was and had difficulties getting my code to work. I think having idiot-proof variable declaration can help here.

### DIFF
--- a/2-ui/4-forms-controls/1-form-elements/article.md
+++ b/2-ui/4-forms-controls/1-form-elements/article.md
@@ -27,12 +27,12 @@ For instance:
 
 <script>
   // get the form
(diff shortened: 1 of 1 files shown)


iliakan:
We use names like `elem` a lot in the tutorial. 

May it be so that you're the only one who got confused?

iliakan:
I don't see a problem here. Any feedback?

jayhybrid:
It was just a suggestion, if you don't see the need to change anything, leave it, we can close this then.

Issue #573, opened 2018-10-26
Clarification on when functions as default parameters are called


mathiasbockwoldt:
In `javascript-tutorial-en/1-js/02-first-steps/14-function-basics/article.md` around line 200, it is mentioned that functions can be used as default parameters: `function showMessage(from, text = anotherFunction()) { ... }`. Could you please specify if `anotherFunction()` is called every time `showMessage(...)` is called? I come from Python and there, the function headers are evaluated only once and the result of `anotherFunction()` at compile/interpretation time would be used for every subsequent call of `showMessage(...)`.
I'm not sure how to phrase this, so I opened an issue instead of sending a PR.

dagolinuxoid:
@mathiasbockwoldt that's interesting. I'm not sure what exactly you're trying to convey either though :). I bet `anotehrFunction()` is called every time upon `showMessage()` and I don't think that JS engine makes some crazy optimization stuff about it (I could be wrong). There are exist  some techniques that you can utilize such as *memoization*  which reduces heavyweight computing for each next call - [osmaniMemoize](https://addyosmani.com/blog/faster-javascript-memoization/). PS. I think the page `article.md` looks just fine (the only concern - the [image](https://github.com/iliakan/javascript-tutorial-en/blob/master/1-js/02-first-steps/14-function-basics/function_basics.png) is a bit pale), thanks for the question.

mathiasbockwoldt:
In Python:

```python
x = 1

def anotherFunction():
	return x
	
def show(text = anotherFunction()):
	print(text)
	
show() # 1
x = 2
show() # 1
```

This is, because `anotherFunction` is called only once during interpretation and the return value saved for any subsequent call, no matter if x is changed or not.

In JavaScript:
	
```javascript
let x = 1;

function anotherFunction() {
	return x;
}

function show(text = anotherFunction()) {
	alert(text);
}

show(); // 1
x = 2;
show(); // 2
```

In JavaScript, `anotherFunction` is called every time, `show` is called and, hence, shows the current value of `x`.

Since you sometimes refer to other common languages (like mentioning that `"0"` evaluates to false in PHP), I thought, this might be a possible addition to the chapter. I had to write the little script above to understand the behaviour that is so different from the Python behaviour. But I can understand if this case is too special for the general introduction.

mathiasbockwoldt:
[...]

PR #622, opened 2018-11-14
Added info about evaluation of default parameters

mathiasbockwoldt:
See issue #573 for reasoning.
I hope this is in accordance with all style guides etc.


iliakan:
Sure, a good note.

PR #631, opened 2018-11-21
update if condition for equal to zero check for n

darshakmehta:
Missing equal to zero check in the if condition


iliakan:
0 is ok.

darshakmehta:
The Task mentioned about considering ```n <= 0``` which confused me. Since currently 0 raise to 0 = 1; which is incorrect? It should be undefined right?

PR #638, opened 2018-11-26
add test for getMaxSubSum

dagolinuxoid:
*Why* : Down below is my first (wrong and sloppy) implementation of the `getMaxSubSum` function which passed all of your tests from plunker btw. So I added one more test case in order to prevent such a crappy code from passing tests.
```js
function getMaxSubSum(arr){
  arr = cleanEdges(arr);
  if (arr.every(e=>e>0)) {
    return arr.reduce((sum,e)=>sum+e,0);
  }
  arr = groupSimilars(arr);
  const sums = [arr[0]];
  let sum = arr[0], curr = arr[0];
  for (let i = 1; i < arr.length; i += 2) {
    const n = arr[i], p = arr[i+1];
    if (p >= Math.abs(n) && sum >= Math.abs(n)) {
      sum += p + n;
      curr += p + n;
    } else {
      p >= sum
      ? curr = sum = p
      : sums.push[sum], curr = sum = p;
    }
  }
  sums.push[sum];
  return sums.some(v=>v>curr)
    ? Math.max(...sums)
    : curr;
}
function cleanEdges (arr) {
  arr = [...arr];
  while (arr[0] <= 0) arr.shift();
  while (arr[arr.length-1] <= 0) arr.pop();
  return arr;
}
function groupSimilars (arr) {
  let reducedArr = [], box = 0;
  for (let i=0; i<arr.length;){
    if (arr[i] >= 0)
      while (arr[i] >= 0) box += arr[i++];
    else
      while (arr[i] < 0) box += arr[i++];
    reducedArr.push(box);
    box = 0;
  }
  return reducedArr;
}
module.exports = {getMaxSubSum, cleanEdges, groupSimilars};
```


dagolinuxoid:
PS. I don't like the presence of the two unnecessary Merge commits though :( and I want to include just this one 7dc3cda

dagolinuxoid:
@iliakan psst - any tips how to avoid using `push[something]` instead of `push(something)` alike ? It's just ridiculous.

iliakan:
You want to test for `push[something]`? That clearly won't work anyway.

dagolinuxoid:
Nope I don't  :) I want to my editor to yell at me when I invoke method using square brackets instead of parenthesis because obviously JS is too friendly and don't consider the action as an illegal thing. It happens probably when a coder is thinking about an array and its fingers automatically press [ ]. Hope it's not going to happen in future.

PR #691, opened 2018-12-19
Fix a typo

Omar-Sh1:
I just fixed the typo and make it more readable .


iliakan:
That's not a typo, `default` is an argument.

PR #692, opened 2018-12-19
Fix positive issue

medifle:


iliakan:
What's wrong here with `0`?

medifle:
> What's wrong here with `0`?

I thought the task is to **Find positive numbers**.
So we do not include 0?

If this PR is merged, the `task.md` also need a bit fix

iliakan:
You're totally right about positive numbers. But the regexp shouldn't totally exclude 0, so it's `[1-9]\d*`. Updated the task, thanks.

PR #706, opened 2018-12-27
Delete redundant Array.from() call

dagolinuxoid:
Since Object.values(obj) returns an array it's an overkill to make additional Array.from() call on it.
PS.
BTW I've used similar but a little bit different approach to solve it on my own before checking answers
```js
function aclean(arr) {
  const s = new Set();
  return arr.reduce((r,v,i) => {
    v = v.toLowerCase().split('').sort().join('');
    return s.has(v) ? r : (s.add(v), [...r, arr[i]]);
  }, []); 
}
```


dagolinuxoid:
I agree and appreciate your feedback. Probably the change makes it more readable but using of negation isn't that cool as well.
```js
if (!s.has(v)){
s.add(v);
r.push(arr[i]);
}
return r
```

PR #718, opened 2019-01-09
Added section for cookie samesite attribute

simmayor:


iliakan:
Good point. 

Please provide a description of the attribute and its possible values, right now a reader may wonder, what are possible values, and what they mean exactly.

simmayor:
I added the attribute's values and a list of types of requests that they block.

iliakan:
Much better now, the attribute is new, good to know, thanks!

PR #722, opened 2019-01-10
remove the question with \n (solutions)

maurodibert:


iliakan:
removed both

maurodibert:
Great Ilya! It's a pleasure to help even a bit with this incredible job! Thanks for your work man!

iliakan:
Thank you too =)

PR #723, opened 2019-01-10
Using negation over ||

maurodibert:
Since logical operators are explained in the next chapter, maybe could be better if here the solution uses negation, which takes advantage of all the great details that you have explained about truthy and falsy values!


maurodibert:
> Since logical operators are explained in the next chapter, maybe could be better if here the solution uses negation, which takes advantage of all the great details that you have explained about truthy and falsy values!

PR #727, opened 2019-01-11
Update task with a hint or another exercise/solution.md

maurodibert:
Logical Operators are explained in the next chapter. I proposed the possibility of another hint for that here from your tutorial but maybe could be better to move the exercise one chapter.


iliakan:
I'll add more details here.

PR #732, opened 2019-01-14
Update article.md

maurodibert:
"Logical operators" is repeated and even the "comma" subject it was not mentioned in that chapter but in operators. Could be better and more coherent with the order of the tutorial maybe?


iliakan:
Fixed thanks.

PR #735, opened 2019-01-14
Update solution.md

maurodibert:
indenting


iliakan:
What's wrong there? Two spaces, as before...

maurodibert:
Sorry! I've understood that since is nested, should have the same two spaces from the else, and four from the begining.

PR #736, opened 2019-01-14
typo "tends (to look) uniform"

11un:
was "tends looks uniform"


iliakan:
"the code tends to look uniform" - what's wrong here?

11un:
it was "code tends looks uniform" and I changed it to "tends to look uniform" as I'm unsure what "tends looks" and assumed "tends to look" is what was meant

11un:
> Please tell me, are you a native speaker? If you are (I'm not), let me know, maybe I should rephrase that.

Yes I am native English speaker from USA. To try to explain again: Basically "tends looks" is two plurals in English language. This is not proper way to state this. But "tends to look" is. Essentially saying something like "it commonly looks a certain way". 

On a side note: I do think your English skills are very good, just trying to help out a bit here as I appreciate the work you've done on this.. :)

PR #740, opened 2019-01-17
Update article.md

JuliaPoberezhnaya:
Is it right, that here "Returning promises allows us to build chains of asynchronous actions."? I think, hier should stay "synchronous actions."


iliakan:
No, async actions.

PR #746, opened 2019-01-21
arr.__proto__.constructor

Kudrik:
It will help people not to confuse where constructor property is stored in objects


iliakan:
Properties are inherited, so naturally it's like that.

Maybe write it like this: `arr.constructor (= arr.__proto__.constructor) = ...`?

PR #749, opened 2019-01-22
update camelcase solution

11un:
Added solution from sandbox,  (```js run ... ``` ) <-- I think this is correct for how you integrate this functionality? If not let me know if I need to change it and for future reference. Thanks.


iliakan:
Indentation looks wrong (not spaces? too much spaces?)

PR #759, opened 2019-01-27
Update solution.md

aman566:
we can use splice() also to add "Classics" into the array


iliakan:
Sorry, invalid js in runnable code.

PR #766, opened 2019-01-28
add sum-salaries solution

11un:
Added solution from sandbox and one-liner option using reduce.

### DIFF
--- a/1-js/05-data-types/08-keys-values-entries/01-sum-salaries/solution.md
+++ b/1-js/05-data-types/08-keys-values-entries/01-sum-salaries/solution.md
@@ -1,0 +1,14 @@
+```js run
+    function sumSalaries(salaries) {
+    
+      let sum = 0;
+      for (let salary of Object.values(salaries)) {
+        sum += salary;
+      }
+    
+      return sum; // 650
+    }
+```
+Or, optionally, we could also get the sum using `Object.values` and `reduce`:
+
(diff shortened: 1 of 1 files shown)


iliakan:
Please remove extra indentation from inside ```.

Also, Object.values and reduce are explained at this point, you sure? :)

11un:
> Please remove extra indentation from inside ```.
Ok sorry, I need to figure where my formatting is incorrect when doing these. I've tried to replicate the other example solutions exactly. It may be related to my env / tools linux / vim / eslint / prettier etc. I'll look into it more soon and try again. 

> Also, Object.values and reduce are explained at this point, you sure? :)

Yes, Object.values, entries, etc. was in previous chapter (https://javascript.info/keys-values-entries) and used in the preexisting destructuring assignment example with the for.. of loop.

Also yes, reduce/reduce right is covered a few chapters back in the Array methods.. (https://javascript.info/array-methods)

iliakan:
Thanks, actually extra indent is not a problem, scripts handle it well.

But it's really unnecessary =)

11un:
> Thanks, actually extra indent is not a problem, scripts handle it well.
> 
> But it's really unnecessary =)

Fixed it, and for moving forward... It was a md/eslint config lint-on-save setting used for an unrelated use-case and I also don't like the extra indentation in most cases. Here is the fixed: https://github.com/iliakan/javascript-tutorial-en/pull/772

PR #774, opened 2019-01-31
solution out of tutorial

maurodibert:
I've took the text and update the task file also!

### DIFF
--- a/1-js/05-data-types/05-array-methods/6-calculator-extendable/solution.md
+++ b/1-js/05-data-types/05-array-methods/6-calculator-extendable/solution.md
@@ -1,3 +1,39 @@
+``` js run
 
-- Please note how methods are stored. They are simply added to the internal object.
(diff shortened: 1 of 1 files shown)


iliakan:
Why copy-pasting the solution from sandbox?

maurodibert:
Because every other exercise has the solution on the same tutorial and it's really annoying, for instance, if you read the tutorial in a Kindle or you have not the possibility of opening the sandbox, to not see the solution.

iliakan:
Then we need to change the code that represents the solution in e-book, not copy-paste it around.

maurodibert:
Oh, sorry! I didn't think like that. First, I've just copy/paste not for laziness, hehe, but because the solution is correct and it's like other exercises (that I've suggested a copy/paste from sandbox).

Again, apologies if I do pull stupid suggestions but I'm really a newbie and want to contribute to your work that I love and think it's hell awesome!

PR #783, opened 2019-02-03
Solution out of tutorial

maurodibert:
Hi! Same case as before. I keep thinking it's useful to have solutions on the same web page. Since you've closed the last one, I'm not sure if I'm explaining the whys correctly.

### DIFF
--- a/1-js/05-data-types/07-map-set-weakmap-weakset/01-array-unique-map/solution.md
+++ b/1-js/05-data-types/07-map-set-weakmap-weakset/01-array-unique-map/solution.md
@@ -1,0 +1,11 @@
+``` js run
+function unique(arr) {
+  return Array.from(new Set(arr));
+}
+
+let values = ["Hare", "Krishna", "Hare", "Krishna",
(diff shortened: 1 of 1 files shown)


iliakan:
Did you copy-paste it?

maurodibert:
Hi! I use the solution and added the first part. Which would be the correct procedure?

iliakan:
What's the purpose of copying the solution? To avoid extra click to sandbox?

maurodibert:
It does not just avoid the extra click but there are situations and devices that don't allow the sandbox. And beyond that, I really don't see consistency with some exercises in the sandbox and the whole rest in the tutorial. Why isn't everything in just one place and the one you're working on?

iliakan:
@bezart please take a look.

iliakan:
I'll try to embed plunker.

maurodibert:
Ilian! Great! And now you said about embed plunker, I've understood what you've said about copy/paste! 🤦‍♂

PR #789, opened 2019-02-06
Removes the confusion from user readability point

jyoti007:


iliakan:
What was wrong here? Also, the word "check" seems to be excessive.

jyoti007:
Hello,

There is nothing wrong. Only I have added some more text for the readers to
make them understand easily.
Jyoti Gupta

iliakan:
Sorry, but the new phrase sounds a bit odd.
"Comparisons check `> < >= <=`  convert `null` to a number..."

Maybe "comparisons checking `> < ...`".

PR #796, opened 2019-02-12
Update solution.md

Garnik8:


iliakan:
Not everyone knows what a "natural" number is. Hence the wording.

Garnik8:
in your solution you doesn't prevent from giving float values to n. And in that case it gives wrong result

PR #809, opened 2019-02-15
"use strict" directive is missing in code example

ndtxt:
but it's supposed to be there, according description below the code.


iliakan:
It's assumed everywhere.

PR #818, opened 2019-02-25
add task/solution to "Extended Clock"

11un:
Added the code from sandbox files here. While I tested this was also wondering about this statement:

> Create a new class ExtendedClock that inherits from Clock and adds the parameter precision -- the number of ms between "ticks". Should be 1000 (1 second) by default.

Should this be changed to 10000? 

In the solution / sandbox it is set to 10000 at least (and shows every ten seconds accordingly):

```
let lowResolutionClock = new ExtendedClock({
      template: 'h:m:s',
      precision: 10000
    });

```

https://next.plnkr.co/edit/ArMtHx1QBohRahJxlOJz?p=preview

### DIFF
--- a/1-js/07-object-oriented-programming/10-class-inheritance/2-clock-class-extended/solution.md
+++ b/1-js/07-object-oriented-programming/10-class-inheritance/2-clock-class-extended/solution.md
@@ -1,0 +1,24 @@
+```js run
+// extended-clock.js
+
+class ExtendedClock extends Clock {
+  constructor(options) {
+    super(options);
+    let { precision=1000 } = options;
+    this._precision = precision;
+  }
+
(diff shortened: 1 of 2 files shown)


iliakan:
Ok, 10000, but why are these changes that you did?

11un:
> Ok, 10000, but why are these changes that you did?

I just added the code that didn't exist on the site itself (from the sandbox) to the task/solution as is. When testing I notice the 10000ms vs the description "1000 (1 second)" which was different.. and still is.

iliakan:
The default value is 1000, I just checked.

Please don't copy-paste the code around, the tutorial md has enough power to insert the code by src =)

Fixed this, thanks.

PR #819, opened 2019-02-25
add new task (for review)

11un:
I created a new directory, task and solution for the "class checking: instanceof" chapter since there was only one task and I really liked this this "advanced typeof / instanceof" section. Anyway if you like it cool, if not, let me know if I should/could add more, change it or whatever etc and I will. Thank you.


iliakan:
The tasks are really welcome, thank you for the great addition.

Although, that exactly task is a little bit not we want.

These details of `toString.call` are not important to know by heart. When a task has `importance: 3`, it must be something very interesting to put it in.

PR #826, opened 2019-03-04
Incorrect way of dealing with imprecise calculations

pobch:
I think one of the methods recommended in **Imprecise calculations** section is incorrect.
For example
```
( (0.28*100) + (0.14*100) ) / 100 === 0.42 // false
```


iliakan:
Great notice! Thank you, the text was updated.

PR #830, opened 2019-03-05
Update solution.md

maurodibert:
Hi Ilya! I proposed this because there's a lot of people studying the tutorial and testing with Chrome Console and the result is not `null` but `[object window]`; also because it was the proposed tool for start the tutorial.

Could be useful (if you did not that in another further place), to explain the specifics of `'strict mode'` and `f.bind(context)` maybe?


iliakan:
A good addition, I put it into the "first steps", when use strict is introduced.

PR #835, opened 2019-03-07
Update article.md

maurodibert:
The tutorial has not explained `try/catch` yet. Maybe, this part of generators could be as part of the `try/catch` article. It's extensive but the section has only those two articles about errors so it gets balanced


iliakan:
Good point! 
Actually, I had troubles finding the proper location for generators. They are not needed there, not used in further chapters.

Moved generators to a separate section now, see title page.

PR #836, opened 2019-03-07
Update article.md

maurodibert:
Moving to the very first place that occurred an added a little bit more of information about.

### DIFF
--- a/1-js/07-object-properties/01-property-descriptors/article.md
+++ b/1-js/07-object-properties/01-property-descriptors/article.md
@@ -122,6 +122,10 @@ user.name = "Pete"; // Error: Cannot assign to read only property 'name'...
 
 Now no one can change the name of our user, unless they apply their own `defineProperty` to override ours.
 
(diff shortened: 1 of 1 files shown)


iliakan:
No, sorry, that should be after. That's like P.S.

We should be using strict all the time.

maurodibert:
Hi Ilya! But P.S. it's at the end of the section?

iliakan:
it's right at the end of the article section, before ## Object.defineProperties

maurodibert:
Yeap! I saw there but what I wanted to say is that if it's mandatory for PS´s, to go at the end. PS = info boxes. Because if not, when you first try the exercise, do not know why it doesn't throw an error.

Oh, and would be great to explain on the 'strict mode' chapter, what happens with chrome console.

iliakan:
I added use strict for console there.

maurodibert:
Ahhh! Understood now! 👍  Awesome!

PR #838, opened 2019-03-07
Update solution.md

Slick3gz:
Sort by objects 'name' property


iliakan:
Should sort by `age`.

PR #847, opened 2019-03-10
Replace old, broken chapter link with new chapters

georgesolomos:
This can be reworded if you want but I think it's fairly clear


iliakan:
Thanks!

PR #848, opened 2019-03-10
Fix Typo

yifeiyin:
ley -> key

This should also make it have the correct style.


yifeiyin:
### About The Second Commit: 7f9851e

Should the spaces be added after `for` for **in-line text** might be arguable, (since `for(;;)` does not have a space). I added them. Let me know if I should do anything different.

PR #862, opened 2019-03-19
Update 07-modifying-document

rlax:
Fix empty plunkr viewbox window

Plunkr preview will be empty if there is no `<div></div>` (or other element besides `<script/>` or `<style/>`). I guess this fix can be applied to all plunkrs code snippets in this repo.

### DIFF
--- a/2-ui/1-document/07-modifying-document/article.md
+++ b/2-ui/1-document/07-modifying-document/article.md
@@ -290,7 +290,7 @@ So here's an alternative variant of showing a message:
   background-color: #dff0d8;
 }
 </style>
-
+<div></div>
 <script>
   document.body.insertAdjacentHTML("afterbegin", `<div class="alert alert-success">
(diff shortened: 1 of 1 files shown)


iliakan:
What's wrong with it being empty?

rlax:
Doublechecked it. Plunkr preview will be empty without `div`.

image

iliakan:
I recheked current page in Chrome.

Both examples show up fine:

image

image

rlax:
It works fine only for a couple of a days now, i guess someone changed the plunkr wrapping code. I think that adds `<html>...<body>...</body>...` to the script thus it works render correctly now.  Plunkr itself will still render nothing if this `body` tag wrap is removed (as it was previously 12 days ago).

 Guess we're done here, thanks.

iliakan:
Oh thank you, there was a change there indeed.

PR #869, opened 2019-03-23
Cleaned up the paragraph about commas in classes

lostwong:
There was some ambiguity in the paragraph that I removed.  Verified that commas between class methods do result in syntax errors in my own code.


iliakan:
"Please note that methods in a class are not permitted and will result in a syntax error." - what? :)

PR #884, opened 2019-03-31
Function declarations are not scoped with if else

orkhan-huseyn:
Function declarations (just like `var`s) are not scoped with if-else blocks. So function which is declared inside if-else block will be visible outside of that block. We can only make function declarations private by wrapping them with a function.


iliakan:
wrong

PR #913, opened 2019-04-14
fix typo in for-in loop

Azatik1000:
I think there should be "let"


iliakan:
No, here is an example of pure "for".

PR #916, opened 2019-04-15
Update article.md

misterabd:
Fixed typos and made grammar changes.


1-js/11-async/08-async-await/article.md
```diff
@@ -250,7 +250,7 @@ async function f() {
 f();
 ```
 
-If we don't have `try..catch`, then the promise generated by the call of the async function `f()` becomes rejected. We can append `.catch` to handle it:
```
iliakan: looks wrong now

PR #917, opened 2019-04-16
Smooth over some wording

dragonwocky:
I've generally just attempted to make a couple parts make a little more sense by improving grammar, removing extra punctuation, and fixing a couple of typos.


iliakan:
Thanks!

PR #932, opened 2019-04-21
Updated article.md (better sentence flow)

cmdlhz:
BEFORE
"3. When this happens, a special “object wrapper” IS CREATED that provides the extra functionality, and then is destroyed."
AFTER
"3. When this happens, a special “object wrapper” that provides the extra functionality IS CREATED, and then is destroyed."

https://github.com/javascript-tutorial/en.javascript.info/blob/master/1-js/05-data-types/01-primitives-methods/article.md


iliakan:
thanks!

PR #952, opened 2019-04-30
replace "press" with "click"

pimonenkov:
I suggest replacing "press" with "click"

"press" is when we push a keyboard key
"click"/"select" is when we use a mouse button.


iliakan:
Actually, people can press "CANCEL" using keyboard "tab" =)
But if it makes things more clear..

PR #955, opened 2019-04-30
Let's avoid ninja_code patterns

F3PiX:
Previously, in https://javascript.info/ninja-code#soar-high-be-abstract, naming variables like `str2` is ninja code. Here's a leftover.
(The `type` name ninja-ism is not a problem here, because `string` can be seen as a description.)


iliakan:
maybe `otherString`, but not `other_string` =)

PR #966, opened 2019-05-07
Set correct prototype in User

kryptoatom:
Added one line of code to show modification of `this` also includes setting correct prototype. This change might prevent users from running into scenarios like below.

```
function User(name, age){
    return {
        name,
        age,
     }
}

User.prototype.display = function (){
    console.log(`${this.name} is ${this.age} years old.`)    
}

let john = new User("John", 23)

john.display()  //TypeError: john.display is not a function
// It failed because User is not a prototype of john.
```


iliakan:
We didn't study prototype yet at this point.

PR #975, opened 2019-05-10
Update article.md (LocalStorage)

gongzemin:
change If we any other type to If we want any other type, add a word -- want.


lex111:
@gongzemin thank you!

PR #979, opened 2019-05-14
Order mismatch

him2312:
Answer is 1 and 3.

Whereas in the solution 1 and 2 show the same result due to different order.


him2312:
Screenshot 2019-05-18 at 11 41 15 PM

iliakan:
thanks

PR #980, opened 2019-05-15
Fix typo (brewity to brevity)

gongzemin:


lex111:
@gongzemin thank you!

PR #982, opened 2019-05-15
Update code in task Function in If - Closure lesson

quangnd:
To fit in the solution content `The result is an error.` We need `use strict`, if not the plunker still run normally. This will cause confusion with the student.

Here is the link [https://javascript.info/closure#function-in-if](https://javascript.info/closure#function-in-if)

We also can add a note in solution text about remove `use strict`, the program will run normally.


iliakan:
use strict is assumed for all examples in the book

PR #988, opened 2019-05-16
article.md summary clarification

AndrewPanov:
1) JavaScript provides many ways to create an object for sure, not class. I don't even think that there is way 'to create class'.
2) 
function User(name) {
  this.sayHi = function() {
    alert(name);
  }
}
That code looks more like closure concept, not like a class one. There is no 'field' (name), but single method that gets parameter from main code block and alerts it.


iliakan:
clarified! thanks

PR #992, opened 2019-05-18
Update task.md

maurodibert:
I think it's better this way because if you read the task fast, could miss that you ask for the two ways which are the practice to do in order to really understand.

And Hi! hehe, I'm back! I was studying Rails and now Ember for my first job as a developer, so had to stop studying JS. Now I'm reviewing everything and also start doing free classes for those who want to start coding, so, back into the good stuff! hehe.

Thanks, Ilya again!


iliakan:
Sorry, but the change is wrong.

PR #994, opened 2019-05-18
adding a small clarifying comment

boazrymland:
The first description of 'children' doesn't clarify clearly, in my taste that 'children' are kind of a subset of the entire child nodes of the node. Adding a small comment that states this, and refers the user to read more below).


iliakan:
no, text nodes are child nodes as well.

PR #1001, opened 2019-05-20
removed misplaced ```

AM-77:


lex111:
Thank you, but in this case there is no error, these symbols are needed (see [screenshot](http://joxi.net/n2YejzqFbywWQ2)).

PR #1018, opened 2019-05-26
Update task.md

maurodibert:
I'm pretty sure that if this work, it's hardcoded! (LoL) It's that the solution button must be in the next line...

### DIFF
--- a/1-js/05-data-types/04-array/5-array-input-sum/task.md
+++ b/1-js/05-data-types/04-array/5-array-input-sum/task.md
@@ -12,4 +12,4 @@ Write the function `sumInput()` that:
 
 P.S. A zero `0` is a valid number, please don't stop the input on zero.
 
-[demo]
+[demo]  


iliakan:
Wait, what did you change? Added spaces?

maurodibert:
Yeap!

And have this question: I'm doing the translation of this article (https://javascript.info/variables), and github doesn't render the `smart header`. There's a special way of doing it?

iliakan:
Spaces change something? They shouldn't.

Github doesn't render some site-specific things, that's ok.

maurodibert:
Reading on Markdown docs, two spaces is equivalent to add a new blank line but I thought probably didn't work and if did, hardcoded.

Sorry about that and thanks for the explanation!

iliakan:
Is something wrong there? Please tell me.

iliakan:
What's the purpose of the fix?

maurodibert:
Yeap, at least for me visually and the consistency with the other articles:: see that "Run the demo" it's "glue" with "solution", instead of two lines, one line each button.

iliakan:
Oh, is it ok now?

PR #1025, opened 2019-05-29
const documents lack of later reassignment, it is not to enforce lack…

graingert:
… of reassignment


graingert:
eg, const isn't a promise not to reassign - it's documentation that the variable is never reassigned.

iliakan:
That should be clear from the text. I mean, the program may be updated to reflect any changes, so the phrase doesn't add anything new.

PR #1052, opened 2019-06-13
Create task.md

maurodibert:


iliakan:
I'm sorry, but the task is too vague. You ask the user a theoretical question. That's not a task.

maurodibert:
Completely true! Is that when I wanted to create a proper task, saw that there was not too much to do! hehe. Thank Ilya! I'll continue with the others!

PR #1065, opened 2019-06-18
Fixed code block

DouglasMV:
Fixed code block by adding an extra `
It was causing all text after it to be inside a code block (at least when viewing in github) for some reason.


iliakan:
Please send a screenshot of what's wrong?

For me all ok:
image

DouglasMV:
At the website it is ok for me too, but at github it is not:

Screenshot from 2019-06-19 10-03-06

iliakan:
That's fine.

PR #1095, opened 2019-06-29
Added info about use strict

ArtemTropanets:
Function Declaration is not visible outside of block only by use strict.
Could you please also check if 'use strict' is enabled generally by script examples. When I run scripts on any page it works like without strict mode.


iliakan:
fixed in 46a9bb5e34a1c0003e1c2367599c2f4dae969c7a

PR #1110, opened 2019-07-02
Update article.md

maurodibert:
Hi Ilya! I'm not sure if there is any other intention using a function User() instead of the class, but since you were comparing, maybe could be an error.

### DIFF
--- a/1-js/09-classes/03-static-properties-methods/article.md
+++ b/1-js/09-classes/03-static-properties-methods/article.md
@@ -20,7 +20,7 @@ User.staticMethod(); // true
 That actually does the same as assigning it as a function property:
 
 ```js
-function User() { }
+class User { }
 
 User.staticMethod = function() {
   alert(this === User);


iliakan:
Why did you close this?

maurodibert:
Hi! Because I was wrong ;-)

iliakan:
As we're talking about classes here, I'd agree to use `class` instead.

iliakan:
Seems better, no?

maurodibert:
I thought that the first time but then didn't know if you wanted to use functions for any other internal specification. i'll restore the branch and commit! ;-)

iliakan:
I changed one more word there and committed locally, thanks!

maurodibert:
Cool! Thanks to you one more time!

maurodibert:
And when I have a little bit of time, I will start doing these things locally and pushing those exercises (and others) that I couldn't here! Have a nice day Ilya!

iliakan:
By the way, I requested changes for #1053

Did you see that?

maurodibert:
hehe yeap! I was answering you that in the previous comment! But I kept the branch not to forget!

PR #1116, opened 2019-07-03
Update article.md

ArtemTropanets:
Info about what return push and unshift methods.


iliakan:
I'd suggest maybe adding another example about this, but not merging two things into one code.

People 99% of time don't use returned value for push/unshift.

PR #1123, opened 2019-07-07
[strings] 2 code units are 4 bytes

aruseni:


iliakan:
This was changed.

aruseni:
Now it is “taking up to 4 bytes”. But that isn’t correct: if you use 2 code units, that is 4 bytes.

PR #1125, opened 2019-07-07
[strings] localeCompare

aruseni:
The specification only mandates negative and positive values.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare


iliakan:
The rules are same as for strings. It just uses locale.

aruseni:
**DO NOT rely on exact return values of -1 or 1.** Negative and positive integer results vary between browsers (as well as between browser versions) because the W3C specification only mandates negative and positive values. Some browsers may return -2 or 2 or even some other negative or positive value.

aruseni:
Please see my link from this PR description.

PR #1126, opened 2019-07-08
Modify syntax error on task section

Violet-Bora-Lee:


iliakan:
What was wrong?

Violet-Bora-Lee:
I should've uploaded the image but I couldn't. You can check the broken syntax on the task section of this article.
http://javascript.info/comparison

Violet-Bora-Lee:
KakaoTalk_20190709_001152282

iliakan:
Sorry, I may be just not getting something. Can't see what's wrong :(

Violet-Bora-Lee:
In order to highlight something in the code box, you are using `*!*text to highlight*/!*` not  `"\n text to highlight \n"`. Isn't it your intention that comparing `null` and `0` in the task? I may be over-thinking.

iliakan:
The intention is to compare `null` against the given string.

Violet-Bora-Lee:
Oh i see now. Ill close the pr

PR #1129, opened 2019-07-08
Remove spaces

aruseni:


aruseni:
Just noticed that there were 3 spaces added on purpose for aligning. But anyway it’s better to remove that, because otherwise we’d ideally need to align other code the same way, where applicable, in order to have consistency in style.

PR #1136, opened 2019-07-10
[strings] 2 code units are 4 bytes

aruseni:
Now it is “taking up to 4 bytes”. But that isn’t correct: if you use 2 code units, that is 4 bytes.


iliakan:
You made PR to old version of the article.

aruseni:
Yeah, because I just repeated the old PR as you closed the other one earlier. :)

aruseni:
Anyway, thanks for resolve!

PR #1142, opened 2019-07-11
Fixed condition (n < 0) → (n <= 0)

decorator-factory:
The condition (n<0) allowed for (n: n >= 0) which a) doesn't match the error message; b) doesn't match the condition in the task (n<=0).
 (Also: maybe all conditions should be changed to n < 0, since x<sup>0</sup> still makes sense and is equal to 1? Anyway, it's not as important as being consistent)


iliakan:
power 0 is actually supported, maybe fix message instead?

PR #1150, opened 2019-07-15
Add stylistic and grammar fixes

Doesntmeananything:
I fixed some of the awkward phrasing:

1. `concrete` is more common in the literary style of writing and is used to denote something opposite of abstract, so placing it here when describing a relation between html elements seems unnecessarily cumbersome. Plus there's some tautology in using `concrete` together with `actually`. I propose something simpler.

2. `It’s possible that event.target equals this` -- the present indefinite tense is not quite correctly used here, because this sentence describes something that only happens conditionally and is not always true. There are many ways to rewrite this sentence, and I propose one that aims to keep the original structure of the idea.


iliakan:
Thank you very much for correcting my poor English!

PR #1152, opened 2019-07-15
Corrected no-declaration object destructuring

caproven:
Close parenthesis was in wrong location, should have been at end of the expression. Page information pertaining to this is still accurate, so the only change is the code block.

As per [this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Assignment_without_declaration) info, modified to follow:
```
let a, b;
({a, b} = {a: "Apple", b:5});

alert(a); // Apple
alert(b); // 5
```


iliakan:
Thank you!

PR #1159, opened 2019-07-17
Exclude negative numbers from the result

1van0v:


1van0v:
From my point of view this task is a bit complicated.
I just sugested a solution that feets the task description.

iliakan:
We can't use features that we didn't teach yet.

iliakan:
Maybe you could suggest another solution instead? 

Thank you!

1van0v:
As an alternative option I can suggest the following:

`"1.5 0 -5 12. 123.4.".replace(/-\d+(\.\d+)?/, ' ').match(/[1-9]\d*(\.\d+)?/g)`

However, the above solution uses two regular expressions and it does not seem to me that it is an optimal one.

iliakan:
Moved it to another section: please check it out
http://javascript.info/task/find-non-negative-integers

PR #1173, opened 2019-07-21
Reorder article section

lewislbr:
I think reordering this section helps with the understanding of the topic because:

1- The section starts explaining that "For loops" are the most commonly used loop, so it should be the first to learn.

2- The section introduces topics in more detail that are also used in the other sections, like the parts of the loop.

I also removed a sentence related to the previous order to not cause confusion. If you think it's still relevant and needs to be rewritten I'm happy to edit it.

### DIFF
--- a/1-js/02-first-steps/12-while-for/article.md
+++ b/1-js/02-first-steps/12-while-for/article.md
@@ -6,82 +6,6 @@ For example, outputting goods from a list one after another or just running the 
 
 *Loops* are a way to repeat the same code multiple times.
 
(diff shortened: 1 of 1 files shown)


iliakan:
The "for" loop is later on, because it's the most difficult one. I guess, it should be there, as we go from simpler things to harder ones.

That said, did I miss anything else?

lewislbr:
You are right, it makes more sense to keep it after the easier ones, even if it's the most common one. I just got it wrong the first times reading through it.

PR #1178, opened 2019-07-23
Fixed the solution to work in strict mode

AdarshNaidu:
**Changed from** (_this does not work in strict mode_)
`result = (a + b < 4) ? 'Below' : 'Over';`
**to**
`let result = (a + b < 4) ? 'Below' : 'Over';`


iliakan:
It's not runnable code.

PR #1187, opened 2019-07-28
element.closest support

roblevintennis:
Adds a small amount of information re element.closest support


iliakan:
That's true for so many advanced stuff. We can't mention IE everywhere =)

roblevintennis:
Gotcha, yeah, I suppose that's fair 👍 thanks for reply!

PR #1194, opened 2019-07-31
Update article.md

Violet-Bora-Lee:


Violet-Bora-Lee:
When the multiple tasks run "right after" another, it is synchronous tasks.

iliakan:
If we schedule something to run after something, that's async.

iliakan:
The key word is not "right now", but 'right after the code is complete'. 

Still, if the phrase confuses you, maybe we should remove it?

Violet-Bora-Lee:
Thanks for your comment. Yeah, there could be ppl like me I guess.

iliakan:
removed this phrase

PR #1197, opened 2019-07-31
Update task.md

lumosmind:
without 'use strict'  this keyword will be equal to global window object. So 3th and 4th will be like this
(method = obj.go)();    // (3) [object Object]
(obj.go || obj.stop)(); // (4) [object Object]


iliakan:
All code is use strict by default.

lumosmind:
> All code is use strict by default.

But in plnkr sandbox which attached javascript.info there isn't default 'use strict', so sandbox giving incorrect results. I think that 'use strict' doesn't be default to understand differences between normal mode and strict mode.

PR #1212, opened 2019-08-03
Problem with rounding negative number to n-th digit after the decimal.

lumosmind:
Math.floor(-123.456) gives us -124 but we want to get -123.


iliakan:
That's ok: floor rounds down.

PR #1215, opened 2019-08-05
paste negation

dagolinuxoid:
Regardless Object.keys and other methods: 
Properties from the prototype are taken into account. -> Properties from the prototype are NOT taken into account.


dagolinuxoid:
It looks like very soon eventually there will be no mistakes in the wonderful resource. ( Until new major updates of course :) )

iliakan:
Thanks a lot!

PR #1216, opened 2019-08-05
Update article.md

dshuran:
The explanation about .`then` returned value is wrong in the article. It has really confused me, so I've searched for the docs and all's become clear. `.then` always returns a promise, it's really important, especially when speaking about typescript. My pull-request isn't the best, so maybe it should be edited some way. 

You can check the docs - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/then#Return_value


iliakan:
Improved explanations.

PR #1231, opened 2019-08-09
add example for function working in strict mode

Violet-Bora-Lee:


iliakan:
We didn't study functions yet, so I'm hesitating to provide such code.

Violet-Bora-Lee:
Ah, I see :)
Your opinion makes sense.
The reason I suggested this code is that Someone could code like this to enable strict mode.

```
'use strict';
function strict() { 
}
```

I did actually. lol

PR #1233, opened 2019-08-09
add explanation for function working in strict mode

Violet-Bora-Lee:
I saw your opinion.

The reason I request this is that someone could think like this to enable strict mode.

```
'use strict';
function strict() { 
}
```

I did actually.

I know the term, function body, is not mentioned yet at the moment.

I hope there could be some ways to reduce misunderstanding.


iliakan:
Changed the wording to reflect it.

Violet-Bora-Lee:
Thanks 👍

PR #1261, opened 2019-08-21
Update article.md

alevmo:
Added some changes to improve readability.

### DIFF
--- a/1-js/02-first-steps/07-operators/article.md
+++ b/1-js/02-first-steps/07-operators/article.md
@@ -26,7 +26,7 @@ Before we move on, let's grasp some common terminology.
     alert( y - x ); // 2, binary minus subtracts values
     ```
 
-    Formally, we're talking about two different operators here: the unary negation (single operand: reverses the sign) and the binary subtraction (two operands: subtracts).
+    Strictly speaking, we've considered two different operators that share the same symbol: the former, known as the negation operator, is a unary operator; the later, known as the substraction operator, is a binary operator. 
 
 ## String concatenation, binary +
 


iliakan:
May I wonder, are you a native speaker?

iliakan:
P.S. substraction is a typo (extra "s")

alevmo:
No, I am not, but I pay careful attention to what I read. That is why,
whenever I find a sentence that sounds odd, I try to reformulate it into
something with better readability.

iliakan:
I'm just not sure the new wording is much better.

@paroche , what you think?

paroche:
I think the suggested version is a little better, but there was nothing about the original that caught my eye as awkward when I first encountered it, or when I look at it now. If we wanted to keep drilling on this (and I think there are bigger fish to fry), I might suggest a third, somewhat hybrid, possibility:

"Formally, in the examples above we have two different operators that share the same symbol: the negation operator, a unary operator that reverses the sign, and the subtraction operator, a binary operator that subtracts one number from another." 
Or, we could belabor the intuitively obvious a bit, since we're being "formal":
"Formally, in the examples above we have two different operators that share the same symbol: the negation operator, a unary operator that reverses the sign, and the subtraction operator, a binary operator that subtracts the second number from the first." 



|  Formally, we're talking about two different operators here: the unary negation (single operand: reverses the sign) and the binary subtraction (two operands: subtracts). |
|  |  |  |  Strictly speaking, we've considered two different operators that share the same symbol: the former, known as the negation operator, is a unary operator; the later, known as the substraction operator, is a binary operator.  |



-----Original Message-----

PR #1267, opened 2019-08-23
Add note about modules enabling strict mode

spacejack:
I was using this book to train a junior programmer about JavaScript (which I think is a great resource by the way!) We no longer need to support IE11 for our projects and are using modules by default, so the chapter on `"use strict"` caused a bit of confusion. There was a brief mention of modules and classes enabling strict mode, but it was at the end of the page and I think it should be made clearer near the beginning of the article.


iliakan:
There's a note about that:
     Several language features, like "classes" and "modules", enable strict mode automatically.

spacejack:
Yes, I did note that. But it's buried at the end of the article. I think if you're using modules it should be more obvious that the stuff about 'use strict' doesn't apply.

iliakan:
In the beginning of the tutorial, the very notion of modules is new. We shouldn’t draw too much attention to that.

 It’s too early. People can see the relevant section about modules

PR #1275, opened 2019-08-27
Update article.md

paroche:
A draft of a proposed notation explanation w/ respect to "str" (as well as "num" and "obj"). It's probably not formatted as well as it might be, but the wording might be good as it is or could provide a starting point for discussion.


iliakan:
Closing this in favor of individual notes in every place.

PR #1276, opened 2019-08-27
clarify var declaration

oizabaiye:
In 02-first-steps > 04-variables > article.md:

Edited text to say that the statement `let message` declares a variable, while the subsequent use of assignment operator defines the variable.

### DIFF
--- a/1-js/02-first-steps/04-variables/article.md
+++ b/1-js/02-first-steps/04-variables/article.md
@@ -12,13 +12,13 @@ A [variable](https://en.wikipedia.org/wiki/Variable_(computer_science)) is a "na
 
 To create a variable in JavaScript, use the `let` keyword.
 
-The statement below creates (in other words: *declares* or *defines*) a variable with the name "message":
+The statement below creates (in other words: *declares*) a variable with the name "message":
 
 ```js
 let message;
 ```
 
-Now, we can put some data into it by using the assignment operator `=`:
+Now, we can put some data into it by using the assignment operator `=`: 
 
 ```js
 let message;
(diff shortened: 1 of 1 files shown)


iliakan:
Maybe let's just use "declare" everywhere?

Variable is declared with `let/var/const`. And don't use "define" with respect to vars/functions.

iliakan:
That's wrong. Assigning does not define the variable. 

> Assigning a value to the variable is how we *define* the variable. This is precisely why a variable with no value assigned is considered [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined).

oizabaiye:
> That's wrong. Assigning does not define the variable.
> 
> > Assigning a value to the variable is how we _define_ the variable. This is precisely why a variable with no value assigned is considered [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined).
> 
> (Maybe that's how you feel, but that would confuse people, as the term "define" isn't used in this case)

How would you describe what defining a variable means? The sources I've seen seem to imply that assigning is generally the same thing as defining:
[Stack overflow](https://stackoverflow.com/questions/20822022/whats-the-difference-between-variable-definition-and-declaration-in-javascript)
But I'm ok with the recommendation to just use "declare".

iliakan:
As I get it, generally, "define a variable" is the same as "declare a variable". But when I look in the language specification https://tc39.github.io/ecma262/, I can't find the word "define" to be used this way. 

So I suggest to just use "declare".

iliakan:
fixed

PR #1313, opened 2019-09-03
Update task.md

paroche:
Was missing some words and was hard to understand. I made my best guess as to what it should say. But it also might be good to be explicit about whether ignored calls are supposed to count toward the delay. That they are, for this task, is made clear in the description of the following task (Throttle decoration), but if one hadn't yet looked at that one, they might not know which way the debounce task is supposed to work (the numbers used in the examples seemed to suggest that maybe ignored attempts did count, but it wasn't enough to clear up the ambiguity) . That's how it was for me, anyway.

### DIFF
--- a/1-js/06-advanced-functions/09-call-apply-decorators/03-debounce/task.md
+++ b/1-js/06-advanced-functions/09-call-apply-decorators/03-debounce/task.md
@@ -6,7 +6,7 @@ importance: 5
 
 The result of `debounce(f, ms)` decorator should be a wrapper that passes the call to `f` at maximum once per `ms` milliseconds.
 
-In other words, when we call a "debounced" function, it guarantees that all other future in the closest `ms` milliseconds will be ignored.
+In other words, when we call a "debounced" function, it guarantees that all future calls to the function made less than `ms` milliseconds after the previous call will be ignored.
 
 For instance:
 


iliakan:
Thank you very much, your changes are usually correct and may no need such a thorough description!

iliakan:
That is: a shorter one would be enough, if it's kind of obvious when you read your changes =)

paroche:
Thanks, but most of it -- everything after the first sentence -- is actually a suggestion for a different change. I could do a PR for that change if you agree with the premise.

iliakan:
Ignored calls are just ignored, they do not affect the delay - seems obvious enough.

paroche:
That's the way I'd THINK it would be supposed to work. But consider the example:

let f = debounce(alert, 1000);

f(1); // runs immediately
f(2); // ignored
setTimeout( () => f(3), 100); // ignored ( only 100 ms passed )
setTimeout( () => f(4), 1100); // runs
setTimeout( () => f(5), 1500); // ignored (less than 1000 ms from the last run)

The "f(4)" attempt is 1000 ms after the last ATTEMPT (even though it was ignored). That 1100 was used rather than 1000 could be seen as implying that that amount of time may have been necessary. Why not have f(4) at 1000? Would remove the ambiguity that at least one person experienced (:-).

Issue #1323, opened 2019-09-08
Prototypal Inheritance: "The value of this" part confusing


muradsofiyev:
https://javascript.info/prototype-inheritance#the-value-of-this
In this part you talk about how this work and in there you say "**That is actually a super-important thing, because we may have a big object with many methods and inherit from it. Then inherited objects can run its methods, and they will modify the state of these objects, not the big one**."
I think this is not good explanation. Your sentence like you say "this value give to us performance optimisation" but this is does not make sense.
In my opinion this sentence must be change or update new one because this sentence can be make confusing.

muradsofiyev:
And before close this issue please discuss about this sentence.

iliakan:
Discussed =)

PR #1324, opened 2019-09-08
My Issue (#1323) can be fix like this.

muradsofiyev:
I think old sentence can be change like this because old sentense can be make confusing and I think old sentence not true absolutely.

### DIFF
--- a/1-js/08-prototypes/01-prototype-inheritance/article.md
+++ b/1-js/08-prototypes/01-prototype-inheritance/article.md
@@ -211,7 +211,7 @@ The answer is simple: `this` is not affected by prototypes at all.
 
 So, the setter call `admin.fullName=` uses `admin` as `this`, not `user`.
 
-That is actually a super-important thing, because we may have a big object with many methods and inherit from it. Then inherited objects can run its methods, and they will modify the state of these objects, not the big one.
+That is actually a super-important thing, because inherited objects can run parent's methods, and they will modify the only state of inherited object not parent's state
 
 For instance, here `animal` represents a "method storage", and `rabbit` makes use of it.
 


muradsofiyev:
https://github.com/javascript-tutorial/en.javascript.info/issues/1323

iliakan:
It looks like you're not a native speaker. Hopefully, a native speaker can read and write back on this.

muradsofiyev:
Yeah I'm not native speaker but current sentence is not meaningful :d

iliakan:
Dear Native Speaker, please take a look at the sentence, what we can improve? ;)

paroche:
I jumped ahead and read that far in the article. What you have isn't bad. I may suggest a possible rephrasing below. I think it's actually a pretty obvious point, anyway (not for everyone, obviously, and I'm usually one of those people) -- that 'this' is defined by the context in which the function is run, in this case the calling objects, and the fact that the function comes from the prototype doesn't matter. This seems to me little different than if it came from the body of the script as in your examples in https://javascript.info/object-methods#this-is-not-bound  (which, as you may recall, I was just rereading yesterday), where also, it is written: "The rule is simple: if obj.f() is called, then this is obj during the call of f". 

Coming back to the actual text:

"That is actually a super-important thing, because we may have a big object with many methods and inherit from it. Then inherited objects can run its methods, and they will modify the state of these objects, not the big one."

I don't think it's bad as it is, but perhaps the second sentence could read, after the comma:
"..., and they will modify their own states, not those of the object that they are inheriting the methods from."
(BTW, don't believe anyone who tells you you can't end a sentence in English with a preposition. I could have had ", not those of the object from which they are inheriting the methods." It was a close call as to which I prefer. Could go either way).

iliakan:
@paroche Could you please suggest a PR how to put that right (or just say that there's no need)?

So we can close this. 

Thank you =)

P.S. I looked up the dictionary about `native speaker`: https://www.merriam-webster.com/dictionary/native%20speaker  Maybe at your area of living people just don't use these words, as they are, well, native speakers? =)

paroche:
It was the capital "N" on "Native" that I was responding to. As in "Native American". I'm in the American Southwest, surrounded by Pueblos (that are reservations. Sovereign nations, in theory, and, to some extent, in practice).

iliakan:
Oh, I see (making a note: "native speakers", never uppercase N") Thanks =)

paroche:
Well, I think I see what you were doing -- whimsically making it a proper name. Just my association, due to my location.

PR #1326, opened 2019-09-10
Update h to H char according ASCII value

HirenTariwala:
I have found that ASCII is not matching with char h so I have a changed.

h = 104
H = 72


iliakan:
Thanks =)

HirenTariwala:
very helpful tutorial (y)

PR #1333, opened 2019-09-12
Update solution.md

paroche:
Wording change. (I actually like the original wording "does not give understanding" -- very concise, and communicates clearly. It's just a bit non-standard in usage).


iliakan:
Great change, sounds much better!

paroche:
Very good.

PR #1337, opened 2019-09-13
minor suggestion

Violet-Bora-Lee:


Violet-Bora-Lee:
I made a suggestion because the value itself can be returned.

iliakan:
It becomes a little bit more difficult to understand. The "value to be replaced" sounds confusing.

PR #1340, opened 2019-09-13
Update article.md

paroche:
Wording adjustments in Summary lines "- Write/delete...", and "- The 'for..in'...".


iliakan:
Why "its"?:  `over its own and its inherited properties`

Is it really needed?

paroche:
It's intelligible the way it is, but I think it reads better with the possessive. "Over own and inherited" sounds a bit odd.

PR #1343, opened 2019-09-14
Update article.md

paroche:
Making clear that bound function does not hold the source function's custom properties.

### DIFF
--- a/1-js/06-advanced-functions/10-bind/article.md
+++ b/1-js/06-advanced-functions/10-bind/article.md
@@ -104,7 +104,7 @@ let boundFunc = func.bind(context);
 
 The result of `func.bind(context)` is a special function-like "exotic object", that is callable as function and transparently passes the call to `func` setting `this=context`.
 
-In other words, calling `boundFunc` is like `func` with fixed `this`.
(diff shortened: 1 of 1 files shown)


iliakan:
This is covered elsewhere, somewhere in decorators. 

Also, it's either obvious or needs a bigger example. Not sure if it's important.

paroche:
Since this is the first discussion of binding, I don't see how it could have been covered previously. And I did spend a lot of time on the decorators article. The word "custom" does not appear there. Still, maybe I missed it. In any case, I didn't know this until I read the solution to the "Function property after bind" task.

iliakan:
I guess, it should be obvious from the line "The result of `func.bind(context)` is a special function-like "exotic object", that is callable as function and transparently passes the call to `func` setting `this=context`."

paroche:
Well, all I can say is that it wasn't that obvious to me, though I wasn't at the time clearly remembering the "Function object NFE" article. But in general I think it's better to be more explicit rather than less (and this seems to be the philosophy you take much of the time -- one of the things I like in the tutorials. Makes it substantially easier to learn). Also I've noticed that the function name property of the original function, at least, DOES come to the bound exotic object (in my tests) as part of it's name property , e.g. boundMyFunc.name -> "bound myFunc". As I write about this and become more cognizant of the distinctions, it does come to be more apparent that the customer properties **probably** wouldn't come along, but that may not be the perspective of someone encountering the information for the first time.

PR #1345, opened 2019-09-14
Update article.md

paroche:
Change wording on inherited methods state changes. Not perfect but probably better.


iliakan:
Isn't the new sentence a bit long and complicated? 😊

paroche:
Well, I said it wasn't perfect. I'll look at it again, though.

iliakan:
maybe remove "we need to be clear" ?

paroche:
Yeah. I was looking for a bridging phrase, but that's a bit awkward. I'm working on it....

iliakan:
The important thing is that methods reside in the parent object, but modify the state of the child object, not the parent one.

PR #1349, opened 2019-09-16
Update article.md

paroche:
Change "and" to "or". Goes better with "no". Because !a && !b == !(a || b).

BTW, I think it would be great if the "Edit on GitHub" link opened in a new tab by default. Otherwise it's kind of a pain back-arrowing to the article.

### DIFF
--- a/1-js/07-object-properties/02-property-accessors/article.md
+++ b/1-js/07-object-properties/02-property-accessors/article.md
@@ -104,7 +104,7 @@ There's no similar method to handle deletion of an accessor property. Only gette
 
 Descriptors for accessor properties are different -- as compared with data properties.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Please use Ctrl+click or Alt+click (smth like that) to open in a new tab =)

It's generally not recommended to open in a new tab, not to confuse people.

paroche:
Surprising. I find it confusing (kind of) when the various GitHub screens are on top of where I was in the tutorial. But I can **try** to remember to press the Ctrl key.

iliakan:
Please create an issue about making "Edit on GitHub" open in a new window.

Generally, opening in a new tab isn't a good thing, but there's a rationale in your request, let's get others' feedback.

PR #1374, opened 2019-09-24
Update article.md

paroche:
Add "the" before "'this' keyword".


iliakan:
May I ask why? Usually keywords don't need articles?

paroche:
Nice of you to approve it on faith, then. :-). But I'd say it's because you followed the keyword with the modifier "keyword". You could say "a method can use 'this'." or "a method can use the 'this' keyword." (as I did), but to say "a method can use 'this' keyword." is awkward (and just a little bit more so  because it sounds like "a method can use this keyword" -- but that's secondary at best).

PR #1376, opened 2019-09-24
Spell correction

siddtheone:
Changed hid -> hide


lex111:
@siddtheone thank you! :)

PR #1380, opened 2019-09-25
Update task.md

chety:


iliakan:
removed that line for utter simplicity (should be clear)

thanks1

PR #1386, opened 2019-09-27
Typo fixed in "Function expressions and arrows"

rahulgupta92:
Minor typo fixed in section "Function expressions and arrows".


lex111:
@rahulgupta92 thank you! :)

PR #1387, opened 2019-09-28
Update solution.md

paroche:
Replace 'this' in 'apply' call with empty object. Figured it was easier to bring up issue w/ PR than as a discussion.

(Or could leave out 'apply' altogether and just have f(...args). (?))

This would be just for clarity, since, as far as I can tell, 'this' in the returned function will always be 'undefined', but having it there in the code anyway could make one wonder if if might actually have a useful value (if the new Function prototype was used with an object method). From what I've been able to discern, it would not -- it would always be 'undefined' (in strict mode, of course).

Same example I was wrong about before -- am I wrong again?

And, if I'm not, IS there a way to make the prototype wrapper work with an object method that uses 'this'?

### DIFF
--- a/1-js/08-prototypes/03-native-prototypes/2-defer-to-prototype-extended/solution.md
+++ b/1-js/08-prototypes/03-native-prototypes/2-defer-to-prototype-extended/solution.md
@@ -4,7 +4,7 @@
 Function.prototype.defer = function(ms) {
   let f = this;
   return function(...args) {
-    setTimeout(() => f.apply(this, args), ms);
+    setTimeout(() => f.apply({}, args), ms);
   }
 };
 


iliakan:
The change is incorrect. 

The value of `this` should be correctly carried over, that's what `apply` does.

iliakan:
It should work correctly with an object method.

paroche:
But as far as I can tell, there is not 'this' inside the function **to** carry over. I made some minimal modifications to your task solution: I've put the function in an object and called it with ".defer", then had the 'defer' prototype function display 'this' just before scheduling the 'apply'. It shows 'this' as undefined. If I try to access 'this' in the method definition, execution stops. 
https://plnkr.co/edit/26bHMKEytc7zAP146y8M?p=info
(or full script below):
<pre>
<!DOCTYPE html><script>
'use strict';
Function.prototype.defer = function(ms) {
  let f = this;
  return function(...args) {
    alert(this); // * undefined
    setTimeout(() => f.apply(this, args), ms);
  }
};

// check it
let testObj = {
  c: 5,
  f(a, b) {
    alert( a + b );
    alert( a + b + this.c ); // ** execution stops
  },
}

testObj.f.defer(1000)(1, 2); // shows 3 after 1 sec
</script>
</pre>

What am I missing?

iliakan:
Here's how: https://next.plnkr.co/edit/G2zcNTymRroNdXZV

```js
Function.prototype.defer = function(ms) {
  let f = this;
  return function(...args) {
    setTimeout(() => f.apply(this, args), ms);
  }
};

// check it
let testObj = {
  c: 5,
  f(a, b) {
    alert( a + b );
    alert( a + b + this.c ); // ** execution stops
  },
}

testObj.f = testObj.f.defer(1000)

testObj.f(1, 2); // shows 3 after 1 sec
```

Should we add this kind of example somewhere?

paroche:
Perhaps. My original point was that the 'this' in the prototype method served no purpose when used  directly on an object method (and, of course, no purpose in a non-object function), and I thought, therefore, that it was kind of confusing to have it there in the 'apply'. This would remedy that by assigning the delay prototype code  directly back to the object method, where it would regain its 'this'. So that is illustrative. Since we don't really want to go around modifying the prototypes anyway (except for polyfilling), I suppose it's not relevant that it makes the prototype a bit harder to use, and I imagine in real life we would have been using a decorator all along anyway -- except that this was a lesson (article) on prototypes. Still, I do think some mention of the issue of the lost 'this' in the function returned by the 'delay' would be helpful.

Again, if I understand it correctly.

iliakan:
Added the example.

PR #1391, opened 2019-09-28
Update article.md

paroche:
Change spelling of "labelled" to "labeled". "Labelled" is British spelling, "labeled" is American. I hadn't noticed British spelling being used in general so thought it should probably be the American version.


iliakan:
Sure, thanks!

PR #1393, opened 2019-09-29
Update cloning algorithm URL

yuanworks:
Url was pointing to a broken page in the spec.

Updated to match the working URL (w3c.github.io/html/ redirects to html.spec.whatwg.org, so it seems it should be prefered).


lex111:
@yuanworks thank you!

PR #1395, opened 2019-09-29
Fix some words to better convey the actual meaning.

sleepyfurious:
The rationale of the fixes is described in commit messages.

### DIFF
--- a/1-js/04-object-basics/02-garbage-collection/article.md
+++ b/1-js/04-object-basics/02-garbage-collection/article.md
@@ -156,7 +156,7 @@ The following "garbage collection" steps are regularly performed:
 - The garbage collector takes roots and "marks" (remembers) them.
 - Then it visits and "marks" all references from them.
 - Then it visits marked objects and marks *their* references. All visited objects are remembered, so as not to visit the same object twice in the future.
-- ...And so on until there are unvisited references (reachable from the roots).
+- ...And so on until there isn't any unvisited reference (reachable from the roots).
 - All objects except marked ones are removed.
 
 For instance, let our object structure look like this:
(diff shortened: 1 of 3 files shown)


iliakan:
Merged except 1-js/04-object-basics/02-garbage-collection/article.md

sleepyfurious:
@iliakan I would like to discuss about the exception of 1-js/04-object-basics/02-garbage-collection/article.md. I can't really get why the original sentence is making sense. Here is the original:
> ...And so on until there are unvisited references (reachable from the roots).

Can you provide any alternative explanation around this sentence?

iliakan:
It visits references until there exist unvisited ones (and reachable from the roots).

What should I fix to make this clear?

sleepyfurious:
Does "There exist unvisited ones" intentionally mean "There are ones that have never been visited before"?

iliakan:
"never visited before during the current garbage collection process"

sleepyfurious:
So, the unvisited ones are those meant to be removed. Is this right?

iliakan:
The algorithm recursively visits all it can, starting from the roots. 

What's not visited after that => is removed.

iliakan:
Please suggest changes if you feel they are neccessary.

sleepyfurious:
The ambiguity of "unvisited references" is the two possible meanings between:
> "the reachable from the root but still unvisited ones"

and
> "the unreachable from the root ones". 

The second is what we mean. The first is surely not making any sense, but a reader can easily struggle with it that way because it is tempting to refer to what is said before (which is unvisited one during traversing in this case), not what is after (the references that are unreachable from the root).

I suggest:
> …And so on until every reachable (from the roots) references are visited.

PR #1401, opened 2019-10-02
Update article.md

paroche:
I thought this would make it clearer that the new object's prototype, where the inherited methods reside, is not the same entity as User.prototype (though it is created from it). For those who are not totally fresh on the subject from the referenced F.prototype article.

### DIFF
--- a/1-js/09-classes/01-class/article.md
+++ b/1-js/09-classes/01-class/article.md
@@ -89,7 +89,7 @@ What `class User {...}` construct really does is:
 1. Creates a function named `User`, that becomes the result of the class declaration. The function code is taken from the `constructor` method (assumed empty if we don't write such method).
 2. Stores class methods, such as `sayHi`, in `User.prototype`.
 
-Afterwards, for `new User` objects, when we call a method, it's taken from the new object's prototype, just as described in the chapter <info:function-prototype>. So the object has access to class methods.
+Afterwards, for `new User` objects, when we call a method it is taken from the new object's prototype, that was created from User.prototype (as described in the chapter <info:function-prototype>). So the object has access to class methods.
 
 We can illustrate the result of `class User` declaration as:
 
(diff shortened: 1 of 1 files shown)


iliakan:
The suggested change has another meaning, not as intended.

I rephrased: 
After `new User` object is created, when we call its method, it's taken from the prototype, just as described in the chapter <info:function-prototype>. So the object has access to class methods.

paroche:
I thought I had deleted that PR, because I had some uncertainty. But if class inheritance works using constructor functions, doesn't that imply that the properties inherited by the 'new' object from the class.prototype property are, after inheritance, independent of class.prototype, rather than being a reference to the same object? I'd think it would be a "one-time gift", as described for Constructor functions:
"F.prototype property is only used when new F is called, it assigns [[Prototype]] of the new object. After that, there’s no connection between F.prototype and the new object. Think of it as a “one-time gift”."

But I decided I would run some tests before making a PR (though I already had -- which is why I tried to delete it). But right now it's late and maybe my thoughts aren't in order.

iliakan:
What you say is correct. I don’t see any contradictions to the article and the current content.

paroche:
There isn't a contradiction if you understand it correctly, but what is meant by "the prototype" could be misconstrued -- is the new prototype [[Prototype]] of the new object, or, perhaps, the "prototype" property of the class from which it was derived? It often looks to me, looking at the diagrams, like the latter is being implied, though it's really the former.

Or, again, it's late and my thoughts aren't in order. I'm struggling to get clarity on this. Because, for instance, to my surprise, with regular objects, in tests I've run, changing the prototype of the object seemed to change the Object prototype and affect other objects, even those that had already been created (I think -- not looking at the code now). So that would be different. But, again, again, it's late.

iliakan:
the prototype is the object referenced by `[[Prototype]]` internal reference.

iliakan:
that's "by definition"

paroche:
Yes, I was only referring to how it could look to those who are not as well grounded in the concepts and definitions, when reading this:

"Stores class methods, such as sayHi, in User.prototype.
Afterwards, for new User objects, when we call a method, it’s taken from the prototype, just as described in the chapter F.prototype. So the object has access to class methods."

Where one could possibly (though incorrectly) think maybe that "the prototype" could refer to "User.prototype" just mentioned. Which is why I had thought to make it explicit here (reinforcing what's in the F.prototype article) that the new object prototype is created from, but is not the same as, User.prototype.

PR #1411, opened 2019-10-03
Update article.md

paroche:
Replace 'cause' with 'case'.


lex111:
@paroche thanks! ;)

paroche:
No prob. Funny how it was just one extra letter forming an unrelated (as far as I know) word, but still almost worked the way it was.

lex111:
Well, we all make mistakes.

paroche:
Sure do. 
But it sort of **was** a cause as well as a case, even though I think "case" was intended.

lex111:
I got you, but we are not native speakers, so such mistakes are typical for us.

paroche:
Of course. You're doing great.

iliakan:
That was actually a typo :/ thanks.

paroche:
That's what I thought was most likely.

PR #1414, opened 2019-10-04
Update article.md

paroche:
Replaced 'the ... property ... works since" with "the ... property ... has worked since".

### DIFF
--- a/1-js/08-prototypes/04-prototype-methods/article.md
+++ b/1-js/08-prototypes/04-prototype-methods/article.md
@@ -71,7 +71,7 @@ Why so?
 
 That's for historical reasons.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Thanks! Please note, that I read all PR changes, so telling what you replace in the PR message is not needed =).

paroche:
I do it largely for my own reference so I can see which one it is without drilling further. And sometimes to explain.

Sent from AOL Mobile Mail
On Sunday, October 6, 2019 Ilya Kantor <reply@reply.github.com> wrote:

Thanks! Please note, that I read all PR changes, so telling what you replace in the PR message is not needed =).

PR #1426, opened 2019-10-08
smart header to actual header on intro article

JamieDanielson:
Smart header sections show as code sections and are not easy to read; they even have a scroll bar at the bottom of the window to see the full text.  By changing to ##, they match layout and are more consistent with the rest of the article.  This makes that adjustment.


iliakan:
That may be a markup issue? Please provide a screenshot in issues. 

Otherwise you just want us to remove "smart blocks" altogether =)

PR #1431, opened 2019-10-09
Update article.md

paroche:
listener to events -> listener for events


iliakan:
Are you sure that "listen to events" is wrong? I've seen that wording so many times.

paroche:
I've seen both, like here:
https://gomakethings.com/why-event-delegation-is-a-better-way-to-listen-for-events-in-vanilla-js/

In the above, one chooses which events to listen TO, but then listens FOR each of them. I think. I'm not sure about JS conventions, but in common speech, when one listens TO something it's continuous: I'm listening to this song (that is playing right now). One listens FOR something that will, or might, occur in the future: "I'm listening for my favorite song to come on the radio". But "listen to" also has the meaning of listening to something in general "I listen to Dave but I don't listen to Mike." So, in the JS context, I would think it would be: I listen TO certain (types of) events (clicks, hovers, etc.), but for a given type of event I listen FOR instances of that event (e.g. a click). In this case, in the text, I think what the handler function is doing is listening for instances of the event with the given name. It listens TO that event type FOR instances of that event.

iliakan:
Replaced everywhere.

PR #1442, opened 2019-10-10
Complement task solution with additional details

nakhodkiin:
Since there were some discrepancies around the mechanism of the built-in `Object` constructor, this PR is supposed to clarify the subject.


iliakan:
It's simpler than that. Added a few words.

PR #1445, opened 2019-10-12
Grammar issues fix

sudhirkumarojhaa:
Made some changes to make it more correct.


iliakan:
Looks wrong to me, but let @paroche decide.

iliakan:
Thanks!

iliakan:
@paroche Please review =)
"each one equipped with its own bells and whistles." better than "each one with its own bells and whistles." ?

paroche:
I think it's better without "equipped". Could even leave out "one", as below:

There exist many frameworks and development methodologies to build them, each with its own bells and whistles. Usually, special CSS classes and conventions are used to provide "component feel" -- CSS scoping and DOM encapsulation.

There is a slight awkwardness about the way it reads, though. Could perhaps rephrase:
"..., and each one has its own bells and whistles."

iliakan:
Thank you! Changed the file.

PR #1447, opened 2019-10-13
Update article.md

paroche:
Long awaited second stab at revising this paragraph (which someone else in the distant past took issue with).


iliakan:
Great. Finally sorted out =)

paroche:
It's a load off my mind.

PR #1450, opened 2019-10-13
Update article.md

paroche:
us, them, commas


iliakan:
@paroche, adding you as a maintainer, please commit directly, unless you'd like to ensure that the meaning is correct.

PR #1457, opened 2019-10-15
Update article.md

paroche:
'Changing DOM needs' -> "Changing the DOM requires'

(still prefer review)


Sebastianokunkelo:
paroche-patch-1

PR #1471, opened 2019-10-17
Update article.md

hrodward:
Grammar fixes


paroche:
Changes look right to me.

iliakan:
"increase to" seems more correct, no?

paroche:
Are 300ms and 1000ms the end results? If so, yes, "increase to". If it is an amount added to some prior value, then "increase by".

iliakan:
Yes it's increase to. I'll fix that.

PR #1482, opened 2019-10-19
minor

Ghost-017:


iliakan:
@paroche should I put articles before `attributeName`?

paroche:
If you mean the event names as in the patch here, then yes -- the several changes by Ghost-017 in patch-4 look right to me. Was that what you meant?

(Only the insertion of "for this" seems optional , i.e. "The reason" works OK by itself, but it is just a little better with "for this" following it).

iliakan:
Yes, all these changes are correct? Including "the" before event names?

I guess, my article skills are far from perfect 🤷‍♂

paroche:
Yes, they read better with all the "the"s. As you may know, it is a stereotype of a native Russian-speaker speaking English that they leave out articles (and the occasional preposition):  "Need go now -- must feed cat and dog!" Perfectly understandable, but should have "to" and at least one "the" or "my" or something (unless you have a cat named "cat" and a dog named "dog"). But I think you usually get it right. And almost all writers can use help proofing and copy editing, even native speakers.

paroche:
It had me down as a pending reviewer, though I thought I had already reviewed it. So I did the merge to get it to shut up. Hope that was OK.

iliakan:
Sure, you're totally right about articles =).

I just wanted to ask, whether or not the list might be an exception?

- `unload` event..
- `load` event..
...

It's easier to read when list items start with the an event name, rather than the "The" article.

Or not? So the question is specifically list-related.

paroche:
Yes, I think in that context is OK.

PR #1484, opened 2019-10-19
minor

Ghost-017:


2-ui/5-loading/03-onload-onerror/article.md
```diff
@@ -49,11 +49,11 @@ script.onload = function() {
 
 So in `onload` we can use script variables, run functions etc.
 
-...And what if the loading failed? For instance, there's no such script (error 404) or the server or the server is down (unavailable).
+...And what if the loading failed? For instance, there's no such script (error 404) or the server is down (unavailable).
 
 ### script.onerror
 
-Errors that occur during the loading of the script can be tracked on `error` event.
+Errors that occur during the loading of the script can be tracked in an `error` event.
```
paroche: Or "tracked through an 'error' event"

2-ui/5-loading/03-onload-onerror/article.md
```diff
@@ -69,12 +69,12 @@ script.onerror = function() {
...
 ```
 
-Please note that we can't get HTTP error details here. We don't know was it error 404 or 500 or something else. Just that the loading failed.
+Please note that we can't get HTTP error details here. We don't know if it was an error 404 or 500 or something else. Just that the loading failed.
 
 ```warn
 Events `onload`/`onerror` track only the loading itself.
 
-Errors during script processing and execution are out of the scope of these events. To track script errors, one can use `window.onerror` global handler.
+Errors during script processing and execution are out of scope for these events. To track script errors, one can use `window.onerror` global handler.
```
paroche: I'm not sure about this one. It depends on whether "out of scope" has a specific meaning in this context (like "out of range for this array" or something). Otherwise, if is a matter of something more like "not included in these events" I prefer the original.

2-ui/5-loading/03-onload-onerror/article.md
```diff
@@ -192,7 +192,7 @@ window.onerror = function(message, url, line, col, errorObj) {
 <script *!*crossorigin="anonymous"*/!* src="https://cors.javascript.info/article/onload-onerror/crossorigin/error.js"></script>
 ```
 
-Now, assuming that the server provides `Access-Control-Allow-Origin` header, everything's fine. We have the full error report.
+Now, assuming that the server provides an `Access-Control-Allow-Origin` header, everything's fine. We have the full error report.
```
paroche: For this, and everything I didn't comment on, changes look good. Should that be assumed? :-)

2-ui/5-loading/03-onload-onerror/article.md
iliakan: I mean: these events do not trigger for such kinds of errors. Which variant is better?

2-ui/5-loading/03-onload-onerror/article.md
iliakan: Great =)

2-ui/5-loading/03-onload-onerror/article.md
paroche: Probably the original. Or you could say "Errors during script processing and execution do not trigger these events."

PR #1485, opened 2019-10-19
minor

Ghost-017:


iliakan:
@paroche do you receive review requests? I just tried to set you as a reviewer for this PR :)

paroche:
Yes, receive review requests. As far as I know, I've responded to them all. Now for this one...

2-ui/99-ui-misc/01-mutation-observer/article.md
```diff
@@ -207,7 +207,7 @@ let demoElem = document.getElementById('highlight-demo');
 observer.observe(demoElem, {childList: true, subtree: true});
 ```
 
-Here's HTML-element and JavaScript that dynamically fills it using `innerHTML`.
+Here's an HTML-element and JavaScript that dynamically fills it using `innerHTML`.
```
paroche: Without seeing more context, it's kind of hard, but I would think it should maybe be "Here's an HTML-element and some JavaScript that...". 

The rest of the changes are fine.

iliakan: You can click to the left at these signs to get more context
image

Probably, you already know that.

paroche: No, just seems obvious after being pointed out.

paroche: Except that there was nothing I could find to expand down after the part in question, which was what I wanted. 

But I found the article using a "StartPage" search. I now think it's OK to just say "and JavaScript" without any words between. So the changes as provided should be fine.

PR #1492, opened 2019-10-21
Update article.md

hrodward:
Grammar

### DIFF
--- a/1-js/08-prototypes/01-prototype-inheritance/article.md
+++ b/1-js/08-prototypes/01-prototype-inheritance/article.md
@@ -246,7 +246,7 @@ The resulting picture:
 
 ![](proto-animal-rabbit-walk-3.svg)
 
-If we had other objects like `bird`, `snake` etc. inheriting from `animal`, they would also gain access to methods of `animal`. But `this` in each method call would be the corresponding object, evaluated at the call-time (before dot), not `animal`. So when we write data into `this`, it is stored into these objects.
+If we had other objects like `bird`, `snake` etc., inheriting from `animal`, they would also gain access to methods of `animal`. But `this` in each method call would be the corresponding object, evaluated at the call-time (before dot), not `animal`. So when we write data into `this`, it is stored into these objects.
 
 As a result, methods are shared, but the object state is not.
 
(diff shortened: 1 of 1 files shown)


1-js/08-prototypes/01-prototype-inheritance/article.md
```diff
@@ -246,13 +246,13 @@ The resulting picture:
 
 ![](proto-animal-rabbit-walk-3.svg)
 
-If we had other objects like `bird`, `snake` etc inheriting from `animal`, they would also gain access to methods of `animal`. But `this` in each method call would be the corresponding object, evaluated at the call-time (before dot), not `animal`. So when we write data into `this`, it is stored into these objects.
+If we had other objects like `bird`, `snake` etc. inheriting from `animal`, they would also gain access to methods of `animal`. But `this` in each method call would be the corresponding object, evaluated at the call-time (before dot), not `animal`. So when we write data into `this`, it is stored into these objects.
```
lex111: @paroche don't we need a comma here?

```suggestion
If we had other objects like `bird`, `snake` etc., inheriting from `animal`, they would also gain access to methods of `animal`. But `this` in each method call would be the corresponding object, evaluated at the call-time (before dot), not `animal`. So when we write data into `this`, it is stored into these objects.
```

paroche: The place where I would add a comma is between "'snake'"  and "etc.". I think a comma after "etc." would only be appropriate if there were also a comma between "objects" and "like". So:

"If we had other objects like `bird`, `snake`, etc. inheriting from 'animal'..."

or

"If we had other objects, like 'bird', 'snake', etc., inheriting from 'animal'..."

lex111: >  I think a comma after "etc." would only be appropriate

This is what I wanted to know and reflected in my suggestion.

@hrodward please accept the suggestion above.

hrodward:
I clicked on “resolve discussion”. Is that enough?

lex111:
@hrodward please click on "Commit suggestion" button.

hrodward:
Done!

1-js/08-prototypes/01-prototype-inheritance/article.md
paroche: I entered a reply here a few hours ago, but I can no longer find it. To paraphrase what I said then, I don't agree with this change. I would suggest a comma **before** "etc.", but would only recommend a comma **after** "etc." IFF there were also a comma before "like" (after "objects"), as in my examples above.

hrodward: I agree with @paroche. My preferred option would be the comma before "etc."

lex111: Oops, fine, let's make it.

paroche: Done. Went w/ 2nd (3 comma) option.

PR #1498, opened 2019-10-21
Fixed object variable user.go() to be obj.go()

koala-lava:


lex111:
@koala-lava thank you!

PR #1505, opened 2019-10-22
Update article.md

hrodward:
Grammar suggestions

### DIFF
--- a/1-js/08-prototypes/04-prototype-methods/article.md
+++ b/1-js/08-prototypes/04-prototype-methods/article.md
@@ -65,9 +65,9 @@ This call makes a truly exact copy of `obj`, including all properties: enumerabl
 
 ## Brief history
 
-If we count all the ways to manage `[[Prototype]]`, there's a lot! Many ways to do the same!
+If we count all the ways to manage `[[Prototype]]`, there are a lot! Many ways to do the same!
 
-Why so?
+Why?
 
 That's for historical reasons.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Why not "**The** `__proto__` is not a property of an object, but an accessor property of `Object.prototype`:..." ?

hrodward:
In my opinion, using "The" would require to specify what __proto__ is, as in "The __proto__ accessor..."

paroche:
Should not be a "The" before "`__proto__`", because "`__proto__`" in this context already indicates a specific thing, so would not have to be specified further with "the"; it is already the only member of its class of things (of things that are "`__proto__`"). 

It's like if you have a dog Rover, you don't say "The Rover is a sheep dog", just "Rover is a sheep dog."* (never met a dog named "Rover", BTW, but it's a classic example dog name). 

**But**, if you had written "`__proto__` **property** is not a property of an object...", THAT would require a "The" before "`__proto__`" (i.e., should have "The `__prototype__` property is not ...") because then it is one of several (more than one) possible properties.
<br>
\* (Actually, you **might** refer to the aforementioned hypothetical dog as "The Rover" if you thought he was putting on airs or something, such that his name had become a title, like "king", but that's another story.)

iliakan:
Thanks a lot for the explanations!

paroche:
Hope was somewhat clear.

paroche:
Please see changes above.

PR #1513, opened 2019-10-24
Update article.md

hrodward:
Grammar suggestions

### DIFF
--- a/1-js/09-classes/06-instanceof/article.md
+++ b/1-js/09-classes/06-instanceof/article.md
@@ -2,7 +2,7 @@
 
 The `instanceof` operator allows to check whether an object belongs to a certain class. It also takes inheritance into account.
 
-Such a check may be necessary in many cases, here we'll use it for building a *polymorphic* function, the one that treats arguments differently depending on their type.
+Such a check may be necessary in many cases. Here we'll use it for building a *polymorphic* function, the one that treats arguments differently depending on their type.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Maybe "That can lead to interesting consequences when **the** `prototype` property is changed after the object is created." ?

hrodward:
Yep, 'the' is better

paroche:
Agreed. "the prototype property". There's only one (for a given class).

paroche:
However, in the same patch:

_2. Most classes do not have `Symbol.hasInstance`. In that case, the standard logic is used: `obj instanceOf Class` checks whether `Class.prototype` equals to one of prototypes in the `obj` prototype chain.
2. Most classes do not have `Symbol.hasInstance`. In that case, the standard logic is used: `obj instanceOf Class` checks whether `Class.prototype` equals to any prototype in the `obj` prototype chain."_

Both are wrong. It should be "is equal to any prototype", **"is equal to one of the prototypes"**, "equals any prototype", or "equals one of the prototypes". Never "equals to". I would somewhat favor the one I put in bold.

iliakan:
@hrodward will you update the patch if the suggestion is acceptable?

hrodward:
> @hrodward will you update the patch if the suggestion is acceptable?

Done!

lex111:
@hrodward thanks!

Issue #716, opened 2019-01-08
Add some additional tests to Throttle task


dagolinuxoid:
I've "solved"  (passed all current provided tests) throttle task, but unfortunately my solution is *wrong* (at least isn't that efficient - it makes few redundant calls). It's not a very important issue although it would be nice to fix it.
srcMyAttempt: [throttle](https://github.com/dagolinuxoid/fccActivityLog/blob/master/0701019/throttle.js)
srcOriginalTask: [throttle task](https://javascript.info/call-apply-decorators#throttle-decorator)

iliakan:
Do you have a suggestion what to fix/add? Maybe submit a PR?

dagolinuxoid:
got it :) The issue isn't that urgent so it can hang for a while. Honestly I'm not strong in writing complicated (heck even regular) tests yet so it's going to be fun. I will submit PR after figuring it out.

iliakan:
@dagolinuxoid would you submit a PR for that?

dagolinuxoid:
Hi! If I'm not going to do something about it until November you can officially give up on me, so there is a hope.

dagolinuxoid:
@iliakan it's been a while =) There is a working solution for node. I don't know why it doesn't work in a browser though.
``` javascript
const assert = require('chai').assert;
const throttle = require('../throttle');

describe('efficiency tests', () => {
  let log = '';
  const f = arg => log += arg;
  const f100 = throttle(f, 100);
  f100('once');

  it('a forwarded call should run just once', done => {
    setTimeout(function() {
      assert.equal(log, 'once');
      done();
    }, 200);
  });
});

**Comment by dagolinuxoid** [comment]
I figured it out. My test is working in browser too. However in order to work it on Plunker it's idex.html file should be modified slightly:
```html
<!doctype html>
<script src="https://en.js.cx/test/libs.js"></script>

<script>
  // ...Your code...
</script>

<script src="test.js"></script>
</html>
```
Placement of srcipts matter : test should be placed *after* a script with 'your code here'.
The repository doesn't have index.html file here (only file with tests) So probably It's best for you to modify the index on Plunker directly so the additional test can be added successfully via PR.
```javascript
describe('efficiency tests', () => {
  let log = '';
  const f = arg => log += arg;
  const f10 = throttle(f, 10);
  f10('once');

  it('a forwarded call should run just once', done => {
    setTimeout(() => {
      assert.equal(log, 'once');
      done();
[...]

PR #1524, opened 2019-10-28
add an asyncronous test for throttle

dagolinuxoid:
The addition to the main test.js guarantees that only unique calls are made.

### DIFF
--- a/1-js/06-advanced-functions/09-call-apply-decorators/04-throttle/_js.view/test.js
+++ b/1-js/06-advanced-functions/09-call-apply-decorators/04-throttle/_js.view/test.js
@@ -45,3 +45,17 @@ describe("throttle(f, 1000)", function() {
   });
 
 });
+
+describe('prevent duplicate calls', () => {
+  let log = '';
+  const f = str => log += str;
+  const f10 = throttle(f, 10);
+  f10('once');
+
+  it('a forwarded call should run just once', done => {
+    setTimeout(() => {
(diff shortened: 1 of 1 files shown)


dagolinuxoid:
It closes #716. However in order to make it work properly *index.html* file on Plunker **should** be edited as well:
Its current state is :
```html
<!doctype html>
<script src="https://en.js.cx/test/libs.js"></script>
<script src="test.js"></script>
<script>

  // ...Your code...

</script>

</html>
```
Its required state should be the fallowing:
```html
<!doctype html>
<script src="https://en.js.cx/test/libs.js"></script>
<script>

  // ...Your code...

</script>
<script src="test.js"></script>
</html>
```

iliakan:
@dagolinuxoid why should we edit index.html?

For other tests it's fine, what's so special about the new test?

dagolinuxoid:
Otherwise it's not going to work, that's most likely due to I used the `done` parameter  for *it* spec. I understand that even small altering of index.html can be a troublesome so it's up to you to make a decision.

dagolinuxoid:
I'll try to solve it without altering html file. If it's not trivial I'll comment here and you can decide either accept or decline the PR. So hold on till evening.

iliakan:
Thank you!

dagolinuxoid:
You put all the code inside of anonymous function and now all things are working properly. Good job.

PR #1527, opened 2019-10-28
string outputs without quotes

lumosmind:
Some of string outputs are represented with quotes but some of not, there is a inconsistency.


lex111:
@lumosmind make sense, thank you!

PR #1529, opened 2019-10-28
'str' inside of 'start'

lumosmind:
**splice** method's minimum parameter is '**start**' and it is **number** type. In this tutorial '**str**' stands for **string** parameters.


lex111:
@lumosmind thank you!

PR #1530, opened 2019-10-28
add parameter definition to arr.find()

lumosmind:
For consistency with definition of "**arr.filter(fn)**" , we need to add '**fn**' parameter to definition of **arr.find()**.


iliakan:
We don't always specify args when mention a method though.

PR #1536, opened 2019-10-29
correction for code template consistency

lumosmind:
'**:**' character is indicating '**var2**' is a **key** of object.


lex111:
@lumosmind thanks!

PR #1541, opened 2019-10-29
JSON.stringfy() output variable type consistency

lumosmind:
Output variable type of **JSON.stringfy()** is **string**. So, we need to show all outputs **in quotes** to ensure consistency. alert() function shows outputs without quotes even if they are string, but this can cause misunderstanding.

### DIFF
--- a/1-js/05-data-types/12-json/article.md
+++ b/1-js/05-data-types/12-json/article.md
@@ -90,14 +90,14 @@ For instance:
 
 ```js run
 // a number in JSON is just a number
-alert( JSON.stringify(1) ) // 1
+alert( JSON.stringify(1) ) // "1"
 
 // a string in JSON is still a string, but double-quoted
 alert( JSON.stringify('test') ) // "test"
 
-alert( JSON.stringify(true) ); // true
+alert( JSON.stringify(true) ); // "true"
 
-alert( JSON.stringify([1, 2, 3]) ); // [1,2,3]
+alert( JSON.stringify([1, 2, 3]) ); // "[1,2,3]"
 ```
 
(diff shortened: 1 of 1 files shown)


lumosmind:
I figured out that why some outputs with quotes and some not. JSON.stringify() method specify string type variables with extra quotes like ""some text"" but other primitive types are specified with minimum quotes like "true", "null", "12" etc. alert() function removes fist quotes from string variables because of visual reasons(ui/ux), but inner quotes are showed by alert() function. In short, all variables are in string for which alert() function shows.

iliakan:
This isn't actual any more, right?

lumosmind:
I was just trying to explain what is going on the code. **These changes are still valid!**

lumosmind:
> This isn't actual any more, right?

I was just trying to explain what is going on the code. These changes are still valid!

iliakan:
Usually the output of `alert` is commented "as it's shown on screen", that's why no quotes.

lumosmind:
I got it. Some times, I saw string values with quotes and these are more self expressive. Thanks.

iliakan:
Then we'll need to put quotes everywhere. And, imagine, how an actually quoted string would look like. `""string""`? :)

lumosmind:
I think we can't make JS look like pretty. We need to accept "is as". And we are programmers, we need to deeply know what is going on really under the hood and we need to ready for weirdness of JS :) Also I think this tutorial has advance level JS knowledge. And alert() function hide something from us. It can be better to use console.log() implementation on in article examples. If anyone wants to learn hard subjects like **closure** or **IIF**, there is no beautiful way. We must see JS without hiding anything. (all of these are subjective)

PR #1543, opened 2019-10-29
linked list must end with 'null'

lumosmind:
We need to put '**null**' to end of linked list. Otherwise when we want to get last item in the linked list, it will be **undefined**. And this situation inconsistent with **object literal definition** of linked list above.


iliakan:
Thank you for noticing such details!

PR #1544, opened 2019-10-29
removes russian text chunk from english text

TobiObeck:


lex111:
@TobiObeck thank you!

PR #1545, opened 2019-10-29
Just wanted to clear that numbers followed by a string.

mdeasindev:
I think people will get it perfectly now then before. Because before there was written only specific number (like 2 numbers) followed by a string but I thought it should be obvious that numbers should be more than a number followed by a string, not only 2 numbers. People might be confused like me :)

Thanks


iliakan:
I'm sorry, but the text is even more confusing after the change (imho)

PR #1549, opened 2019-10-30
avoid to override built-in properties

lumosmind:
In the snippet we aren't using **var** keyword so, it looks like **modern JS codding** standard. we block the **var** danger but there is a way to override built-in properties in window object like Math object and this is very dangerous for project. To avoid this situation we need to check for "**is property name in use**?".

### DIFF
--- a/1-js/06-advanced-functions/05-global-object/article.md
+++ b/1-js/06-advanced-functions/05-global-object/article.md
@@ -40,9 +40,11 @@ If a value is so important that you'd like to make it available globally, write 
 ```js run
 *!*
 // make current user information global, to let all scripts access it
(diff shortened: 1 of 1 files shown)


GeriHAl:
Please I want to unsubscribe this thread 

Geraldine Hernández Al Attrach

iliakan:
@GeriHAl then click "unsubscribe"

iliakan:
In our app, we do know that `window.currentUser` is not reserved.

lumosmind:
> In our app, we do know that `window.currentUser` is not reserved.

But, **defensive programming** is much better specially when we working with `windows `object.

PR #1552, opened 2019-10-30
use moq instead of pseudo

lumosmind:
Using **receiveCodeFromServer()** (hypothetical mock function) is better than using pseudo code. Pseudo code can't be handled by code highlighter in right way. So, pseudo code make real code unreadable and hard to understand.

Server can send us **function as declaration** bu I think you mean that server sends **code snippet**( maybe function body) for generalization, not **function declaration**. Because you don't **call** the function which comes from server bu you talking about **execution** of it.

### DIFF
--- a/1-js/06-advanced-functions/07-new-function/article.md
+++ b/1-js/06-advanced-functions/07-new-function/article.md
@@ -33,10 +33,10 @@ The major difference from other ways we've seen is that the function is created 
 
 All previous declarations required us, programmers, to write the function code in the script.
 
-But `new Function` allows to turn any string into a function. For example, we can receive a new function from a server and then execute it:
+But `new Function` allows to turn any string into a function. For example, we can receive a new code from a server and then execute it:
 
 ```js
-let str = ... receive the code from a server dynamically ...
+let str = receiveCodeFromServer(); // asume that it returns code in string format.
 
 let func = new Function(str);
 func();


iliakan:
The communication is usually performed asynchronously, while a function call, like you did, is sync.

lumosmind:
> The communication is usually performed asynchronously, while a function call, like you did, is sync.

Lets make it asynchronous and more complex :) There is a trade-of and I think my suggestion is still valuable.

iliakan:
The particular code example is about another thing, it's better not to distract from it.

lumosmind:
I got you, but the code example has `let str =` **variable definition** and **assignment**  so how can we think this code as **asynchronous**? also code flow is completely **synchronous**. 

Current code example has **spread operator** and `from `keyword in pseudo code and this is so strange.

I reread the article and there is no another example code directly related with the code example.

> For example, we can receive a **new function** from a server and then execute it:

> It is used in very specific cases, like when we receive **code** from a server, or
these words are in conflict with each other.

iliakan:
is there something confusing and unclear about the topic explained in the article, that requires your changes?

lumosmind:
> is there something confusing and unclear about the topic explained in the article, that requires your changes?

1-)   the example code is not clean. (this is subjective)
2-)  I think, **executing function** and **calling function** are same thing. 

> For example, we can receive a **new function** from a server and then execute it

` let str = "  function(){ alert('hello'); }  "; `
How can we execute(or call) this function from in **new Function**?
I think we need to change this sentence. like this
> For example, we can receive a **new code** from a server and then execute it

Because we can execute any code in **new Function** like this code. And this code can execute function which comes from server. But we can't call that code as just a function.
`let str = " let func = function(){ alert('hello');   func(); } ";`

PR #1554, opened 2019-10-30
code non-compatible with clarification drawing

lumosmind:
In **clarification drawing**(setinterval-interval.svg) `func `is calling with different argument values like `func(1)`, `func(2)` ...
But we can't see why variable `i` changing. Maybe we need `func(i++)` to explain this situation.


iliakan:
Thanks a lot! Good catch.

Issue #1523, opened 2019-10-27
redundant Timeout Error


lumosmind:
Examples about  "**before/after and beforeEach/afterEach**"  is not usefull. Because, it gives **Timeout Error** when we use **makeTest** function and create a lot of test checks. So, we must use **console.log** function instead of **alert** function for **beforeEach** and **afterEach** functions.

iliakan:
Where is the problem exactly?

lumosmind:
alert() function waits us for user interaction and this waiting process takes time. Mocha try to finish test in limited time. so I take a lot of error when I can't click super fast to **ok** button of aler() function message.

iliakan:
Maybe just increase timeout? 
`console.log` requires people to open the console, while `alert` doesn't.

```
describe...{
 this.timeout(500);
}
```

lumosmind:
Example code isn't executable in article or it doesn't have link to plunker. It is just static code and readers need to copy and paste the code to their local .js file and run. opening console can't be a problem for readers at this level. (They are learning unit test.)
Before the article they already have read **Debugging in Chrome** article. 

`describe...{
 this.timeout(500);
}`
thanks this code solve my issue but article needs more touch.

PR #1562, opened 2019-11-02
console.log() instead of alert()

lumosmind:
possible fix for issue #1523
alert() function exceeds **Mocha's time limit** for test run. So, mocha creates **timeout error** several times when we can't click "**ok**" button of alert() message immediately.

Also I think you worked with console.log() function when you were preparing for article but you converted it to alert(). Because you are talking about running sequence like console.log() outputs:

> Testing started – before all tests (before)
> Before a test – enter a test (beforeEach)
> 1
> After a test – exit a test   (afterEach)
> Before a test – enter a test (beforeEach)
> 2
> After a test – exit a test   (afterEach)
> Testing finished – after all tests (after)

### DIFF
--- a/1-js/03-code-quality/05-testing-mocha/article.md
+++ b/1-js/03-code-quality/05-testing-mocha/article.md
@@ -269,14 +269,14 @@ For instance:
 ```js no-beautify
 describe("test", function() {
 
-  before(() => alert("Testing started – before all tests"));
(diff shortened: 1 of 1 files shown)


iliakan:
Maybe just increase timeout there? We use alerts everywhere normally for simplicity.

PR #1563, opened 2019-11-02
not proved feature in article

lumosmind:
Some things has been mentioned about : "If user object changes while setTimout() waiting for run callback function which is method of user object, what will happen?"

> "What if before setTimeout triggers (there’s one second delay!) user changes value? Then, suddenly, it will call the wrong object!"

And some things has been mentioned about "we will solve the problem with function binding"

> "The next solution guarantees that such thing won’t happen."

**But there is nothing to show that _function binding_ really solves the problem which has been mentioned before.**


iliakan:
Thanks, added few more words.

PR #1564, opened 2019-11-02
wrapper instead of helper

lumosmind:
"**helper function**" term reminds "**utility libraries**" but "**wrapper**" term refers "**function decorators**" and in this context, wrapper can be more self-descriptive and meaningful.


iliakan:
Wait, no, `partial` returns a wrapper, but is used as a helper here.

iliakan:
reverted the change.

lumosmind:
I am illuminated literally :) thank you!

PR #1568, opened 2019-11-02
decorator instead of helper

lumosmind:
It can be more expressive.


iliakan:
maybe just remove that word? it's complex enough without an additional "decorator" note =)

iliakan:
done

PR #1576, opened 2019-11-04
copy is not suit for this case

lumosmind:
Copying method into another object differs from using call/apply, I think.
Functions are passed by reference and they are kind of special objects. So, copying of functions can make us recall wrong things like object copying or deep object copying etc.
If I copy a method into another object, I want to reach the method from the another object. But we can't do that by using call() or apply().

### DIFF
--- a/1-js/08-prototypes/03-native-prototypes/article.md
+++ b/1-js/08-prototypes/03-native-prototypes/article.md
@@ -157,7 +157,7 @@ alert( "La".repeat(3) ); // LaLaLa
 
 In the chapter <info:call-apply-decorators#method-borrowing> we talked about method borrowing.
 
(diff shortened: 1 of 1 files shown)


iliakan:
The word `copy` makes things easier to understand. It's also correct. Methods are copied by reference, we know that already at this point of the tutorial.

lumosmind:
> The word `copy` makes things easier to understand. It's also correct. Methods are copied by reference, we know that already at this point of the tutorial.

let's copy a function :)

```
function sum(a,b){return a+b;}
let sumCopy = new Function('a', 'b', sum.toString());

sumCopy === sum; // false
sum === sum // true
sumCopy === sumCopy // true
```

PR #1577, opened 2019-11-04
assign instead of copy

lumosmind:
Methods are passed by **reference** not copied like primitives.


iliakan:
Copied by reference. That's fine.

lumosmind:
> Copied by reference. That's fine.

Analogy is the murder of truth :)

iliakan:
I don't understand. “copy by reference” is a well-known term.

lumosmind:
> I don't understand. “copy by reference” is a well-known term.

you are right. that is actually what  I mean: "copying an object" is different from "copying by reference".
this tutorial consists very deep level knowledge about js. So, using right term in convention, is very important to figured out complex things.

PR #1581, opened 2019-11-05
constructor function instead of constructor method

lumosmind:
Constructor methods make us recall "**the constructor method in class declaration structure**". But in here, you show that the function has prototype object and this object has an attribute "constructor" which is a reference to the function itself. So we learn this pattern as a "constructor function" in this article:

https://javascript.info/function-prototype

### DIFF
--- a/1-js/09-classes/01-class/article.md
+++ b/1-js/09-classes/01-class/article.md
@@ -106,7 +106,7 @@ class User {
 // class is a function
 alert(typeof User); // function
 
-// ...or, more precisely, the constructor method
+// ...or, more precisely, the constructor function 
 alert(User === User.prototype.constructor); // true
 
 // The methods are in User.prototype, e.g:


iliakan:
From the other hand, it's still an object method...

lumosmind:
> From the other hand, it's still an object method...

So, let's call all functions as a method. Because all functions has [[Prototype]] object and [[Prototype]].constructor refers functions itself so there is no need to use "function" term in JavaScript :)

iliakan:
When we are talking about classes, functions defined in class declaration are "class methods". They have special label about that (`[[HomeObject]]'). Why should we replace the word "method" by "function"?

lumosmind:
[ecma method](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-method)

> function that is the value of a property
> 
> NOTE
> When a function is called as a method of an object, the object is passed to the function as its this value.
I got it finally.
We can't call **User()** without **new** keyword and **new** keyword create **new object** for **User()** call. so we have to call **User() function** with an object. So, in duration of calling,  it is the **method** of **newly created empty object**.

PR #1582, opened 2019-11-05
"syntactic sugar" instead of "syntax sugar"

lumosmind:
Historically and in general use, this phenomena is called as "syntactic sugar". I think we can change it for better understand-ability and engineering term convention.

https://www.computerhope.com/jargon/s/syntactic-sugar.htm


lex111:
@lumosmind thank you!

PR #1586, opened 2019-11-05
example code must be unexecutable

lumosmind:
Example code must be executable because, it is dependent on another example code above. So it can't work in plunker container.


lex111:
@lumosmind thank you!

PR #1597, opened 2019-11-08
Update article.md

paroche:
Changes to Summary description of `clientLeft/clientTop`. I **think** this is what was meant.


paroche:
Thanks!

PR #1600, opened 2019-11-09
Use `Number.isInteger()` to check integers

AhmedElaguab:
Use `!Number.isInteger(n)` rather than `Math.round(n) != n`.


iliakan:
I'm confused about that change, as it involves knowledge about Number.isInteger :/

PR #1601, opened 2019-11-10
modified a sentence

Haggr:
added a missing word


lex111:
@Haggr seems legit, thanks!

PR #1603, opened 2019-11-11
Added clarification

koala-lava:
The solution explanation is not very obvious so added another example to make things clear


iliakan:
Does it really make things more clear? 😕

koala-lava:
@iliakan At the time of writing this PR I was sure it was clearer, but now I'm not sure. The answer took me time to understand even though I had a prior experience in programming and OOP. I'm pretty sure many novices struggle with the current state answer.

If only there was an opportunity to test if this PR makes things clear. Some kind of voting.

PR #1605, opened 2019-11-12
typo fixed

koala-lava:
fixed chart -> cart


lex111:
@koala-lava good catch, thanks for this!

PR #1607, opened 2019-11-13
Update article.md

MattWilcox:
Clarify the colloquium "there is no return"; as "return" is an actual programatic concept its use here may confuse people (there most certainly _is_ return in Javascript even with "use strict" declared).


iliakan:
I hope that "there's no going back" is ok

PR #1609, opened 2019-11-13
Change "Looping over keys" part

cfinnberg:
Update the looping part with more modern techniques


iliakan:
This involves working with `localStorage` as with a plain JS object. That's not a good way to use it.

cfinnberg:
@iliakan Sorry, can you elaborate why it's not a good way to use it? Thanks!

iliakan:
Because it's unsafe. If keys are user-generated, a key can be `__proto__` or match an existing method, such as `length`, `key`, etc.

cfinnberg:
But as far I can tell, I can't overwrite length or __proto__ (I tried). And in your own example you are using Object.keys(). I just was adding the values() and entries() methods. Why is that different from yours? (just to be clear, I ask to learn, not to challenge your decision. I have tried to look for an answer myself, but haven't found anything useful)

PR #1612, opened 2019-11-18
Make minor grammar corrections/updates

jchue:
Namely in the Async section


iliakan:
There should be a space before dashes that you propose, such as `methods—`.
Also I'm not sure about moving punctuation signs inside of quotes.

@paroche could you please review that?

paroche:
Yes, space also before dash, There is a style which has no spaces on either side, and a style with spaces on both sides. For instance, newspapers that follow the AP style have spaces on both sides. That's what I favor in general. 

Here:
_The script is executed "asynchronously", as it starts loading now, but runs later, when the function has already finished.
The script is executed "asynchronously," as it starts loading now but runs later when the function has already finished._

It does seem to be the case that in American-style writing the commas go inside quotes, so the change would be "right", though I'm not a fan of that rule.
https://www.thepunctuationguide.com/quotation-marks.html

Are a lot of changes here, some OK, some not. Kind of hard to comment on the whole thing. But oh well....

_Here we did it in `loadScript`, but of course, it's a general approach.
Here we did it in `loadScript`, but of course it's a general approach._

Change is an improvement. Or could say "but, of course, it's...". 

_How can we load two scripts sequentially: the first one, and then the second one after it?
How can we load two scripts sequentially: the first one and then the second one after it?_

I don't think this is particularly an improvement. It's a matter of taste. I prefer the original.

_In the above examples we didn't consider errors. What if the script loading fails? Our callback should be able to react on that.
In the above examples, we didn't consider errors. What if the script loading fails? Our callback should be able to react to that._

I think it's better w/out the comma (1st change). But definitely, "to" is better than "on".

_Also, the functions named `step*` are all of single use, they are created only to avoid the "pyramid of doom." No one is going to reuse them outside of the action chain. So there's a bit of namespace cluttering here.
Also, the functions named `step*` are all of single use— they are created only to avoid the "pyramid of doom." No one is going to reuse them outside of the action chain. So there's a bit of namespace cluttering here._

Either a space before the dash, or use a semicolon (";").

_1. A "producing code" that does something and takes time. For instance, a code that loads the data over a network. That's a "singer".
2. A "consuming code" that wants the result of the "producing code" once it's ready. Many functions  may need that result. These are the "fans".
1. A "producing code" that does something and takes time. For instance, a code that loads the data over a network. That's a "singer."
2. A "consuming code" that wants the result of the "producing code" once it's ready. Many functions  may need that result. These are the "fans."_

Having commas and periods inside the quotes is the American style. So yes to the changes. (I think the Brits put the ending punctuation outside the quotes, which I do think is logical). I don't love that rule and it looks especially weird to me when it's only one word in quotes, but that does seem to be the way it is.

_That is, `.then` handler in line `(*)` now returns `new Promise`, that becomes settled only after the call of `resolve(githubUser)` in `setTimeout` `(**)`.
That is, the `.then` handler in line `(*)` now returns `new Promise` that becomes settled only after the call of `resolve(githubUser)` in `setTimeout` `(**)`. The next `.then` in the chain will wait for _that.__

Yes to the "the" before `.then`. But I prefer keeping the comma after `new Promise`.

_`Promise.all` takes an array of promises (it technically can be any iterable, but is usually an array) and returns a new promise.
`Promise.all` takes an array of promises (it technically can be any iterable but is usually an array) and returns a new promise._

I like it better with the comma, though it's correct either way.

_A bigger example with fetching user information for an array of GitHub users by their names (we could fetch an array of goods by their ids, the logic is identical):
A bigger example with fetching user information for an array of GitHub users by their names (We could fetch an array by their IDs, and the logic would be identical.):_

I'm not really in favor of either change.

_For example, `loadCached` function below fetches a URL and remembers (caches) its content. For future calls with the same URL it immediately gets the previous content from cache, but uses `Promise.resolve` to make a promise of it, so that the returned value is always a promise:
For example, the `loadCached` function below fetches a URL and remembers (caches) its content. For future calls with the same URL, it immediately gets the previous content from cache but uses `Promise.resolve` to make a promise of it, so the returned value is always a promise:_

I prefer keeping the comma. Slightly better without 'that'.

For everything not mentioned, I'm OK with the changes, or at least somewhat neutral. Or, I didn't notice them.

iliakan:
@paroche thanks, so do we stick to `"...,"` style? There are many other places in the tutorial, I think, where I used `"...",`

As I got it, we require spaces around a dash.

Any other changes we require PR author to make? Or we just take nothing from it and close it?

Also, when you'd like to correct an exact sentence, there's a way in GitHub to do that, probably you know, the "plus" sign that appears here:
image

And then "Review" button in the upper-right corner allows to request changes.

paroche:
I would not really want to see you change all the ".....", s to ".....," s. That is considered a rule in American usage but I think it's stupid, and despite many years of reading English it still looks strange to me. It is apparently (I've read) a holdover from manual typesetting when it helped keep the little commas and periods from falling off the plate. So I think it would be fine to stick with the British style for this, at least for lists of quoted items and the like: _'The boxes I see are labeled "red", "green", and "blue", but I think somewhere there's one labeled "purple"._  In the preceding, I especially prefer the comma to be after "blue". I could live with the period being inside of "purple.", but still wouldn't favor it (though it does look somewhat comfortingly familiar that way). Also, since this is technical writing, sometimes the things in quotes have to be taken literally, and a comma inside the quote could confuse things.

As far as the other changes, I have already expressed my opinions on most of them -- some of them are good, some are marginally better, some, in my opinion, slightly less good, some I didn't like, and the ones I didn't comment on I thought were at least OK. So I don't know how you should proceed with the PR.

I think I may have known about the "+" thing in GitHub at one time, for a little while.

iliakan:
@jchue thank you for your effort! Could you please update the Pull Request according to the discussion?

paroche:
You mean start a discussion on each of the, what, 70?, lines being edited? Or is there a way to change them directly (without creating a new PR)? The former would be incredibly tedious. The latter might be workable.

iliakan:
The author may correct the PR by pushing more commits. Or we can just manually import some worthy changes from this PR and close it.

iliakan:
@paroche maybe you could review that and just push to master the good changes?

I can do that myself, but I'm unable to validate all the changes. So if you approve everything (Except the quotes and the spaces around dash), let me know, then I'll do that.

jchue:
Gents, I have no issue making the changes. Just let me know what you prefer. Would it help if I updated the quotes and dashes first and/or broke up the PR into smaller portions to help with your review?

iliakan:
@paroche, please respond as you're the grammar guy in the room =)

paroche:
If jchue would revise his changes w/ respect to the spaces before dashes and leaving the punctuation out of the quotation marks, and perhaps consider my other opinions from my long comment above, then break it into more bite-sized pieces, that would be cool. Not sure which comment you were asking me to respond to.

1-js/02-first-steps/05-types/article.md
```diff
@@ -62,14 +62,33 @@ Special numeric values formally belong to the "number" type. Of course they are
...
+`BigInt` type was recently added to the language to represent integers of arbitrary length.
+
+A `BigInt` is created by appending `n` to the end of an integer literal:
+
+```js
+// the "n" at the end means it's a BigInt
+const bigInt = 1234567890123456789012345678901234567890n;
+```
+
+As `BigInt` numbers are rarely needed, we devoted them a separate chapter <info:bigint>.
```
paroche: Suggest "we devoted a separate chapter to them...". (as is is ungrammatical).

1-js/05-data-types/02-number/article.md
```diff
@@ -29,14 +33,13 @@ In other words, `"e"` multiplies the number by `1` with the given zeroes count.
...
 
-
 Now let's write something very small. Say, 1 microsecond (one millionth of a second):
 
 ```js
 let ms = 0.000001;
 ```
 
-Just like before, using `"e"` can help. If we'd like to avoid writing the zeroes explicitly, we could say:
+Just like before, using `"e"` can help. If we'd like to avoid writing the zeroes explicitly, we could say the same as:
```
paroche: Recommend "say the same thing as:". Or, maybe better, just "say instead:"

1-js/05-data-types/02-number/article.md
```diff
@@ -271,13 +274,11 @@ JavaScript doesn't trigger an error in such events. It does its best to fit the
 ```smart header="Two zeroes"
 Another funny consequence of the internal representation of numbers is the existence of two zeroes: `0` and `-0`.
 
-That's because a sign is represented by a single bit, so every number can be positive or negative, including a zero.
+That's because a sign is represented by a single bit, so it can be set or not set for any number including a zero.
```
paroche: I prefer the original. Or, to get more wordy: "...a single bit which can be set for any number, so every number can be positive or negative, including zero". But I find it pretty clear the way it is (or was).

To me "set" does not necessarily mean "set to '1'" -- it could also be set to '0'. So I don't see "set or not set" as being necessary.

1-js/05-data-types/02-number/article.md
```diff
@@ -409,10 +410,10 @@ There are more functions and constants in `Math` object, including trigonometry,
 
 ## Summary
 
-To write big numbers:
+To write numbers with many zeroes:
 
-- Append `"e"` with the zeroes count to the number. Like: `123e6` is `123` with 6 zeroes.
-- A negative number after `"e"` causes the number to be divided by 1 with given zeroes. That's for one-millionth or such.
+- Append `"e"` with the zeroes count to the number. Like: `123e6` is the same as `123` with 6 zeroes `123000000`.
```
paroche: I thought it was OK the way it was, but if you're going to do it this way, it would be better with a colon after "zeroes":

"the same as `123` with 6 zeroes: `12300000`.

1-js/05-data-types/02-number/article.md
```diff
@@ -409,10 +410,10 @@ There are more functions and constants in `Math` object, including trigonometry,
...
+- A negative number after `"e"` causes the number to be divided by 1 with given zeroes. E.g. `123e-6` means `0.000123` (`123` millionth).
```
paroche: (`123` millionth**s**)

1-js/07-object-properties/02-property-accessors/article.md
```diff
@@ -94,11 +94,7 @@ alert(user.name); // Alice
 alert(user.surname); // Cooper
 ```
 
-As the result, we have a "virtual" property `fullName`. It is readable and writable, but in fact does not exist.
-
-```smart header="No way to handle `delete`"
-There's no similar method to handle deletion of an accessor property. Only getter/setter methods may exist.
-```
+As the result, we have a "virtual" property `fullName`. It is readable and writable.
```
paroche: I kind of liked the "in fact does not exist" part. Though that is already implied by "virtual", a bit of emphasis can be helpful. (Though I would have been tempted to rephrase it as "but does not in fact exist"). Without it, "It is readable and writable" seems to just kind of hang there.

2-ui/1-document/07-modifying-document/12-sort-table/solution.md
```diff
@@ -2,18 +2,18 @@ The solution is short, yet may look a bit tricky, so here I provide it with exte
 
 
 ```js
-let sortedRows = Array.from(table.rows)
-  .slice(1)
-  .sort((rowA, rowB) => rowA.cells[0].innerHTML > rowB.cells[0].innerHTML ? 1 : -1);
+let sortedRows = Array.from(table.tBodies[0].rows) // (1)
```
paroche: I like it. 😎

2-ui/1-document/11-coordinates/article.md
```diff
@@ -75,17 +75,17 @@ Please note:
...
 
 ![](coordinates-negative.svg)
 
-The rectangle starts at its bottom-right corner and then spans left/up, as negative `width/height` lead it backwards by coordinates.
-
-As you can see, `left/top` are not `x/y` here. So these are actually not duplicates. Their formula can be adjusted to cover negative `width/height`, that's simple enough, but rarely needed, as the result of `elem.getBoundingClientRect()` always has positive width/height.
+As you can see, `left/top` do not equal `x/y` in such case.
 
-Here we mention negative `width/height` only for you to understand why these seemingly duplicate properties are not actually duplicates.
+In practice though, `elem.getBoundingClientRect()` always returns positive width/height, here we mention negative `width/height` only for you to understand why these seemingly duplicate properties are not actually duplicates.
```
paroche: If keeping it all as one sentence, recommend moving the comma: "In practice, though `elem....` . 

Either that or change later part to "...width/height. Here we ....". And maybe have commas on both sides of 'though': 

"In practice, though, `elem...` always returns positive width/height. Here we mention ..." 

That would be my first choice.

2-ui/3-event-details/3-mousemove-mouseover-mouseout-mouseenter-mouseleave/2-hoverintent/source.view/style.css
```diff
@@ -20,9 +24,15 @@
   top: 0;
 }
 
-.tooltip {
+#tooltip {
```
paroche: I **thought** the styling had suddenly changed!

2-ui/99-ui-misc/01-mutation-observer/article.md
```diff
@@ -101,15 +101,15 @@ So, `MutationObserver` allows to react on any changes within DOM subtree.
...
 
 Naturally, the third-party script provides no mechanisms to remove it.
 
-Using `MutationObserver`, we can detect when such element appears in our DOM and remove it. While leaving the useful functionality intact. Surely though, creators of that script won't be happy that you took their useful stuff and removed the ads.
+Using `MutationObserver`, we can detect when the unwanted element appears in our DOM and remove it.
 
 There are other situations when a third-party script adds something into our document, and we'd like to detect, when it happens, to adapt our page, dynamically resize something etc.
 
-`MutationObserver` can easily handle this.
+`MutationObserver` allows to implement this.
```
paroche: " allows **us** to implement this"

5-network/02-formdata/article.md
```diff
@@ -54,14 +54,14 @@ In this example, the server code is not presented, as it's beyound our scope. Th
...
+- `formData.append(name, blob, fileName)` - add a field as if it were `<input type="file">`, the third argument `fileName` sets file name (not form field name), as it were a name of the file in user's filesystem,
 - `formData.delete(name)` - remove the field with the given `name`,
 - `formData.get(name)` - get the value of the field with the given `name`,
 - `formData.has(name)` - if there exists a field with the given `name`, returns `true`, otherwise `false`
 
 A form is technically allowed to have many fields with the same `name`, so multiple calls to `append` add more same-named fields.
 
-There's also method `set`, with the same syntax as `append`. The difference is that `.set` removes all fields with the given `name`, and then appends a new field. So it makes sure there's only field with such `name`, the rest is just like `append`:
+There's also method `set`, with the same syntax as `append`. The difference is that `.set` removes all fields with the given `name`, and then appends a new field. So it makes sure there's only one field with such `name`, the rest is just like `append`:
```
paroche: I think 

'only one field with such **a** `name`. The rest is ..."

or 'only one field with such **a** `name`; the rest is..."

or 'only one field with such a `name`, then the rest is..."

would be better. Probably prefer the last option.

PR #1613, opened 2019-11-18
add consistency

dagolinuxoid:
quick description:
before: .catch(f) is the same as promise.then(null, f)
after: .catch(f) is the same as .then(null, f)


dagolinuxoid:
I like the chapter about Promises very much. It's well written and even better than Jake's article.

iliakan:
thanks

PR #1614, opened 2019-11-18
Update solution.md

spaceinvadev:
Changed the reference to each item of `arr` since it was previously defined as the variable `word` (for...of loop) in the code block above. So, to be consistent, the same reference should be used when splitting the definition of `sorted` in multiple lines.


spaceinvadev:
Please disregard the last two commits as those changes will be proposed via another PR.

iliakan:
Maybe close this PR then?

spaceinvadev:
Sorry for the confusion, @iliakan. The changes proposed in this PR still holds so you can review to consider merging them. What I meant by `disregard the last two commits as those changes will be proposed via another PR.` was that in the last two commits I delivered additional changes proposal that I was supposed to do in a separate PR. But, as I said, the initial changes proposed in this PR are subject to review. Thanks.

PR #1625, opened 2019-11-24
Update article.md

generally23:
in the control control forms section, fixed mistake selectionEnd and selectionStart were both said to be  start position


lex111:
@generally23 good catch, thank you!

PR #1628, opened 2019-11-25
Update article.md

dashaezhova:
Removed duplicated `also`.


lex111:
@dashaezhova good catch, thanks!

PR #1632, opened 2019-11-26
Fixing urls to start with 'https'

### LINKED ISSUE #1622: The link doesn't follow the url
In this page the first task. Clicking on the link doesn't follow to w3.org. Something is not right.
https://javascript.info/default-browser-action

masterashu:
On the following page https://javascript.info/default-browser-action#catch-links-in-the-element 
the link will not redirect the page as expected. This happens because the links are http(not secure)
and most modern browsers block request redirecting to http links from https sites.
>Chrome Developer Console:
image

closes #1622


lex111:
@masterashu thank you!

PR #1636, opened 2019-11-28
Remove solution code

koala-lava:
The task source code(sandbox for the task) already contains the solution. Replaced the solution with "Your code".


iliakan:
thanks fixed

PR #1639, opened 2019-11-30
Create task2-solution.md

juuni26:
Adding more problem related arrow function to deepen understanding.


iliakan:
This supposed to be one more task, right?
You should structure it the same way as an existing one: a task resides in a folder.

juuni26:
thanks for the feedback. let me know if there is more that i can help 😄

iliakan:
Is that supposed to be a one more standalone task?

juuni26:
sorry my bad. didn't realize it until before

iliakan:
I'm sorry, that doesn't look good.

I don't see why a variable is needed. Maybe you could make the task simpler.

iliakan:
Closing this due to many mistakes in English language and in file names.

PR #1640, opened 2019-11-30
Fix code

fxcoudert:
- Code was not running because variable is named `theBiggestInt`, not `biggestInt`
- Naming the variable `biggestInt` when it is a BigInt (and not, by definition, the “biggest”) is confusing


iliakan:
This content was rewritten, sorry!

PR #1641, opened 2019-11-30
Remove unnecessary space

uxitten:


iliakan:
We put such spaces everywhere in alerts for readability.

PR #1644, opened 2019-11-30
Private method definition

spaceinvadev:
The current approach of defining the private method #checkWater leads to a SyntaxError (since browsers don't recognize that syntax for private methods just yet). In order to get the expected error message, such a method should be defined as private property with a function assigned to it (at the end of the day, private methods are just immutable own private fields which are functions).


iliakan:
There's a note about that, stating that the feature is not yet supported in engines.

PR #1649, opened 2019-12-02
Update article.md

CyberMew:


iliakan:
If you give a lot of examples, that doesn't help.

People won't read them, so I can't accept this. 

Please if you want to add something, let the overall lines count stay below 9.

iliakan:
Also, remainders of negative numbers are valid math-wise, but they are VERY rare in real life.

PR #1652, opened 2019-12-02
Add clarification

CyberMew:
A lot of people confuse remainder operator with modulo operator. This line tells the reader of such existence and that they are not the same.


iliakan:
That sounds REALLY confusing for those people who're not familiar with both operators.

Are you sure we should talk about an operator that's not in JavaScript at all?

CyberMew:
I agree too, but it is a useful tidbit to know.. 
How about adding an expandable Additional/Advanced Information box instead? This way without overwhelming new programmers, programmers with background with other languages can also learn more about the language by expanding the box. We can add tips or more advanced information in other places as well.

iliakan:
That information seems too much extra. 

You're talking about an operator that's not in JS. There are many operators in different languages :)
We need a solid basement, a good reason to tell people about them, I can't see such reason here sorry.

PR #1654, opened 2019-12-05
Fix wrong answer

ogaclejapan:
May be wrong `open` flags with the [IDBKeyRange interface](https://www.w3.org/TR/IndexedDB/#keyrange)

> 4.7. The IDBKeyRange interface
> ...
>  If open is true, key is not included in the range. 

The default value is false, so articles may be wrong.

Thank you for awesome articles! 
It was very helpful :+1:


iliakan:
Thanks!

PR #1658, opened 2019-12-06
Fix incorrect link address

LaievskyiSasha:


lex111:
@LaievskyiSasha thank you!

PR #1660, opened 2019-12-08
Make minor grammar corrections/updates to async/promise-basics

jchue:


1-js/11-async/02-promise-basics/article.md
```diff
@@ -31,9 +31,9 @@ When the executor obtains the result, be it soon or late - doesn't matter, it sh
 - `resolve(value)` — if the job finished successfully, with result `value`.
 - `reject(error)` — if an error occurred, `error` is the error object.
 
-So to summarize: the executor runs automatically, it should do a job and then call either `resolve` or `reject`.
+So to summarize: the executor runs automatically. It performs a job and then calls either `resolve` or `reject`.
```
iliakan: Is the original variant wrong?

paroche: Not very wrong. Probably wouldn't have triggered a PR from me (I hadn't read this article yet), but the changed version is a bit better. However, I would offer another alternative:

"...runs automatically and attempts to do a job. When it is finished with the attempt it calls `resolve` if it was successful or `reject` if there was an error."

If it's not too wordy.

jchue:
Updated

iliakan:
@paroche could you please advise, should I merge this PR or not?

paroche:
I would say yes --  from my review of them just now, looks like all the changes are improvements to a greater or lesser degree. I prefer my suggested wording on line 34, but that could be handled later if desired.

iliakan:
@paroche thanks, merged then!

PR #1661, opened 2019-12-08
Make minor grammar corrections/updates to async/promise-chaining

jchue:

### DIFF
--- a/1-js/11-async/03-promise-chaining/article.md
+++ b/1-js/11-async/03-promise-chaining/article.md
@@ -287,7 +287,7 @@ fetch('/article/promise-chaining/user.json')
   });
 ```
 
-The code works; see comments about the details. However, there's a potential problem in it, a typical error of those who begin to use promises.
+The code works; see comments about the details. However, there's a potential problem in it, a typical error for those who begin to use promises.
 
 Look at the line `(*)`: how can we do something *after* the avatar has finished showing and gets removed? For instance, we'd like to show a form for editing that user or something else. As of now, there's no way.
 
@@ -321,7 +321,7 @@ fetch('/article/promise-chaining/user.json')
 
 That is, the `.then` handler in line `(*)` now returns `new Promise`, that becomes settled only after the call of `resolve(githubUser)` in `setTimeout` `(**)`. The next `.then` in the chain will wait for that.
 
-As a good practice, an asynchronous action should always return a promise. That makes it possible to plan actions after it. Even if we don't plan to extend the chain now, we may need it later.
+As a good practice, an asynchronous action should always return a promise. That makes it possible to plan actions after it; even if we don't plan to extend the chain now, we may need it later.
 
 Finally, we can split the code into reusable functions:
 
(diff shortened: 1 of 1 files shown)


1-js/11-async/03-promise-chaining/article.md
```diff
@@ -287,7 +287,7 @@ fetch('/article/promise-chaining/user.json')
   });
 ```
 
-The code works, see comments about the details. However, there's a potential problem in it, a typical error of those who begin to use promises.
+The code works; see comments about the details. However, there's a potential problem in it, a typical error of those who begin to use promises.
```
paroche: Change OK. I would say "a typical error **for** those who begin..." ("for" rather than "of"). But small diff.

1-js/11-async/03-promise-chaining/article.md
```diff
@@ -319,13 +319,9 @@ fetch('/article/promise-chaining/user.json')
...
 
-That is, `.then` handler in line `(*)` now returns `new Promise`, that becomes settled only after the call of `resolve(githubUser)` in `setTimeout` `(**)`.
+That is, the `.then` handler in line `(*)` now returns `new Promise`, that becomes settled only after the call of `resolve(githubUser)` in `setTimeout` `(**)`. The next `.then` in the chain will wait for that.
 
-The next `.then` in chain will wait for that.
-
-As a good practice, an asynchronous action should always return a promise.
-
-That makes it possible to plan actions after it. Even if we don't plan to extend the chain now, we may need it later.
+As a good practice, an asynchronous action should always return a promise. That makes it possible to plan actions after it. Even if we don't plan to extend the chain now, we may need it later.
```
paroche: Small change to addition might be to use ";" (or, "--", though I prefer ";" here) instead of ".", like:

 "...after it; even if we...".

iliakan:
@paroche should we merge it in its current state?

paroche:
Would be OK to merge as is. I had some small suggestions but they could be done later, or not at all.

iliakan:
@paroche that means I'm unable to resolve the matter, as it's not about JavaScript, but rather about the English language, the wording and syntax.

So I kindly ask you to take a look and merge if you deem appropriate :)

paroche:
OK. Just seems like you've asked me to look at things before without "assigning" them to me. I guess it's the "merge if I deem appropriate" part that might be different. 

Though with these more extensive changes where I agree with some, not with others, and for others might have a third (and/or forth) suggestion, I'm not sure how to proceed -- don't think I can push it piece-meal or make revisions (except after pushing it all). And may want to give the author of the PR a chance to make revisions or comments first. So then I just comment.

iliakan:
@paroche you use create a review and "request changes" using the button in the right-upper corner, right?

Then the PR author should review and update their PR.

1-js/11-async/03-promise-chaining/article.md
jchue: Agreed.

jchue:
Updated

paroche:
OK. Too bad approving it piece-meal isn't an option (or doesn't seem to be). Seems like would be easier for long series of edits in one PR.

jchue:
Agreed.

paroche:
That was in response to Ilya's last comment, but I'm glad you agree. Is harder to keep track this way and it can lead to a lot of redundancy in reviewing changes.

PR #1665, opened 2019-12-08
Make minor grammar corrections/updates to async/microtask-queue

jchue:


1-js/11-async/07-microtask-queue/article.md
```diff
@@ -72,7 +72,7 @@ promise.catch(err => alert('caught'));
 window.addEventListener('unhandledrejection', event => alert(event.reason));
 ```
 
-...But if we forget to add `.catch`, then, after the microtask queue is empty, the engine triggers the event:
+But if we forget to add `.catch`, then after the microtask queue is empty, the engine triggers the event:
```
paroche: I like it better with the comma. Or, take out both of the last two commas: "`catch`, then after the microtask queue is empty the engine triggers the event:". But prefer it w/ all the commas as was.

1-js/11-async/07-microtask-queue/article.md
```diff
@@ -93,20 +93,20 @@ setTimeout(() => promise.catch(err => alert('caught')), 1000);
 window.addEventListener('unhandledrejection', event => alert(event.reason));
 ```
 
-Now, if you run it, we'll see `Promise Failed!` first and then `caught`. 
+Now if we run it, we'll see `Promise Failed!` first and then `caught`.
```
paroche: I would prefer the comma be left as was, but changing "you" to "we" makes sense.

1-js/11-async/07-microtask-queue/article.md
```diff
@@ -93,20 +93,20 @@ setTimeout(() => promise.catch(err => alert('caught')), 1000);
...
 
-If we didn't know about the microtasks queue, we could wonder: "Why did `unhandledrejection` handler run? We did catch the error!".
+If we didn't know about the microtasks queue, we could wonder: "Why did `unhandledrejection` handler run? We did catch and handle the error!"
 
-But now we understand that `unhandledrejection` is generated when the microtask queue is complete: the engine examines promises and, if any of them is in "rejected" state, then the event triggers.
+But now we understand that `unhandledrejection` is generated once the microtask queue is complete. The engine examines promises and, if any of them is in the "rejected" state, then the event triggers.
```
paroche: I prefer the original "when" over "once", though it's a pretty minor difference. 

Adding "the" is almost certainly better.

If the second sentence is an expansion/definition of the first, I think the ":" is OK. Or could use ";" or "--". Would prefer those over starting a new sentence.

jchue:
Updated

1-js/11-async/07-microtask-queue/article.md
```diff
@@ -93,11 +93,11 @@ setTimeout(() => promise.catch(err => alert('caught')), 1000);
 window.addEventListener('unhandledrejection', event => alert(event.reason));
 ```
 
-Now if we run it, we'll see `Promise Failed!` first and then `caught`.
+Now, if we run it, we'll see `Promise Failed!` first and then `caught`.
 
 If we didn't know about the microtasks queue, we could wonder: "Why did `unhandledrejection` handler run? We did catch and handle the error!"
 
-But now we understand that `unhandledrejection` is generated once the microtask queue is complete. The engine examines promises and, if any of them is in the "rejected" state, then the event triggers.
+But now we understand that `unhandledrejection` is generated when the microtask queue is complete: The engine examines promises and, if any of them is in the "rejected" state, then the event triggers.
```
paroche: Style guides may differ on this, but I strongly prefer to not capitalize after the colon here: "...is complete; the engine examines...".

jchue:
Updated

iliakan:
Thanks!

PR #1672, opened 2019-12-11
Fix grammar and construction in 2.1.10

galacemiguel:
There were various grammatical and construction-related errors in the [Window sizes and scrolling article](https://javascript.info/size-and-scroll-window). I did my best to correct them.

Please let me know if I made a mistake myself! Thank you!


2-ui/1-document/10-size-and-scroll-window/article.md
```diff
@@ -2,11 +2,11 @@
 
 How do we find the width and height of the browser window? How do we get the full width and height of the document, including the scrolled out part? How do we scroll the page using JavaScript?
 
-For most such requests, we can use the root document element `document.documentElement`, that corresponds to the `<html>` tag. But there are additional methods and peculiarities important enough to consider.
+For these, we can use the root document element `document.documentElement`, which corresponds to the `<html>` tag. But there are additional methods and peculiarities to consider.
```
paroche: I would leave this as is, or maybe replace "most such requests," with "for this type of information" (probably a little better w/out comma). And I prefer "that" over "which".

2-ui/1-document/10-size-and-scroll-window/article.md
```diff
@@ -16,33 +16,33 @@ For instance, this button shows the height of your window:
...
+Browsers also support properties like `window.innerWidth/innerHeight`. They look like what we want, so why not to use them instead?
 
-If there exists a scrollbar, and it occupies some space, `clientWidth/clientHeight` provide the width/height without it (subtract it). In other words, they return width/height of the visible part of the document, available for the content.
+If there exists a scrollbar, and it occupies some space, `clientWidth/clientHeight` provide the width/height without it (subtract it). In other words, they return the width/height of the visible part of the document, available for the content.
 
-...And `window.innerWidth/innerHeight` include the scrollbar.
+`window.innerWidth/innerHeight` includes the scrollbar.
 
-If there's a scrollbar, and it occupies some space, then these two lines show different values:
+If there's a scrollbar, and it occupies some space, then these two lines return different values:
```
paroche: I would stick with "show", especially since "return" has special meaning.

2-ui/1-document/10-size-and-scroll-window/article.md
```diff
@@ -16,33 +16,33 @@ For instance, this button shows the height of your window:
...
 ```js run
 alert( window.innerWidth ); // full window width
 alert( document.documentElement.clientWidth ); // window width minus the scrollbar
 ```
 
-In most cases we need the *available* window width: to draw or position something. That is: inside scrollbars if there are any. So we should use `documentElement.clientHeight/Width`.
+In most cases, we need the *available* window width—to draw or position something within scrollbars if there are any. So we should use `documentElement.clientHeight/clientWidth`.
```
paroche: Suggest: "In most cases we need the *available* window width, in order to draw or position something within the scrollbars (if there are any). So we should use `documentElement.clientHeight/clientWidth`."

2-ui/1-document/10-size-and-scroll-window/article.md
```diff
@@ -16,33 +16,33 @@ For instance, this button shows the height of your window:
...
 ````
 
 ```warn header="`DOCTYPE` is important"
-Please note: top-level geometry properties may work a little bit differently when there's no `<!DOCTYPE HTML>` in HTML. Odd things are possible.
+Please note: top-level geometry properties may work a little bit differently when there's no `<!DOCTYPE HTML>` in HTML.
```
paroche: I like the chatty little addition. Would leave it in.

2-ui/1-document/10-size-and-scroll-window/article.md
```diff
@@ -56,13 +56,13 @@ let scrollHeight = Math.max(
 alert('Full document height, with scrolled out part: ' + scrollHeight);
 ```
 
-Why so? Better don't ask. These inconsistencies come from ancient times, not a "smart" logic.
+Why so? Well, the gist of it is these inconsistencies come from ancient times.
```
paroche: Again, would leave as is. Humanizing. Even if 'not a "smart" logic' is a bit unusual, grammatically.

galacemiguel: It's too awkward I think. I found it to be inconsistent with the rest of the writing style in the article, especially when read top-down. Maybe we can rephrase it to be more idiomatic?

```suggestion
Why, you might ask? These inconsistencies come from ancient times, not bad design.
```

2-ui/1-document/10-size-and-scroll-window/article.md
galacemiguel: Good point. Will bring it back.

galacemiguel:
```suggestion
If there's a scrollbar, and it occupies some space, then these two lines show different values:
```

2-ui/1-document/10-size-and-scroll-window/article.md
paroche: Breaking the "that" clause out into a separate sentence doesn't seem as good to me as way it is in the original. Still favor my original suggestion.

2-ui/1-document/10-size-and-scroll-window/article.md
paroche: Could go with or without the comma there. No strong preference.

2-ui/1-document/10-size-and-scroll-window/article.md
paroche: Either way. I still prefer the original, though.

2-ui/1-document/10-size-and-scroll-window/article.md
paroche: Doesn't bother me the way it was, and the statement that "JavaScript can be quite finicky" seems to go beyond the scope of what is being discussed.

iliakan:
/me stands with @paroche on the review comments, please update the PR :)

iliakan:
@galacemiguel Could you please fix the PR according to the suggestions?

iliakan:
ping @galacemiguel 😮

galacemiguel:
> ping @galacemiguel 😮

Sorry, this one slipped my mind. 😅 

Getting to it now! @iliakan

2-ui/1-document/10-size-and-scroll-window/article.md
galacemiguel:
```suggestion
For this type of information, we can use the root document element `document.documentElement`, that corresponds to the `<html>` tag. But there are additional methods and peculiarities to consider.
```

2-ui/1-document/10-size-and-scroll-window/article.md
galacemiguel:
```suggestion
Please note: top-level geometry properties may work a little bit differently when there's no `<!DOCTYPE HTML>` in HTML. Odd things are possible.
```

2-ui/1-document/10-size-and-scroll-window/article.md
galacemiguel:
```suggestion
Why so? Better don't ask. These inconsistencies come from ancient times, not a "smart" logic.
```

galacemiguel:
Hi, @iliakan! I've committed https://github.com/javascript-tutorial/en.javascript.info/pull/1672/commits/d3ad4c3d82db101b51da6f41de94aa351db2abfe to revert and apply some changes based on the suggestions.

Let me know if there's anything I missed.

iliakan:
Wow! So I've got to review it again! ;)

iliakan:
Great, thanks!

PR #1673, opened 2019-12-14
Updated solution to task 10 of "Modifying the document"

DrTwstr:
@iliakan 

Corrected the solution to task 10 of Modifying the document. The current code is missing   `.slice(1)` which is important to avoid sorting the header along with the names. However, the solution is correct in the task sandbox.

### DIFF
--- a/2-ui/1-document/07-modifying-document/12-sort-table/solution.md
+++ b/2-ui/1-document/07-modifying-document/12-sort-table/solution.md
@@ -3,16 +3,18 @@ The solution is short, yet may look a bit tricky, so here I provide it with exte
 
 ```js
 let sortedRows = Array.from(table.tBodies[0].rows) // (1)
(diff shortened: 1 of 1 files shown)


iliakan:
The header is not in tbody, it should be in thead...

DrTwstr:
@iliakan 

The following is the html for the table as it is in the sandbox:
  ```
<table id="table">
    <tr>
      <th>Name</th>
      <th>Surname</th>
      <th>Age</th>
    </tr>
    <tr>
      <td>John</td>
      <td>Smith</td>
      <td>10</td>
    </tr>
    <tr>
      <td>Pete</td>
      <td>Brown</td>
      <td>15</td>
    </tr>
    <tr>
      <td>Ann</td>
      <td>Lee</td>
      <td>5</td>
    </tr>
  </table>
```

The headers are also selected when you do `table.tBodies[0].rows`.
There is no `<thead>` in the html.

iliakan:
Hopefully fixed?

DrTwstr:
Yes, it should work better now.

PR #1675, opened 2019-12-15
wiki link corrected

hy505:
A Wiki link for reference had ``` " " ```  around it.  which led to a broken link. it's now updated


lex111:
Already fixed, but thanks!

PR #1679, opened 2019-12-18
fix: type error of 1-js/03-code-quality/05-testing-mocha/article.md

leviding:
image


lex111:
@leviding seems legit, thanks!

PR #1682, opened 2019-12-20
remove incorrect info about synchronous postMessage

nylki:
The text in the info-box is incorrect. `postMessage` is not synchronous, but indeed async.
It may have been synchronous in IE8 and lower as some resources claim, but it is not on current supported browsers.

An easy check to see that it is async, is to set a breakpoint, eg. via `debugger`, in the `onmessage` event handler and check the call stack.  here is a screenshot of the chrome debugger call stack:

postMessage_async

It also contradicts other documentation like [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#Notes).

### DIFF
--- a/3-frames-and-windows/03-cross-window-communication/article.md
+++ b/3-frames-and-windows/03-cross-window-communication/article.md
@@ -335,10 +335,6 @@ The full example:
 
 [codetabs src="postmessage" height=120]
 
-```smart header="There's no delay"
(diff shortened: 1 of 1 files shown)


iliakan:
Actually, the event is kind-of sync, faster than `setTimeout`. But I guess, these words are a bit wrong, that's a bit peculiar in-depth matter, requires much more explanations, so removing it totally.

PR #1683, opened 2019-12-20
Add flat/flatMap array methods

spaceinvadev:
I added the new array methods: `flat()` and `flatMap()`. I tried my best to align with the overall writing style of the site.

P.S. I'm not 100% sure that the reference links (MDN) work as intended. This due to the internal URL handling convention (mdn:js/Array/...). I hope it works!


iliakan:
Sorry for the delay, was busy rewriting some stuff, making it simpler.

These methods are kind of rare, maybe make their explanation really short, but put references to docs instead?

iliakan:
Closed due to inactivity.

PR #1690, opened 2019-12-26
fix run-on sentence

KennethKinLum:


iliakan:
Replace `,` with `;` ?  Why? Not sure that's needed at all.

iliakan:
Please explain if you're sure =)

KennethKinLum:
in English, you can't say

> the ice cream is so good, i ate all of it.

because it is two verbs: "is" and "ate" in one sentence, without using a conjunction.

So either use a conjunction, or use ";" to connect them.  Otherwise, it is called a [run-on sentence](https://www.grammarly.com/blog/run-on-sentence-basics/).

So 

> They don't pass the result to each other, instead they process it independently.

"instead" is an adverb, not a conjunction.

KennethKinLum:
how do I reopen this pull request... I don't see such a button here.

iliakan:
@paroche please review this.

paroche:
I do think it's better with the ";".

KennethKinLum:
hm, I remember 3 ways to change a run-on sentence: make it 2 sentences, or use a semicolon, or use a conjunction.

paroche:
I don't know exactly what the grammarians say, and they may not all agree with each other, but for some sentences with two independent clauses, they can, and even should be separated by a comma. 

For instance:

"The ice cream was so good I ate all of it."

Is fine even without even a comma. Gives more of a "rushed" feeling to the sentence, which may be what you want*. 

Or you could use "that" where the comma would have been. But it's not necessary.

And this is fine too:

"The ice cream was so good, I ate all of it."

You **could** also do this:

"The ice cream was so good. I ate all of it."

But that introduces a pause that changes the nuance of the sentence. Perhaps for dramatic or slightly humorous effect.

There are a lot of examples of sentences with multiple independent clauses that could be separated by commas, or semicolons, or made into separate sentences. Sometimes a semicolon or separate sentence is better, but not always. It's a matter of feel (and usage).

So,

"Mary's lamb really loves her, she's its favorite person."

Is OK. Could use a semicolon instead of the comma, but it changes it, separates the thoughts a bit more, which may not be exactly what you want. With the comma has more continuity, flows more directly from the first clause, but is still easy to read (at least for me).


* I would not call it a run-on sentence but if I just kept typing without punctuation even if I was using conjunctions then that would be a run on sentence but those can get really hard to follow after a while because of the dearth of phrasing and breathing points that help you know where one thought ends and the next begins and that is really hard to process unless you really take your time and figure it out but usually it's something that you want to avoid if at all possible because it just isn't kind to your reader and who are you writing for if not your reader well perhaps for yourself but don't you want your own thoughts to be organized with phrasing and all that because if not then it may be hard to get certain things done like programming and stuff.

iliakan:
@paroche personally I feel like this PR is unnecessary.

I merged it because you said I should, and I trust in your English knowledge.
Let me know if I should keep the change or not.

KennethKinLum:
> The ice cream was so good I ate all of it.

actually, the line above sounds more like 

> The ice cream was so good that I ate all of it.

which actually should be ok... I don't know all the technical names of grammar... but this one should be ok.  Actually when we went to ESL and English classes, one of the first things they taught was run-on sentences, such as

Don't go, otherwise you will miss the TV show.

this would be a run-on sentence, as "otherwise" is not a conjunction.
But some other constructs, like

It was raining, so I didn't go.   (this is ok... so is a conjunction)
The movie was so good that he watched it twice.
I went outside so that the dog can walk with me.  (ok too)
I bought a fountain pen and some notebooks in order that he can practice calligraphy. (should be ok too).

These I don't know the formal grammatical names but they should be ok... I am checking...

paroche:
As you were asking originally, the change was not necessary. I did think that it was a little bit better with the semicolon, and I still think that, but just barely. If you'd rather go back to the comma, I wouldn't argue. 

Actually, I have more of an issue with the sentence preceding it, which is a bit awkward (perhaps should say something like: "What we did here was just set up several handlers for one promise."?), but I won't worry about that any further until I read the article (I'm off studying other things right now) so I understand better exactly what you're saying.

paroche:
kennethkinlum:

Yes, as I think I said, with or without the comma, "that" is implied.

I'm afraid you can't always believe what they tell you in ESL class -- for instance, I remember a German friend who was surprised to find out that native English speakers frequently end sentences with prepositions, even though she was taught that you never do. (And there's nothing wrong with ending a sentence in a preposition in general -- sometimes it's better if you don't, but other times it's very stilted and awkward if your try to rearrange the sentence to avoid it). 

"Don't go, otherwise you will miss the TV show."

 I don't think the above is wrong because "otherwise" is not a conjunction -- you could say  "You should stay, otherwise you'll miss the TV show" and that would be fine. The biggest problem is that "otherwise" implies the contrary/contrasting case -- it doesn't mean "if you do", but in this context means "if you **don't**".

"I bought a fountain pen and some notebooks in order that he can practice calligraphy."

Not too good because of "in order that he" is awkward. Should maybe simply say "so that he". Usually "in order" is followed by "to".

But we're getting kind of away from JavaScript here :-)

PR #1694, opened 2019-12-27
change about how some code "waits" for something

KennethKinLum:
I can't think of a better way... somehow the idea of some code "waiting" for something is hard to think about. Is it by some kind of magical `.wait()` or by some kind of `sleep()` call?  I think the mechanism is: it just exists and waits to be called. I can't think of a better way to  put "wait for it" and "wait to be called" together.

### DIFF
--- a/1-js/11-async/02-promise-basics/article.md
+++ b/1-js/11-async/02-promise-basics/article.md
@@ -264,7 +264,7 @@ It's not exactly an alias of `then(f,f)` though. There are several important dif
 3. Last, but not least, `.finally(f)` is a more convenient syntax than `.then(f, f)`: no need to duplicate the function `f`.
 
 ````smart header="On settled promises handlers run immediately"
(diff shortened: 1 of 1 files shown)


iliakan:
@paroche is there anything to change here, how do you think?

jianlin:
I guess it is like when an object "draws" itself... `obj.draw()`  it is difficult to map the "wait" for something... `obj.wait()`?  It actually is just sitting idle and waiting for some code to invoke it... I guess it can mean ".then()` handlers not being called, waiting for it (the promise settling) -- or waiting for the promise to settle.

KennethKinLum:
I guess it is like when an object "draws" itself... `obj.draw()`  it is difficult to map the "wait" to something... `obj.wait()`?  It actually is just sitting idle and waiting for some code to invoke it... I guess it can mean ".then()` handlers not being called, waiting for it (the promise settling) -- or the 
".then()` handlers not being called, waiting for the promise to settle.

paroche:
I haven't read the article, but original seems OK, probably better, to me. Handlers are always waiting for things.

PR #1695, opened 2019-12-27
add a Node.js way to run JavaScript code

KennethKinLum:


iliakan:
We use browser in the tutorial as the demo tool, no need to specify other environment.

iliakan:
If we were to add node.js, more explanations needed. What is it, what's the difference versus a web-browser.

PR #1699, opened 2019-12-28
adding a precaution of using map[key]

KennethKinLum:


iliakan:
Thanks, added (rephrased).

PR #1710, opened 2020-01-16
Function return example update

mdizo:
Just a couple of nitpicks:
 - Updated code to check for ages 18 and up
 - Renamed let variable because debugger throws redeclaration syntax error


iliakan:
Acciddentally merged it. PR is wrong. `ageCheck` is incorrect var name, that won't work. Please check your code.

mdizo:
Ypu're right, my bad. Created a new pull request @ #1712

PR #1712, opened 2020-01-17
Correction to #1710

mdizo:


mdizo:
Never mind, just noticed you already corrected it. Thanks!

PR #1725, opened 2020-01-28
Update article.md

Beginnerds:
## Proposed changes
Corrected grammatical errors in article.

### DIFF
--- a/5-network/07-url/article.md
+++ b/5-network/07-url/article.md
@@ -3,9 +3,9 @@
 
 The built-in [URL](https://url.spec.whatwg.org/#api) class provides a convenient interface for creating and parsing URLs.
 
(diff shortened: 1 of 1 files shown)


Violet-Bora-Lee:
Could you sign the Contributor License Agreement?

cf. https://www.quora.com/Which-is-correct-A-URL-or-An-URL

iliakan:
@paroche Could you review and maybe merge this if everything's correct?

@Violet-Bora-Lee for typo fixes we can merge without CLA. That's because typo fixes are not objects of copyright, no need an explicit permission to "publish" them.

paroche:
I agree with the changes. The "a -> an" before a word starting with a vowel (or an "h") rule should only apply if the word actually starts with a vowel **sound**. Since "url" is generally pronounced starting with a "y" sound ("you-are-el"), if should be preceded with "a" rather than "an". (If it were generally pronounced more like "earl", then "an" would be appropriate). It's all about making it easier to say (even in your mind).

PR #1726, opened 2020-01-29
Fixing a typo

sohailashraf003:
Check if `pattern:Y` is immediately after `pattern:X`  and then the regex will check if Z is immediately after Y,  Like this it's more understandable.


lex111:
@sohailashraf003 thank you!

sohailashraf003:
Thanks @lex111 .

PR #1731, opened 2020-01-30
Add definition of Template Literals and MDN link

Violet-Bora-Lee:
There could be some people who google the keyword 'Template Literal(s)'.
So, I've added the definition of the term to better SEO.


iliakan:
It reads somewhat obscure now.

If you want to add a reference, then we can add a separate link in `# References` section.

P.S. There's no need in SEO for our tutorial, it's good enough.

PR #1738, opened 2020-02-07
Update article.md

paroche:
My proposed change may not be right, but "make a request to URL" was unclear, and it's `fetch` that's used below to make the request. If 'request to URL' has some special meaning, and is what you want to say, maybe some rephrasing is still in order.


iliakan:
Sorry for the delay.

paroche:
Oh yeah, I remember that one, though I had kind of stopped wondering about it  😓. 

But all's well that ends well, and all that. 😌

PR #1748, opened 2020-02-13
Fix translation

GeorgySerga:


lex111:
@GeorgySerga oops, thank you!

PR #1751, opened 2020-02-13
Either quotation marks or italicization

u-n-known:
Highlighting twice is unnecessary.

### DIFF
--- a/1-js/01-getting-started/1-intro/article.md
+++ b/1-js/01-getting-started/1-intro/article.md
@@ -4,7 +4,7 @@ Let's see what's so special about JavaScript, what we can achieve with it, and w
 
 ## What is JavaScript?
 
-*JavaScript* was initially created to *"make web pages alive"*.
+*JavaScript* was initially created to "make web pages alive".
 
 The programs in this language are called *scripts*. They can be written right in a web page's HTML and run automatically as the page loads.
 


iliakan:
You mean italic is bad there?

u-n-known:
@iliakan, the actual problem is that you can't figure out wether it's a cite or the apostrophes suggest some kind irony because you can not really make a web page alive, shortly because it's a personification. 

In the first case the apostrophes suffice, for the second I wouldn't even mark it up as it is kind of difficult to mark something up like irony, and if so it would've been sufficient to italicize just "alive" as I suppose you know so I don't think this was your intend. Or am I mistaken?

iliakan:
@paroche could you please advise here?

paroche:
Never caught my attention before (the one or two times I read it), but, now that it's come up, I think it would be a bit better with just the quotes (not apostrophes), without the italicizing.

u-n-known:
>  quotes (not apostrophes)

Oh, you're right.

iliakan:
Thanks!

paroche:
@u-n-known  
<i>quotes (not apostrophes)

Oh, you're right.</i>

Well, you got it in the title! :-)

And, of course, in JS itself they are almost interchangeable.

PR #1759, opened 2020-02-18
Gender pay gap noticed in recursion article.md

cycduck:
Alice's salary was $600 while the salaries of her male counterparts were $1000 to $2000. 

Updated Alice's salary to $1600 (average of all male salaries and rounded up for easy comprehension) to modernize the example in light of raised awareness about the gender pay gap.


iliakan:
Sure, let Alice be happy.

Issue #1743, opened 2020-02-09
Mac OS multiline Commenting shortcut is should use 'Option' instead of 'Shift'


nishantsinghcs:
In most editors, a line of code can be commented out by pressing the Ctrl+/ hotkey for a single-line comment and something like Ctrl+Shift+/ – for multiline comments (select a piece of code and press the hotkey). For Mac, try Cmd instead of Ctrl.

Should be --> 

In most editors, a line of code can be commented out by pressing the Ctrl+/ hotkey for a single-line comment and something like Ctrl+Shift+/ – for multiline comments (select a piece of code and press the hotkey). For Mac, try Cmd instead of Ctrl and Option instead of Shift.

iliakan:
Please submit a PR?

ktwolf123:
The syntax for the built-in remapping feature is OriginKey::DestinationKey. For example, a script consisting only of the following line would make the A key behave like the B key:

a::b
S↓
The above example does not alter the B key itself. The B key would continue to send the "b" keystroke unless you remap it to something else as shown in the following example:

a::b
b::a
S↓
The examples above use lowercase, which is recommended for most purposes because it also remaps the corresponding uppercase letters (that is, it will send uppercase when CapsLock is "on" or the Shift key is held down). By contrast, specifying an uppercase letter on the right side forces uppercase. For example, the following line would produce an uppercase B when you type either "a" or "A" (as long as CapsLock is off):

a::B
S↓
However, a remapping opposite to the one above would not work as one might expect, as a remapping never "releases" the modifier keys which are used to trigger it. For example, A::b is typically equivalent to A::B and ^a::b is equivalent to ^a::^b. This is because each remapping internally uses {Blind} to allow the key or key combination to be combined with other modifiers.

ktwolf123:
remapping

PR #1760, opened 2020-02-19
Mac OS multiline Commenting shortcut is should use 'Option' instead o…

nishantsinghcs:
…f 'Shift' #1743

Fix Submitted.


iliakan:
Thank you!

PR #1791, opened 2020-03-09
TypedArray only represents other classes.

amundo:
This may not be the best way to explain it, but I was mislead by the way that TypedArray was included in code examples as if it were an actual class. The list is a little unwieldy, but I don’t see anywhere else in this article that lists them all.


amundo:
Oops, I do see them listed now. So perhaps the list should just be removed here, but I still think some kind of explanation about `TypedArray` would help.

PR #1794, opened 2020-03-10
minor

qucumbah:


iliakan:
Error? Why?

qucumbah:
if you execute the code it throws:
ReferenceError: Must call super constructor in derived class before accessing 'this' or returning from derived constructor
the description even says "Here’s an example of such code (it doesn’t work – why? fix it?)"

iliakan:
Right, thanks!

iliakan:
There was an error there indeed.

PR #1799, opened 2020-03-13
Fix error in 1-js/11-async/05-promise-api/article.md

leviding:
The code is:

```js
if(!Promise.allSettled) {
  Promise.allSettled = function(promises) {
    return Promise.all(promises.map(p => Promise.resolve(p).then(value => ({
      state: 'fulfilled',
      value
    }), reason => ({
      state: 'rejected',
      reason
    }))));
  };
}
```

Some time ago, you update the code above, but did not update the corresponding text description.

Screen Shot 2020-03-13 at 4 04 59 PM


iliakan:
Thanks!

PR #1800, opened 2020-03-13
minor fixes

Violet-Bora-Lee:
- fix erratas
- fix wrong error message
- add 'run' on runnable example
- add explanation about window.onload property corresponding load event
- add comment about window.onstorage property corresponding storage event


iliakan:
Thanks. Clarifications about `on*` events are generally not needed though =)

PR #1818, opened 2020-03-25
manually listing the all possible keys can be prone to error

KennethKinLum:
manually listing the all possible keys can be prone to error and if we want to test for 4 or 5 items, then need to list all 24 or 120 possible keys.


iliakan:
The code demonstrates the faulty algorithm. It should be easy to understand and explicit, that's why.

PR #1819, opened 2020-03-25
Change elem.addEventListener(event, handler[, phase]) to elem.addEventListener(event, handler, [phase])

leviding:
RT.

### DIFF
--- a/2-ui/2-events/01-introduction-browser-events/article.md
+++ b/2-ui/2-events/01-introduction-browser-events/article.md
@@ -440,7 +440,7 @@ There are 3 ways to assign event handlers:
 
 1. HTML attribute: `onclick="..."`.
 2. DOM property: `elem.onclick = function`.
-3. Methods: `elem.addEventListener(event, handler[, phase])` to add, `removeEventListener` to remove.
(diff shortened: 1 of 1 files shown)


leviding:
I see that in the tutorial, somewhere optional args write in the form of someMethod(A, B, [C]), and somewhere write in the form of someMethod(A, B[, C]).

Thus, if it doesn't, you can just close this PR.

Thanks 👍

lex111:
@leviding the comma is also optional, look at MDN for an example - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

```
arr.find(callback(element[, index[, array]])[, thisArg])
```

It would be great to unify optional args! :thinking: 
And at the moment I will close this PR, but thanks for your efforts.

leviding:
Got it, thanks!

leviding:
It lines which number are 219 and 238 are written in the form of `, [`, maybe it should be fixed.

iliakan:
Personally, I prefer `param, [optional]` syntax, rather than `param[, optional]`.

The first one looks cleaner.

PR #1845, opened 2020-04-05
fix typo

baooab:
1. then we don't need Array.from, разумеется, не нужен. -> then we don't need Array.from any more.
2. delete the extra table header


lex111:
@baooab thanks!

PR #1846, opened 2020-04-05
fix typo

Ahmed-Magdy-S:
I'm not sure if that is a normal or a typo
so, take a look please
you talk about "min" but you say the pair "max, result" instead , I think it should be "min, result"


iliakan:
All correct.

PR #1850, opened 2020-04-07
Typo: Remove Duplicate Assignment

Ahmed-Magdy-S:
the first example, there is a duplicate assignment as a typo, so I fixed it.


iliakan:
Not typo, it's equality.

PR #1859, opened 2020-04-12
trick: use array destructuring to swap variables

wonderingabout:
credit for teaching me this trick :) https://www.typescriptlang.org/docs/handbook/variable-declarations.html#array-destructuring

### DIFF
--- a/1-js/05-data-types/10-destructuring-assignment/article.md
+++ b/1-js/05-data-types/10-destructuring-assignment/article.md
@@ -121,6 +121,23 @@ for (let [key, value] of user) {
 }
 ```
 ````
+
+### Swap variables
+
+It is also possible to use destructuring assignment to easily swap variables
+
+```js run
+let a = "1";
+let b = "2";
+alert(`${a}, ${b}`); // 1, 2
+
+[a, b] = [b, a];
+alert(`${a}, ${b}`); // 2, 1: successfully swapped!
+```
+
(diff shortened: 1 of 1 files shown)


1-js/05-data-types/10-destructuring-assignment/article.md
```diff
@@ -121,6 +121,44 @@ for (let [key, value] of user) {
...
+
+```js run
+let temp = a;
+a = b;
+b = temp;
+
+alert(`${a}, ${b}`); // 2, 1: successfully swapped!
+```
+
+But such code is long, unefficient, and error-prone. With deconstructing, we can do it in a much simpler way:
```
Dorus, line 147: unefficient -> inefficient

wonderingabout: fixed, ty :1st_place_medal:

iliakan:
Sorry, that trick does not deserve such a long explanation.

I'd say, swapping variables is a very rare thing in practice. Maybe make it small?

wonderingabout:
> Sorry, that trick does not deserve such a long explanation.
> 
> I'd say, swapping variables is a very rare thing in practice. We could add it though, if it were 3+ times smaller =)

sure, i'll submit a much smaller with a 1 commit rebasing long explanation, then a 1 commit short 
 explanation (so that we can compare short vs long PR)

wonderingabout:
what do you think of the new version @iliakan ?

iliakan:
:rocket:

PR #1878, opened 2020-05-01
Fix minor typo error

leviding:
Alert show:

```
{
  "name": "iliakan", "isAdmin": true
}
```

Not:

```
{
  "name": "iliakan", isAdmin: true
}
```
```


lex111:
@leviding thanks!

PR #1881, opened 2020-05-02
Typo Error in Description of "alert" 

thediveshsoni:
Repeated word in the description of alert


lex111:
@thediveshsoni good catch, thanks!

PR #1884, opened 2020-05-04
small phrase change

matjax:


matjax:
My first PR - woot!

iliakan:
Why? @paroche

paroche:
Because 'just yet' is generally only used in the negative:

"I'm not going home just yet" -- correct
"I am going home just yet" -- incorrect
"I'm going home right now" -- correct

The change suggested:

> In this chapter, we won't cover objects. We'll only be talking about primitives.

 is OK, though I would probably say instead:

> In this chapter, we won’t cover objects. For now we'll just be talking about primitives.

Issue #1872, opened 2020-04-25
needs explanation on the example about dynamic script


Violet-Bora-Lee:
Hi, Ilya and the Team! How's the COVID situation there? I hope all the members are fine.

I'm reviewing the translated article of https://javascript.info/script-async-defer.

All other examples have an explanation of what it demonstrating but this example has no explanation.

![image](https://user-images.githubusercontent.com/6451384/80273142-bc7db880-870a-11ea-9420-85f84aadabe8.png)

Could you add?

Violet-Bora-Lee:
Btw, I wonder the differences between loading and downloading the script in the article.

I think the two words have the same meaning but I want to verify.

iliakan:
Hi, I'm ok so far, but our team member (admin) living in Czech got Covid.  Quite a nasty thing he says. So take care.

Yes, loading/downloading is the same. 
Could you suggest a PR please?

Violet-Bora-Lee:
I'm not sure why you explicitly assign false on script.async.

I hope Alexey will be recovered soon!

iliakan:
Yeah, was unclear for me too some time ago.

By default, scripts added via DOM methods behave as if they had `async=true`.

That is, they execute in "first loaded - first run" order.

But if there's an explicit `script.async = false`, then they execute in page order.  If you create two scripts and append them to `document.body`, then the first that you add always runs first. Even if the second one is fully loaded, it waits.

iliakan:
I remember this was fixed?

PR #1886, opened 2020-05-05
#1872 modify example

Violet-Bora-Lee:
https://github.com/javascript-tutorial/en.javascript.info/issues/1872

Why we need to assign the `false` on `async` to load the scripts in page order is already mentioned in the example below. So, I think it's better to remove the example above.


iliakan:
Modified the example, thanks!

PR #1895, opened 2020-05-08
Adding missing 'pattern:'

Aakodal:
Adding missing 'pattern:' next to 'm' in the line "Without the flag `m`"


lex111:
@Aakodal thanks!

PR #1907, opened 2020-05-16
Confucius never said black cat quote

leeyc0:
Confucius never said anything about black cat

Ref: https://www.washingtonpost.com/world/2019/03/25/chinese-proverb-that-russia-cited-respond-mueller-report-does-not-appear-be-chinese-proverb/

### DIFF
--- a/1-js/03-code-quality/04-ninja-code/article.md
+++ b/1-js/03-code-quality/04-ninja-code/article.md
@@ -105,7 +105,7 @@ A quick read of such code becomes impossible. And when there's a typo... Ummm...
 ## Smart synonyms
 
 ```quote author="Laozi (Tao Te Ching)"
- he name that can be named is not the eternal name.
+The name that can be named is not the eternal name.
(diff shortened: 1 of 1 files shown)


iliakan:
Who said it then? Maybe we can replace it with another, real Chinese metaphor?

leeyc0:
Apparently this "quote" actually came from a Soviet drama. The character who said this "quote" claimed it came from Confucius, and the misunderstanding stuck since then.

iliakan:
What should be a proper replacement? :ponder:

leeyc0:
Unfortunately I am still looking for a proper replacement...

leeyc0:
I think the best one would still be Tao Te Ching, given on the context.
https://www.wussu.com/laotzu/laotzu01.html

名可名，非常名
The name that can be named is not the eternal name.

iliakan:
Sounds good! 🚀  🎸

PR #1925, opened 2020-05-25
typo

Li-YangZhong:


Li-YangZhong:
I double checked and found this may not be a typo. The two dots between '0..4' and '5..9' are used to indicte from '0 to 4' and from '5 to 9'. But is the short dash sign '-' better for this purpose or not? I am not sure. Please let me know.

iliakan:
Maybe rephrase it?

It means that up to 4 is rounded down, e.g. 1.4 → 1, while up from 5 is rounded up: 1.5 → 2.

PR #1928, opened 2020-05-28
google js guide link

tiagocastro070:
Accordingly to Google there is a newer and updated styleguide. The current one on javascript.info is outdated.

Should this pull request change the link on other languages too?


iliakan:
The translators should catch up, I hope.

Issue #1645, opened 2019-12-02
When to use const or let


CyberMew:
https://javascript.info/variables

I think it would be good to indicate which is more preferred, i.e. instead of just saying what it does, adding the _when_ to use will be better. For example _const_ would be preferred/recommend whenever possible because you can't modify the value by accident or prevent some other programmer from reusing it for other stuffs and it helps with debugging in the long run. It is not mainly just for constants most of the time, the article worded it out to be that way. Maybe under best practices section or something?

iliakan:
A good idea. Could you make a PR with such addition?

icncsx:
Hey @iliakan , can I work on this?

iliakan:
Here's an excerpt:

```
When a programmer is sure that a variable will never change, they can declare it with const to guarantee and clearly communicate that fact to everyone.
```

Surely you can try to improve the article, @icncsx!

moustachio-belvedere:
Is it also worth mentioning compiler optimisations that may occur with different variable types?

iliakan:
An article about optimizations and internals of JS engines could be useful.

icncsx:
@iliakan Better late than never right? I made a PR - albeit for a separate reason re: redeclaration syntax error for `let` and `const`. I can raise an issue if you'd like. See: #1931

PR #1931, opened 2020-05-30
Enhance variables article with mention of redeclaration

icncsx:
In the [Variables](https://javascript.info/variables) section, we warn that reassignment cannot happen w/ `let` and `const`,  but we don't mention anything about redeclaration (in the same scope) as being problematic. This PR includes that detail plus some grammatical and sentence fluency related additions.

For context, @iliakan, I was originally working to resolve issue #1645, but I think your excerpt does fine, so I only added one sentence after it.

```md
When a programmer is sure that a variable will never be reassigned nor redeclared, they can use `const` to clearly communicate that information to other developers which will help with maintaining and debugging the codebase in the long run.
```  

Let me know what you think!


iliakan:
Maybe a single small `warn` section would be better to tell about that. To separate it as a small note.

PR #1936, opened 2020-06-01
prototype-methods  shallow-clone remark

joaquinelio:
"fully identical shallow clone" seems contradictory in plain English.
It makes no much sense if you are learning, that's why I suggest to remark as a concept: 
identical 'shallow-clone' 
I wanted to put a reference to  <info:object-copy> 
but it's too weak there and I couldn't figure where to put it.


iliakan:
Truly identical shallow clone is all right imho.

The "equality" is defined differently, when we talk about shallow cloning vs deep cloning, so the phrase is correct.

iliakan:
I removed the confusing phrase with the "shallow" word though =)

joaquinelio:
@iliakan  Perfect!!!
You are right,  
It was correct before
and it WAS confusing. It is better now. No need to distract the student.

PR #1942, opened 2020-06-02
thisArg improvement

nagarjuna-immadi:
In standalone functions value of 'this' is 'window' object, not 'undefined'. Calling with default value doesn't produce an instant error, we will get a wrong result.


iliakan:
You forgot "use strict".

PR #1957, opened 2020-06-13
Add bottom and right properties for getCoords()

chanjsq:
The `createMessageUnder()` function below needs the `bottom` coordinate, but it doesn't in the result of `getCoords()`.


iliakan:
Why these are needed?

chanjsq:
Because the code below (`createMessageUnder`) is using them.

iliakan:
Oh, good catch then! =)

PR #1959, opened 2020-06-15
Fixed onupgradeneeded handling

xehpuk:
`openRequest.result.version` in `onupgradeneeded` is not the current DB version. It is the requested version, equal to `event.newVersion`. `event.oldVersion` contains the current version.


iliakan:
Thanks!

PR #1970, opened 2020-06-23
Update article.md

rromanenko:
Fixing a typo. There are 2 articles before a noun, I think one of them should be removed.


iliakan:
Thank you!

PR #1980, opened 2020-07-01
a little confused by this description

liuernan:
My proposed change may not be right, but local variable is only reachable when it is currently used, so is this a mistake ? 

> For instance, if there's an object in a *~~local~~global* variable, and that object has a property referencing another object, that object is considered reachable. And those that it references are also reachable. Detailed examples to follow

This also dose not fit the description above:

> 2. Any other value is considered reachable if it's reachable *from a root* by a reference or by a chain of references.

Please let me know if I made a mistake myself! Thank you!


iliakan:
Doesn't matter much here, but ok.

PR #1982, opened 2020-07-01
x is an operand at ( typeof x )

lumosmind:
naming of `x` changes depends on usage of `typeof`(as a function or operator)
1. As an operator: `typeof x`.  --> `x `is an **operand**
2. As a function: `typeof(x)`.   --> `x `is an **argument**


iliakan:
operand is approximately the same as argument for operators

lumosmind:
> operand is approximately the same as argument for operators

Yes but, you need to take into consider that "analogies hide truths".
What do you think about using **function** instead of **operator**.
This is a big sin  and Math gods will curse us:)

iliakan:
I don't see a real reason to change, sorry.

PR #1983, opened 2020-07-01
is 'if' operator?

lumosmind:
The title makes me ask "is `if `keyword operator or not?".
I think `if `is not operator because it does not return any value and it is called **if-statement** in [MDN ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/if...else)and [ECMA](https://tc39.es/ecma262/#sec-if-statement).


iliakan:
Yeah, better.

PR #1985, opened 2020-07-02
clarifying the example

lumosmind:
There is a conflict between explanation and example. 
_"Otherwise, if we omit parentheses, ... "_
But in the example of this words, there are some extra parentheses. these parentheses make the article difficult to understand.
Also I added result of example which is  without parentheses to see difference clearly.


iliakan:
Without parens, the code works like this.

PR #1992, opened 2020-07-05
Change an expression

minseokkang8571:
I thought about whether I choose "powers of 10" or "power of 10".
As a result, I choose "powers of 10" because it was in the document.

If there is a problem, tell me. :)


iliakan:
Added slightly modified fix. Thanks!

PR #1998, opened 2020-07-08
Update article.md in boolean conversion for falsy value

contact-ouyang:
Added -0 and 0n in boolean conversion description for falsy value


iliakan:
Please, just add `0n`, don't add `-0`, as it confuses people.

PR #1999, opened 2020-07-09
Fix a comment

zhang1pr:
The example says **doesn't run here** in then, but the top comment says the opposite.
Fix the comment.


iliakan:
Seems right, thanks!

PR #2005, opened 2020-07-12
Fix some confusing expression in chapter of optional chaining

leviding:
As title.


leviding:
Line 43:

```md
Further in this article, for brevity, we'll be saying that something "exists" if it's not `null` and not `undefined`.
```

It should be highlighted, I think.

leviding:
In line 67:

```
But if the `user` actually exists, then the further intermediate properties, such as `user.address` must exist.
```

Why `user.address` must exist? I think it should be explained further.

iliakan:
Made modifications there. Hope it's better.

PR #2008, opened 2020-07-16
Update solution.md

prakx1:
Updated code to return response json instead of waiting for it.


iliakan:
awaiting is better, it makes changing the code easier.

PR #2011, opened 2020-07-17
Browser LocalStorage capacity updated to 5 MB.

Pankajtanwarbanna:
Hi,

I think, max local storage size is 5 MB in modern browsers not 2MB. 

Ref  - https://web.dev/storage-for-the-web/


iliakan:
Okay =)

PR #2016, opened 2020-07-19
fix iterator explanation in the summary section

MarkAdell:


iliakan:
Here, the word 'next' is simpler to understand. Added parens.

PR #2017, opened 2020-07-19
Comma place

joaquinelio:
JUST a suggestion.  
I'm not sure about the correct use of the colon either.
The "correction" may be trivial (even wrong) for English. 
I HAD to do this change in the insanely long resulting sentence in Spanish. I asked Fundéu (Spanish academy) and things are more clear now.

Wrong about being wrong about being wrong.
I was studying
https://owl.purdue.edu/owl/general_writing/punctuation/commas/extended_rules_for_commas.html
Rules are different from Spanish,  not THAT different.  
So the suggestion. Again, it may be wrong.

### DIFF
--- a/1-js/04-object-basics/02-object-copy/article.md
+++ b/1-js/04-object-basics/02-object-copy/article.md
@@ -225,4 +225,4 @@ Objects are assigned and copied by reference. In other words, a variable stores 
 
 All operations via copied references (like adding/removing properties) are performed on the same single object.
 
-To make a "real copy" (a clone) we can use `Object.assign` for the so-called "shallow copy" (nested objects are copied by reference) or a "deep cloning" function, such as [_.cloneDeep(obj)](https://lodash.com/docs#cloneDeep).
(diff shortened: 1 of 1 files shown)


iliakan:
Semicolon shouldn't be there.

joaquinelio:
@iliakan It's OK you closed it. I'll stop doing comma suggestions, its not my language.

just a clarification, there is no semicolon. 
I put a colon to show the options shallow/deep (I saw you using it in other places) 
And a comma position, I thought it was in a wrong place.
Here it is not important, but in Spanish it's easy to loose track of thinking.

iliakan:
I'm quite sure there should be colon there, sorry.

iliakan:
We do have typos, just not here. Thanks!

joaquinelio:
...
...oh,
you dont mean colon but comma I guess...


> I'm quite sure there should be colon there, sorry.

> Semicolon shouldn't be there.

PR #2018, opened 2020-07-21
minor fixes: There is no character type.

JiSop:
minor fixes
[String](https://javascript.info/types#string) in **There is no character type.** 
> In JavaScript, there is no such type. There's only one type: `string`. A string may consist of only one character or many of them.

A string may consist of ~~only one~~ character or many of them.
->
A string may consist of **zero** character or many of them.


iliakan:
Added zero =)

PR #2021, opened 2020-07-21
to make binding more clear

lumosmind:
understanding 

> pre-bound value

  can be little bit hard.


iliakan:
which is THE reference maybe?

lumosmind:
> which is THE reference maybe?
OLD is more clear. THE can refer both of old and new user value.

iliakan:
"reference to old user object" → maybe more correct → "reference to THE old user object"?

lumosmind:
> "reference to old user object" → maybe more correct → "reference to THE old user object"?

I got it.

iliakan:
Thanks!

PR #2027, opened 2020-07-25
clarifying example code

lumosmind:
> An interesting question may arise in the example above: what’s the value of this inside set fullName(value)? Where are the properties this.name and this.surname written: into user or admin?
> 
> The answer is simple: this is not affected by prototypes at all.
> 
> No matter where the method is found: in an object or its prototype. In a method call, this is always the object before the dot.
> 
> So, the setter call admin.fullName= uses admin as this, not user.
> 
> That is actually a super-important thing, because we may have a big object with many methods, and have objects that inherit from it. And when the inheriting objects run the inherited methods, they will modify only their own states, not the state of the big object.

The example code doesn't show these concepts. Also, these additions can make readers ask questions before the explanation of the example code.


iliakan:
Thank you!

PR #2042, opened 2020-08-03
Use spread syntax

SCR-IR:
add use spread syntax


iliakan:
Covered https://javascript.info/rest-parameters-spread#get-a-new-copy-of-an-array-object

PR #2046, opened 2020-08-08
repeated word 2-3-6 article

joaquinelio:


joaquinelio:
Hi, @iliakan 
Not confident enough to put a PR, so I just ask down here.

In this article "pointer events",  you use "As of now" lines 19 & 229. 
We should translate it to something like "From this moment/Starting right now" (and translator did that). But it sounds wrong to me.
I suggested "al momento" something like "At the moment". We can say "Today" too.

The question, is this an inaccurate expression "As of now, browser supports this spec."
Or, I am wrong. Apologies if that's the case.
regards

iliakan:
Maybe @paroche could kindly give us a piece of advice?

paroche:
I think "As of now" is OK, though "These days", which I believe you've used before in similar circumstances, would probably be better. It's a bit casual/colloquial sounding, but that fits with your general writing style and I like it. In fact, I feel so strongly about it that I'm making the change.

W/ respect to "at the moment", since I believe you are referring to something that has become true relatively recently but that will remain true for the foreseeable future, "at the moment" wouldn't work too well -- it would imply that it might stop being true at any time.

joaquinelio:
I thought "As of now" means "-From- now  -to- the future",
"As of now I WILL stop smoking"   

I could, but I won't say
"Level 2 is supported in all major browsers now."

About "at the moment" I hoped a good transl of "Al momento",
What you say is exactly what I wanted to say.
It changed recently, it is that way now, it may change in the future  (It doesnt mean deprecated)
19 "... Level 2 is supported in all major browsers, while the Level 3 is in the works..."

Spanish "Al momento" stays.
chan-chán  : )

joaquinelio:
2/3 of repo translated.  
If I collapse, please consider @ maksumi , he is good reviewing and helping new translators.
Should be a limit in the number of  maintainers? You can add him.

THANKS to this site, I wrote a program that  
-fetch issue "spanish transl progress" 
-await for responses  body and comments
-regexp json 
then (rough) code lists claimed and ABANDONED articles sorted by date  (It was a good help!)
and I loved doing it.  
Thanks, @iliakan

iliakan:
@paroche So it's ok to leave the text "as is", without additional changes?

@joaquinelio Added @maksumi. 

We can publish the Spanish version then? If 2/3 is translated, that's enough. 
Just please make sure that each translation gets a review from a non-novice JS developer, to ensure its quality =)

paroche:
Yes, the part under discussion makes perfect sense to me the way it is, though as I was double checking I noticed some little changes I wanted to make in the latter part of the sentence and the following sentences, and, as it turned out, the rest of the article, which I have now made.

iliakan:
@paroche Thanks, as always! =)

paroche:
Your're welcome. That was definitely the idea. :-). 

I think there are more things that could be changed, but I just did what was quick and obvious. For the rest it would have taken more thought (and a better understanding of the material -- this was the first time I had looked at that article).

joaquinelio:
"as of now"
I believe @peroche.  The problem is ours, a too rigid translation.  

> We can publish the Spanish version then? If 2/3 is translated, that's enough.

114 of 174 articles right now. 
I'll check if they are not too sparse... First section looks pretty good.  
But I would need to start another eviction process to finish it : )    

> Just please make sure that each translation gets a review from a non-novice JS developer, to ensure its quality =)
Hard to get them, hard to keep them.  
I'm an old, experienced programmer... old techniques, and trusty sql design. 
I never did JS/web  in production 
I feel I'm not qualified to check high level code. Still, we all do our best to understand it first.

bytheway, 
What did you do with "proxy & reflect", nobody wants it.  It looks hard. 
There is another long article, indexedDB, not a problem.  I can do it, my very first learning project was plain js/callback based indexedDB working nicely.  I dont see it as a priority though.

iliakan:
These two articles are not top priority. Proxy is something you need once a year (good to know though).

IndexedDB is also a very specific thing, mostly for offline apps.

joaquinelio:
Thanks Ilya, we guessed that.
maksumi keeps "Proxy & reflect" just to prevent discouraging in new translators. It is intimidating. 
I only used indexedDB because of the async thing, Chrome & Firefox (what a difference! I hate the freedom of JS)  Is it safe from the "cleaners" apps?  

Oh, Hi @paroche, I put last comment before noticing you were here.
It's good to learn from your changes.  I know I can trust you. Well, native speaker advantage...

iliakan:
@joachimklug Most browser allow to clear IndexedDB. I can't be sure about "cleaner apps" out there

paroche:
@joachimklug. Yes, native speaking definite advantage. And love of language and clarity and stuff.

joaquinelio:
I did replicated db...   
2 sites, replicated SQL, sync by phone once a day via 2 US Robotics modems...
That was... last century...
Today I could use indexedDB, but the cache woudnt need to stay long with no sync anyway. 
Times changing...

About JS+indexedDB, Chrome is very forgiving... I tried the working code on Firefox, its not just different implementation, it forced me to write better code.

PR #2048, opened 2020-08-08
Fix wrong expression

OMEGA-Y:


iliakan:
I really mean "vertical" there.

OMEGA-Y:
I thought this sentence means to separate the code and write it in multiple lines because it was not clear when read horizontally, so something could be missed.
I wonder why you mean "vertical" there.

iliakan:
We usually eye-scan the code vertically when we read it.

Issue #2039, opened 2020-08-02
Mixins - EventMixin - Example description is misleading


arcticmatt:
The description of the example code says the following:

> After adding the mixin, an object user will be able to generate an event "login" when the visitor logs in. And another object, say, calendar may want to listen for such events to load the calendar for the logged-in person.
> 
> Or, a menu can generate the event "select" when a menu item is selected, and other objects may assign handlers to react on that event. And so on.

However, based on my understanding of the example code, these type of cross-object use cases are not supported.

For example, if you run the example code, `_eventHandlers` is stored on the objected referenced by `menu`. So if you created another `Menu` object, the two will not be able to interact via events.

iliakan:
Can you please show me the problem that you see? Maybe a piece of code?
Everything should be ok.

arcticmatt:
```
let eMixin = {
  /**
   * Subscribe to event, usage:
   *  menu.on('select', function(item) { ... }
  */
  on(eventName, handler) {
    if (!this._eventHandlers) this._eventHandlers = {};
    if (!this._eventHandlers[eventName]) {
      this._eventHandlers[eventName] = [];
    }
    this._eventHandlers[eventName].push(handler);
  },

  /**
   * Cancel the subscription, usage:
   *  menu.off('select', handler)
   */
  off(eventName, handler) {
    let handlers = this._eventHandlers?.[eventName];
    if (!handlers) return;
    for (let i = 0; i < handlers.length; i++) {
      if (handlers[i] === handler) {
        handlers.splice(i--, 1);
      }
    }
  },

  /**
   * Generate an event with the given name and data
   *  this.trigger('select', data1, data2);
   */
  trigger(eventName, ...args) {
    if (!this._eventHandlers?.[eventName]) {
      return; // no handlers for that event name
    }

    // call the handlers
    this._eventHandlers[eventName].forEach(handler => handler.apply(this, args));
  }
};

// Make a class
class Menu {
  choose(value) {
    this.trigger("select", value);
  }
}
// Add the mixin with event-related methods
Object.assign(Menu.prototype, eMixin);

let menu = new Menu();

// add a handler, to be called on selection:
menu.on("select", value => alert(`Value selected: ${value}`));

// triggers the event => the handler above runs and shows:
// Value selected: 123
menu.choose("123");

let menu2 = new Menu();
// This does nothing
menu2.choose("123");
```

@iliakan the last two lines are what I'm talking about. The lines above are taken from the article.

[...]

PR #2052, opened 2020-08-09
9.7 - Clarify the description for the mixins example

arcticmatt:
The description of the example code says the following:

> After adding the mixin, an object user will be able to generate an event "login" when the visitor logs in. And another object, say, calendar may want to listen for such events to load the calendar for the logged-in person.
> 
> Or, a menu can generate the event "select" when a menu item is selected, and other objects may assign handlers to react on that event. And so on.

However, based on my understanding of the example code, these type of cross-object use cases are not supported.

For example, if you run the example code, `_eventHandlers` is stored on the objected referenced by `menu`. So if you created another `Menu` object, the two will not be able to interact via events.

See https://github.com/javascript-tutorial/en.javascript.info/issues/2039 for more details.


iliakan:
I tried hard to understand the reasoning behind this change.

As we're talking about the `user` object in the example above, it makes sense to talk about `user`, and maybe `menu` later, not vise versa.

PR #2055, opened 2020-08-10
1-9-1 Class basic syntax, "class fields" content

joaquinelio:
I'm very comfortable with the term "field" because of my past in SQL 
but I wonder if it may be confusing for first timers (still in "js basics") 
It can be important to remark :  it's still a property, and the why of "The important difference...".

then maybe expand the example with
332 let anotherUser = new User(); 
etc

Do we need 
298 Previously, our classes only had methods.? I wouldn't care if I were learning. They work or they dont. 
Just asking.

I'm more confident doing requesting changes 
because they can rejected BUT it may lead to a better ideas. = )

### DIFF
--- a/1-js/09-classes/01-class/article.md
+++ b/1-js/09-classes/01-class/article.md
@@ -317,7 +317,7 @@ new User().sayHi(); // Hello, John!
 
 So, we just write "<property name> = <value>" in the declaration, and that's it.
 
-The important difference of class fields is that they are set on individual objects, not `User.prototype`:
+The important difference of properties defined this way is that they are set on the individual objects, not shared in User.prototype:
 
 ```js run
 class User {


iliakan:
"class fields" is the official term for properties defined this way.

iliakan:
See above ""Class fields" is a syntax that allows to add any properties."

joaquinelio:
>"class fields" is the official term for properties defined this way.

ok, I didnt mean to remove the definition spec, it stays.

>See above ""Class fields" is a syntax that allows to add any properties."

It was confusing (when learning), I already got that possibility . 
In fact I wanted to put the mod there but it would force to make more changes... not good ones.

I just thought I would want to know the  *why*  sooner, adding 
 "individual, not *shared* in the prototype" 
It's ok.

PR #2058, opened 2020-08-11
5.4 - change 7 basic types to eight basic types

arcticmatt:
The text in https://javascript.info/array is inconsistent with the text in https://javascript.info/types. The latter is correct.


iliakan:
Thank you!

PR #2059, opened 2020-08-12
fix undeclared variables in task

acordado:
replace variables to number to fix undeclared variables


iliakan:
You got the task wrong, sorry.

PR #2060, opened 2020-08-12
fix undeclared variables in solution

acordado:
replace variables to number to fix undeclared variables.


iliakan:
the solution isn't runnable, so that's ok

PR #2072, opened 2020-08-21
Remove Edge and Firefox warning

josephrocca:
Support in Edge 79 and Firefox 78: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Browser_compatibility


iliakan:
=)

PR #2073, opened 2020-08-22
once in -> once every

joaquinelio:
I think:
just once, once in x secs
periodically, once every x secs.

Even less confident, line 73,75
The way I rewrote on Spanish review.

Not my language, I put this here just because peroche wasnt here to "blame"


iliakan:
Looks good, Thank you!

PR #2078, opened 2020-08-24
Removed the parts about browser compatibility

moisheschwartz:
`globalThis` is now supported in all major browsers, see link https://caniuse.com/#search=globalThis


iliakan:
Thank you!

PR #2084, opened 2020-08-26
focus/blur events dont make focus but track it

joaquinelio:
Im not a good English speaker, 
rewrite my suggest in good English plis...


iliakan:
Thanks, updated that section a bit.

PR #2109, opened 2020-09-09
Update article.md

Jarvis-3000:
I have modified 
**width:200**     to      **width=200**
at line **72** (params)


iliakan:
Right, thanks =)

PR #2110, opened 2020-09-09
Update article.md

cristian-wang:
Hi guys, I looked at your article to solve my CORS issue but it didn't work. So eventually I found a question on StackOverflow that explained in order to make things work properly, I needed to add also "Access-Control-Allow-Origin: https://javascript.info" to the preflight response. Did anyone have the same issue? If so, we need to update the article in order to save the time of future readers :)

StackOverflow question: https://stackoverflow.com/questions/42558221/how-to-cors-enable-apache-web-server-including-preflight-and-custom-headers

### DIFF
--- a/5-network/05-fetch-crossorigin/article.md
+++ b/5-network/05-fetch-crossorigin/article.md
@@ -259,6 +259,7 @@ Access-Control-Request-Headers: Content-Type,API-Key
 ### Step 2 (preflight response)
 
 The server should respond with status 200 and headers:
+- `Access-Control-Allow-Origin: https://javascript.info`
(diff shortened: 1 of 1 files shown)


MuhammedZakir:
> add also "Access-Control-Allow-Origin: https://javascript.info" to the preflight response.

AFAIK, we have to add `Access-Control-Allow-Origin` header in both preflight and actual responses.

> If so, we need to update the article in order to save the time of future readers :)

If we are going to do it, we will have to add it in other places too, including image, code snippet etc.

iliakan:
Thank you, I'll add the additional changes.

PR #2124, opened 2020-09-17
Update article.md

kskashyap94:
The default behaviour of dynamic scripts can be changed by setting script.async=false, instead it is mentioned as script.async=true


iliakan:
No, the article is correct.

PR #2125, opened 2020-09-17
Add Edge to list of V8-powered browsers

shanesc:


iliakan:
Are you sure Edge uses V8 by default? On my Windows machine, it looks like Edge is still not V8.

shanesc:
I believe so. The Microsoft support page states that as of Jan 15, 2020, Edge is based on Chromium. (https://support.microsoft.com/en-gb/help/4501095/download-the-new-microsoft-edge-based-on-chromium)

And automatic rollout began in June, so I think going forward it's safe to assume it's running V8. The old version is now referred to as Edge Legacy.

By the way, thank you for all the work you've put into this site. It is so well written and your knack for explaining complex topics in a digestible way is one of the best I've found. Very much appreciated!

iliakan:
Thank you! How can I check that my Edge uses V8?

I mean, I thought I know a few ways, but they show I'm not on V8, so maybe I'm missing something.

MuhammedZakir:
> Thank you! How can I check that my Edge uses V8?
> 
> I mean, I thought I know a few ways, but they show I'm not on V8, so maybe I'm missing something.
> 
> Or maybe the update of Edge to V8 version doesn't happen automatically?

https://en.wikipedia.org/wiki/Microsoft_Edge:
> The first public release followed on January 15, 2020 [16]. In June 2020, Microsoft began automatic rollout of Chromium-based Edge via Windows Update for Windows 7, 8.1, and all Windows 10 versions from 1803 to 2004. [17]
> 
> [17] https://en.wikipedia.org/wiki/Microsoft_Edge#cite_note-17

Did you disable automatic updates?

FYI, I think Edge can't be uninstalled (in a normal way?) -- https://support.microsoft.com/en-gb/help/4576988/why-can-t-i-uninstall-microsoft-edge.

iliakan:
I'll check my Windows VM for updates. Merging this now, thanks.

PR #2126, opened 2020-09-17
Typos

peachesontour:


6-data-storage/03-indexeddb/article.md
```diff
@@ -40,15 +40,15 @@ The call returns `openRequest` object, we should listen to events on it:
 
 **IndexedDB has a built-in mechanism of "schema versioning", absent in server-side databases.**
 
-Unlike server-side databases, IndexedDB is client-side, the data is stored in the browser, so we, developers, don't have "any time" access to it. So, when we published a new version of our app, and the user visits our webpage, we may need to update the database.
+Unlike server-side databases, IndexedDB is client-side, the data is stored in the browser, so we, developers, don't have full-time access to it. So, when we have published a new version of our app, and the user visits our webpage, we may need to update the database.
```
MuhammedZakir, line 43:
```suggestion
Unlike server-side databases, IndexedDB is client-side - the data is stored in the browser. So we, developers, don't have full-time access to it. Due to that, we can only change/update the database when the user visits our webpage.
```

6-data-storage/03-indexeddb/article.md
```diff
@@ -121,13 +121,13 @@ So there's a tab with an open connection to DB version `1`, while the second tab
 
 The problem is that a database is shared between two tabs, as it's the same site, same origin. And it can't be both version `1` and `2`. To perform the update to version `2`, all connections to version 1 must be closed, including the one in the first tab.
 
-In order to organize that, the `versionchange` event triggers in such case on the "outdated" database object. We should listen to it and close the old database connection (and probably suggest the visitor to reload the page, to load the updated code).
+In order to organize that, the `versionchange` event triggers on the "outdated" database object. We should listen for it and close the old database connection (and probably suggest the visitor reloads the page, to load the updated code).
```
MuhammedZakir, line 124:
```suggestion
In order to organize that, the `versionchange` event triggers on the "outdated" database object. We should listen for it and close the old database connection (and probably suggest the visitor to reload the page, to load the updated code).
```

6-data-storage/03-indexeddb/article.md
```diff
@@ -103,9 +103,9 @@ let deleteRequest = indexedDB.deleteDatabase(name)
 ```warn header="We can't open an older version of the database"
 If the current user database has a higher version than in the `open` call, e.g. the existing DB version is `3`, and we try to `open(...2)`, then that's an error, `openRequest.onerror` triggers.
 
-That's odd, but such thing may happen when a visitor loaded an outdated JavaScript code, e.g. from a proxy cache. So the code is old, but his database is new.
+That's rare, but such a thing may happen when a visitor loads outdated JavaScript code, e.g. from a proxy cache. So the code is old, but his database is new.
 
-To protect from errors, we should check `db.version` and suggest him to reload the page. Use proper HTTP caching headers to avoid loading the old code, so that you'll never have such problem.
+To protect from errors, we should check `db.version` and suggest he reloads the page. Use proper HTTP caching headers to avoid loading the old code, so that you'll never have such problems.
```
MuhammedZakir, line 108:
```suggestion
To protect from errors, we should check `db.version` and suggest a page reload. Use proper HTTP caching headers to avoid loading the old code, so that you'll never have such problems.
```

MuhammedZakir: Actually, I am not quite sure if your usage is grammatically correct. Something just doesn't feel right. I am not sure, so you could be right.

6-data-storage/03-indexeddb/article.md
MuhammedZakir: OR

```suggestion
In order to organize that, the `versionchange` event triggers on the "outdated" database object. We should listen for it and close the old database connection (and probably suggest a page reload, to load the updated code).
```

6-data-storage/03-indexeddb/article.md
```diff
@@ -179,16 +179,16 @@ Despite being named an "object store", primitives can be stored too.
 
 IndexedDB uses the [standard serialization algorithm](https://www.w3.org/TR/html53/infrastructure.html#section-structuredserializeforstorage) to clone-and-store an object. It's like `JSON.stringify`, but more powerful, capable of storing much more datatypes.
 
-An example of object that can't be stored: an object with circular references. Such objects are not serializable. `JSON.stringify` also fails for such objects.
+An example of an object that can't be stored: an object with circular references. Such objects are not serializable. `JSON.stringify` also fails for such objects.
 
 **There must be a unique `key` for every value in the store.**     
 
-A key must have a type one of: number, date, string, binary, or array. It's an unique identifier: we can search/remove/update values by the key.
+A key must be one of the these types: number, date, string, binary, or array. It's a unique identifier: we can search/remove/update values by the key.
```
MuhammedZakir, line 186:
```suggestion
A key must be one of the these types - number, date, string, binary, or array. It's a unique identifier, so we can search/remove/update values by the key.
```

6-data-storage/03-indexeddb/article.md
```diff
@@ -343,7 +343,7 @@ So, in the example above no special call is needed to finish the transaction.
 
 Transactions auto-commit principle has an important side effect. We can't insert an async operation like `fetch`, `setTimeout` in the middle of transaction. IndexedDB will not keep the transaction waiting till these are done.
 
-In the code below `request2` in line `(*)` fails, because the transaction is already committed, can't make any request in it:
+In the code below `request2` in line `(*)` fails, because the transaction is already committed, and can't make any request in it:
```
MuhammedZakir, line 346:
```suggestion
In the code below, `request2` in line `(*)` fails, because the transaction is already committed, and can't make any request in it:
```

6-data-storage/03-indexeddb/article.md
peachesontour: Yes, you are right, it's a subjunctive. Glad you caught it.

iliakan:
Thank you very much for such thorough and beneficial edits!

PR #2081, opened 2020-08-25
Restructure the Solution for 'Army of Functions' task and Fix Typos

MuhammedZakir:
Fix #2068 - Army of Functions
Fix #2070 - Typo
Fix #2056 - Grammatical Error
Fix #2074 - Remove semi-colon after function declaration

PR #2127, opened 2020-09-18
Improve make-army task

MuhammedZakir:
Fix grammar and indent some paragraphs.

Complements javascript-tutorial/en.javascript.info#2081.

### DIFF
--- a/1-js/06-advanced-functions/03-closure/10-make-army/solution.md
+++ b/1-js/06-advanced-functions/03-closure/10-make-army/solution.md
@@ -6,7 +6,7 @@ Let's examine what exactly happens inside `makeArmy`, and the solution will beco
     ```js
     let shooters = [];
     ```
-2. Fills it in the loop via `shooters.push(function...)`.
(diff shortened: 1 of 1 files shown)


iliakan:
Had to make a few amendments in commit e61bfa26

MuhammedZakir:
> Had to make a few amendments in commit [e61bfa2](https://github.com/javascript-tutorial/en.javascript.info/commit/e61bfa2631d34b32044121767ae6ee13223cb983)

I will keep that in mind. :-)

About that last one, 

```
Now, as you've put so much effort into reading this, and the final recipe is so simple - just use `for`, you may wonder -\- was it worth that?
```

Without backslash, HTML output of this paragraph was strange. That's why, I added it. I forgot what the problem was though.

iliakan:
`--` gets autoconverted to – (long dash) by the rendering engine.

MuhammedZakir:
> `--` gets autoconverted to – (long dash) by the rendering engine.

I see. So does it show correctly now? Also, how long does the site take to show updated content?

PR #2128, opened 2020-09-18
Update article.md

maslaff:
Based on the fact that the precision is high and low, here it should be maximum. Or need to talk about the value - "the minimum possible value".
Check grammar, please.


iliakan:
Changed.

PR #2131, opened 2020-09-19
Address some possible issues in 1.2.10

vsemozhetbyt:
1. Renumber outdated task order.
2. Make variable name more readable.


iliakan:
Please don't renumber without a real need. All translators have to merge this change, but it does not affect the site at all.

PR #2138, opened 2020-09-20
Address some possible issues in 1.2.12

vsemozhetbyt:


iliakan:
The original article was heavily modified 💥 💣

PR #2140, opened 2020-09-21
Typos

peachesontour:


7-animation/2-css-animations/article.md
```diff
@@ -1,14 +1,14 @@
...
 
-JavaScript can be used to control CSS animation and make it even better with a little of code.
+JavaScript can be used to control CSS animations and make them even better, with little code.
 
 ## CSS transitions [#css-transition]
 
 The idea of CSS transitions is simple. We describe a property and how its changes should be animated. When the property changes, the browser paints the animation.
 
-That is: all we need is to change the property. And the fluent transition is made by the browser.
+That is: all we need is to change the property. And the fluid transition is done by the browser.
```
MuhammedZakir, line 11:
```suggestion
That is, all we need is to change the property; and, the fluid transition will be done by the browser.
```

7-animation/2-css-animations/article.md
```diff
@@ -70,25 +70,25 @@ growing.onclick = function() {
 </script>
 ```
 
-Now let's cover animation properties one by one.
+Now, let's cover animation properties one by one.
 
 ## transition-property
 
-In `transition-property` we write a list of property to animate, for instance: `left`, `margin-left`, `height`, `color`.
+In `transition-property` we write a list of properties to animate, for instance: `left`, `margin-left`, `height`, `color`.
```
MuhammedZakir, line 77:
```suggestion
In `transition-property`, we write a list of properties to animate, for instance: `left`, `margin-left`, `height`, `color`.
```

7-animation/2-css-animations/article.md
```diff
@@ -70,25 +70,25 @@ growing.onclick = function() {
...
 
-Not all properties can be animated, but [many of them](http://www.w3.org/TR/css3-transitions/#animatable-properties-). The value `all` means "animate all properties".
+Not all properties can be animated, but [many of them can](http://www.w3.org/TR/css3-transitions/#animatable-properties-). The value `all` means "animate all properties".
```
MuhammedZakir, line 79: Id in the URL, aka URI fragment, does not point anywhere in the document. I think it's better to replace W3C link with MDN. MDN page has a list of animatable properties.

```suggestion
Although not all properties can be animated, [almost all of the properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties)  generally used can be animated. The value `all` means "animate all properties".
```

"almost all of the properties generally used"  OR
"almost all of the generally used properties"  OR
"almost all of the generally used ones" ?

7-animation/2-css-animations/article.md
```diff
@@ -70,25 +70,25 @@ growing.onclick = function() {
...
 
 ## transition-duration
 
 In `transition-duration` we can specify how long the animation should take. The time should be in [CSS time format](http://www.w3.org/TR/css3-values/#time): in seconds `s` or milliseconds `ms`.
 
 ## transition-delay
 
-In `transition-delay` we can specify the delay *before* the animation. For instance, if  `transition-delay: 1s`, then animation starts after 1 second after the change.
+In `transition-delay` we can specify the delay *before* the animation. For instance, if  `transition-delay: 1s`, then the animation starts 1 second after the property change.
```
MuhammedZakir, line 87:
```suggestion
In `transition-delay` we can specify the delay *before* the animation. For instance, `transition-delay: 1s`, starts the animation 1 second after the property change.
```

7-animation/2-css-animations/article.md
```diff
@@ -70,25 +70,25 @@ growing.onclick = function() {
...
 
-Negative values are also possible. Then the animation starts from the middle. For instance, if `transition-duration` is `2s`, and the delay is `-1s`, then the animation takes 1 second and starts from the half.
+Negative values are also possible. Then the animation starts from the middle. For instance, if `transition-duration` is `2s`, and the delay is `-1s`, then the animation takes 1 second and starts from the halfway point.
```
MuhammedZakir, line 89: Negative values are also possible. Then the animation is shown immediately, but the starting point of the animation will be after given value (time). For example, if `transition-delay` is `-1s` and `transition-duration` is `2s`, then animation starts from the halfway point and total duration will be 1 second. You can think it like this:

    `transition-delay + transition-duration = animation duration`

So, if for instance, `transition-delay` is `-2s` and `transition-duration` is `2s`, then it will be as if there was no transitioning animation for that property and the changes are shown instantly. Therefore, "animation duration" must always be greater than `0s` for any animation to be shown.
```

Comment (MuhammedZakir) [reply 7-animation/2-css-animations/article.md:89]: Some more polishing is probably needed. Please wait for more suggestions before adding this.

File: 7-animation/2-css-animations/article.md [hunk 6]
```diff
@@ -129,25 +129,25 @@ stripe.onclick = function() {
 
 ## transition-timing-function
 
-Timing function describes how the animation process is distributed along the time. Will it start slowly and then go fast or vise versa.
+The timing function describes how the animation process is distributed along its timeline. Will it start slowly and then go fast or vice versa.
```
MuhammedZakir, line 132:
```suggestion
The timing function describes how the animation process is distributed along its timeline. Will it start slowly and then go fast, or vice versa.
```

7-animation/2-css-animations/article.md
```diff
@@ -129,25 +129,25 @@ stripe.onclick = function() {
...
 
 ### Bezier curve
 
-The timing function can be set as a [Bezier curve](/bezier-curve) with 4 control points that satisfies the conditions:
+The timing function can be set as a [Bezier curve](/bezier-curve) with 4 control points that satisfy the conditions:
 
 1. First control point: `(0,0)`.
 2. Last control point: `(1,1)`.
-3. For intermediate points values of `x` must be in the interval `0..1`, `y` can be anything.
+3. For intermediate points the values of `x` must be in the interval `0..1`, `y` can be anything.
```
MuhammedZakir, line 144:
```suggestion
3. For intermediate points, the values of `x` must be in the interval `0..1`, `y` can be anything.
```

7-animation/2-css-animations/article.md
```diff
@@ -244,21 +244,21 @@ But if you click the train, you'll see that:
...
-Why it happens -- pretty obvious if we look at the graph of the given Bezier curve:
+Why it happens is pretty obvious if we look at the graph of the given Bezier curve:
 
 ![](bezier-train-over.svg)
 
-We moved the `y` coordinate of the 2nd point below zero, and for the 3rd point we made put it over `1`, so the curve goes out of the "regular" quadrant. The `y` is out of the "standard" range `0..1`.
+We moved the `y` coordinate of the 2nd point below zero, and for the 3rd point we made it over `1`, so the curve goes out of the "regular" quadrant. The `y` is out of the "standard" range `0..1`.
 
-As we know, `y` measures "the completion of the animation process". The value `y = 0` corresponds to the starting property value and `y = 1` -- the ending value. So values `y<0` move the property lower than the starting `left` and `y>1` -- over the final `left`.
+As we know, `y` measures "the completion of the animation process". The value `y = 0` corresponds to the starting property value and `y = 1` -- the ending value. So values `y<0` move the property beyond than the starting `left` and `y>1` -- past the final `left`.
```
MuhammedZakir, line 253:
```suggestion
As we know, `y` measures "the completion of the animation process". The value `y = 0` corresponds to the starting property value and `y = 1` -- the ending value. So values `y<0` move the property beyond the starting `left` and `y>1` -- past the final `left`.
```

7-animation/2-css-animations/article.md
```diff
@@ -369,7 +369,7 @@ The event object for `transitionend` has few specific properties:
 
 We can join multiple simple animations together using the `@keyframes` CSS rule.
 
-It specifies the "name" of the animation and rules: what, when and where to animate. Then using the `animation` property we attach the animation to the element and specify additional parameters for it.
+It specifies the "name" of the animation and rules: what, when and where to animate. Then using the `animation` property we can attach the animation to the element and specify additional parameters for it.
```
MuhammedZakir, line 372:
```suggestion
It specifies the "name" of the animation and rules - what, when and where to animate. Then using the `animation` property we can attach the animation to the element and specify additional parameters for it.
```

7-animation/2-css-animations/article.md
peachesontour: Looks good.

7-animation/2-css-animations/article.md
peachesontour: Thanks. To be consistent maybe we should format the 2s as `2s` too.
...and `transition-duration` is `2s`, then...

7-animation/2-css-animations/article.md
```diff
@@ -70,25 +70,25 @@ growing.onclick = function() {
...
+Do note that, there are properties which can not be animated. However, [most of the generally used properties are animatable](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_animated_properties).
 
 ## transition-duration
 
 In `transition-duration` we can specify how long the animation should take. The time should be in [CSS time format](http://www.w3.org/TR/css3-values/#time): in seconds `s` or milliseconds `ms`.
 
 ## transition-delay
 
-In `transition-delay` we can specify the delay *before* the animation. For instance, if  `transition-delay: 1s`, then animation starts after 1 second after the change.
+In `transition-delay` we can specify the delay *before* the animation. For instance, if `transition-delay` is `1s` and `transition-duration` is 2s, then the animation starts 1 second after the property change and the total duration will be 2 seconds.
```
MuhammedZakir, line 87:
```suggestion
In `transition-delay` we can specify the delay *before* the animation. For instance, if `transition-delay` is `1s` and `transition-duration` is `2s`, then the animation starts 1 second after the property change and the total duration will be 2 seconds.
```

7-animation/2-css-animations/article.md
MuhammedZakir: You're welcome! 👍 

> Thanks. To be consistent maybe we should format the 2s as `2s` too.
> ...and `transition-duration` is `2s`, then...

I missed that. I have added a new comment with this change.

7-animation/2-css-animations/article.md
```diff
@@ -1,14 +1,14 @@
...
 
-JavaScript can be used to control CSS animation and make it even better with a little of code.
+JavaScript can be used to control CSS animations and make them even better, with little code.
 
 ## CSS transitions [#css-transition]
 
 The idea of CSS transitions is simple. We describe a property and how its changes should be animated. When the property changes, the browser paints the animation.
 
-That is: all we need is to change the property. And the fluent transition is made by the browser.
+That is, all we need is to change the property; and, the fluid transition will be done by the browser.
```
iliakan, line 11: Is the semicolon needed? `;`

iliakan: The syntax seems just wrong around there.

iliakan (changes requested):
Please see @MuhammedZakir 's comments as well.

7-animation/2-css-animations/article.md
peachesontour: You're right, it isn't needed.

MuhammedZakir: > The syntax seems just wrong around there.

It indeed looks a bit awkward.

```suggestion
That is, all we need is to change the property, and the fluid transition will be done by the browser.
```

7-animation/2-css-animations/article.md
MuhammedZakir: @iliakan : Is this okay as it is? Any input?

iliakan:
Thank you!

PR #2141, opened 2020-09-22
typo

haond10adp:


2-ui/4-forms-controls/1-form-elements/article.md
```diff
@@ -205,7 +205,7 @@ Here is an example:
 </script>
 ```
 
-Unlike most other controls, `<select>` allows to select multiple options at once if it has `multiple` attribute. That's feature is rarely used. In that case we need to use the first way: add/remove the `selected` property from `<option>` subelements.
+Unlike most other controls, `<select>` allows to select multiple options at once if it has `multiple` attribute. That feature is rarely used. In that case we need to use the first way: add/remove the `selected` property from `<option>` subelements.
```
MuhammedZakir, line 208:
```suggestion
Unlike most other controls, `<select>` allows to select multiple options at once if it has the attribute, `multiple`. In such cases, we need to use the first way - add/remove the `selected` property from `<option>` subelements.
```

1. Changed the first sentence to avoid interpreting "*multiple* attribute" as "*multiple attributes*".
2. Removed "feature is rarely used" statement, as it disrupts the flow and is unnecessary(?).

Suggestions are welcome! :-)

iliakan:
Is it really better? Any native English speakers please? ;)

dorelljames:
@iliakan in my opinion, it doesn't seem to make it better. I've made pull request #2190 to supplement.

PR #2142, opened 2020-09-22
One more step please

joaquinelio:
I... can't... 
If I see step1, I expect the 2
Forgive me, Im a maniac.
maybe
Start creating a controller
Lets create the controller....


iliakan:
Probably a leftover from the previous version :/

PR #2145, opened 2020-09-24
Typo Fixed

Ramjonchhen:
Fixed the typo of the Chapter JavaScript Fundamentals
Topic: Functions
Lesson Navigation: Default Values.
The second paragraph of the topic contained a sentence containing typo:
For instance, the aforementioned function showMessage(from, text) can be called with a single argument:
the typo is afroementioned which should have been as: above mentioned functions which is trying to point a function
defined above showMessage(from,text) function in the chapter function.

### DIFF
--- a/1-js/02-first-steps/15-function-basics/article.md
+++ b/1-js/02-first-steps/15-function-basics/article.md
@@ -179,7 +179,7 @@ alert( from ); // Ann
 
 If a parameter is not provided, then its value becomes `undefined`.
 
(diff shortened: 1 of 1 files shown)


MuhammedZakir:
"aforementioned" = "denoting a thing or person previously mentioned". So that's not a typo.

iliakan:
Being too formal there, eh?

Ramjonchhen:
thanks for the comment on the pull request. I should have checked the meaning of the word. Hope to contribute more to this repository in the future.

PR #2147, opened 2020-09-24
Changed "index" to "position" for splice() method

Ruslan-Aleev:
RU - https://github.com/javascript-tutorial/ru.javascript.info/pull/1111

### DIFF
--- a/1-js/05-data-types/05-array-methods/article.md
+++ b/1-js/05-data-types/05-array-methods/article.md
@@ -41,10 +41,10 @@ The [arr.splice(start)](mdn:js/Array/splice) method is a swiss army knife for ar
 The syntax is:
 
 ```js
-arr.splice(index[, deleteCount, elem1, ..., elemN])
+arr.splice(pos[, deleteCount, elem1, ..., elemN])
 ```
 
-It starts from the position `index`: removes `deleteCount` elements and then inserts `elem1, ..., elemN` at their place. Returns the array of removed elements.
+It starts from the position `pos`: removes `deleteCount` elements and then inserts `elem1, ..., elemN` at their place. Returns the array of removed elements.
 
 This method is easy to grasp by examples.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Why?

Ruslan-Aleev:
I didn't quite understand the question.
1) There are 2 spellings in the text, which is confusing.
See https://github.com/javascript-tutorial/en.javascript.info/pull/2147/files#diff-be29ec7c19bb09855fdabe5c4b5c2a1cL44 and https://github.com/javascript-tutorial/en.jpullascript.info2147/files#diff-be29ec7c19bb09855fdabe5c4b5c2a1cL714
2) Position is not an index position. For example in https://github.com/javascript-tutorial/en.javascript.info/pull/2147/files#diff-be29ec7c19bb09855fdabe5c4b5c2a1cL92 `arr[2] == "JavaScript"`.

Ruslan-Aleev:
By the way, this line also says "index", but you need to write "position" - https://github.com/javascript-tutorial/en.javascript.info/pull/2147/files#diff-be29ec7c19bb09855fdabe5c4b5c2a1cL92

Ruslan-Aleev:
Or I'm confusing something? :)

Ruslan-Aleev:
As I understand it s**P**lice uses position, and slice uses index. This should be indicated in the course text as well.

iliakan:
I suppose index and position are synonyms.

Ruslan-Aleev:
Yes, I already figured it out, thanks.

I was confused by the comment on line 92.
Instead of `// from index 2` worth to specify - `// 2 - index at which to start changing the array`.
In russian it is also not entirely clear:
`// с позиции 2`, but it is correct to write `// на позиции 2` or `// вместо позиции 2`.

Although, perhaps this only confuses me :) Anyway, thanks for the awesome site!

PR #2152, opened 2020-09-26
typo

haond10adp:


dorelljames:
The original version is grammatically correct

PR #2156, opened 2020-09-27
Fix code in 1.2.15

vsemozhetbyt:
The current code excludes `n` from the results.
With the fix, its behavior is the same as in the task from "Loops: while and for".


iliakan:
"Up to n" doesn't always mean "including n"

iliakan:
It's vague in this aspect, but that's an example, not a task, hope it's ok.

PR #2160, opened 2020-10-01
Improve Array method ( indexOf) document

ctoaugurs:


iliakan:
Too many typos, I don't get it.

PR #2161, opened 2020-10-01
change and correct if-else block  ( Update solution.md)

Hasnain01-hub:
(string_name) is just a variable name..
It checks the condition and if it is true alert box will pop up with a hello msg...
Add this solution.md :-)


iliakan:
I don't get it =( Where does `string_name" come from?

iliakan:
Me too.

iliakan:
That's gonna ruin the code, no such variable (`string`)

PR #2165, opened 2020-10-04
Update article.md

rutvik110:
The changes I've made are relative to how we calculate the shifts for ball.The one way i proposed is quite faster and gives the same result as we get from the current calculations for finding shifts and I've also mentioned when you should use this way to calculate shifts.


iliakan:
Is it supported across the major modern browsers?

rutvik110:
yes,i checked it and it's supprted across all browsers.

iliakan:
Maybe we could add it to the events section... That said, I'd rather leave it as is, because the ball element may have subelements, that are event targets, while we need the offset to the ball.

rutvik110:
Ya sure,np👍

PR #2168, opened 2020-10-05
Fix syntax enumeration in 1.4.7

vsemozhetbyt:
Currently, syntax of the 1 and 3 cases is the same.


lex111 (approved):
Good catch!

lex111:
@vsemozhetbyt thanks!

PR #2171, opened 2020-10-07
Improve some English grammar problems

tunggnu:
I did it with the help of the Grammarly tool.


iliakan:
"For example" doesn't require a comma.

Maybe @paroche could say better here.

paroche:
I agree that it's better without the comma. However, it may be a bit better with a semicolon _before_ "for example", a change I am about to make.

paroche:
Don't want this overriding what I just did. I think the semicolon is only somewhat better. I considered an m-dash, but went w/ the semi. Preceding comma wasn't bad, though. Following comma would be less good.

iliakan:
Thank you!

PR #2190, opened 2020-10-12
Update 2-ui/4-forms-controls/1-form-elements/article.md

dorelljames:
This updates a small grammar error and adds some helpful info. Hoping it's better now. Let me know if you need any more changes and I'll update it immediately.

### DIFF
--- a/2-ui/4-forms-controls/1-form-elements/article.md
+++ b/2-ui/4-forms-controls/1-form-elements/article.md
@@ -205,7 +205,9 @@ Here is an example:
 </script>
 ```
 
-Unlike most other controls, `<select>` allows to select multiple options at once if it has `multiple` attribute. That's feature is rarely used. In that case we need to use the first way: add/remove the `selected` property from `<option>` subelements.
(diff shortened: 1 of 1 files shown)


2-ui/4-forms-controls/1-form-elements/article.md
```diff
@@ -205,7 +205,9 @@ Here is an example:
 </script>
 ```
 
-Unlike most other controls, `<select>` allows to select multiple options at once if it has `multiple` attribute. That's feature is rarely used. In that case we need to use the first way: add/remove the `selected` property from `<option>` subelements.
+Unlike most other controls, `<select>` allows to select multiple options at once if it has `multiple` attribute. Although such functionality is available, it is rarely used because of the different ways of doing the selection which vary in different operating systems and browser. And also, because you have to inform the user that multiple selection is available, it is more user-friendly to use checkboxes instead.
```
iliakan, line 208: Unlike most other controls, `<select>` allows to select multiple options at once if it has `multiple` attribute. Although such functionality is available, it is rarely used.

iliakan (changes requested):
Make it shorter? More words don't put it clear, I'm afraid.

2-ui/4-forms-controls/1-form-elements/article.md
dorelljames: Got it.

dorelljames:
I've made the necessary changes. Thanks

iliakan:
Thank you.

PR #2197, opened 2020-10-13
Fix possible typos, add a few details in 1.5.11

vsemozhetbyt:


iliakan:
We shouldn't speak too much about Node.js there, I think. I mentioned "other ways".

vsemozhetbyt:
Should I craete a separate PR for other changes or are they erroneous?

iliakan:
I removed "then" and "but", thanks ;)

PR #2200, opened 2020-10-15
Update callback-hell.svg

ish-kafel:
Reuse font glyphs to reduce file size


iliakan:
SVGs are auto-generated from Sketch file. Let me know if I need to fix something in it.

PR #2208, opened 2020-10-16
OPT: add <code> symbol for operators

leviding:
As title.


iliakan:
Headers have no special formatting for code (as they are headers, plain big text). We usually add quotes, e.g. "this" or "??".

PR #2209, opened 2020-10-16
Added to the Intro to JS summary

mikejgurch:
I thought that the summary should also include the fact that JavaScript makes dynamic and interactive webpages via manipulation of HTML and CSS elements. This is one of the main takeaways from the intro.


iliakan:
That's a bit incorrect. JavaScript has `alert` and `prompt`, and a bunch of other methods, not related to HTML and CSS and DOM.

PR #2217, opened 2020-10-20
Fix typo on line 97

jeremylumanbailey:
changed "still execute." to "still executes."


jeremylumanbailey:
Never mind. It is correct as is.

PR #2220, opened 2020-10-20
Add a safety note in 1.6.3 (closure) task solution

vsemozhetbyt:


iliakan:
This is a theory piece that may be added to the main text. Definitely not to a task.

That said, I see so little value in it. Doesn't worth it afaik.

PR #2232, opened 2020-10-24
FIX: clientX/Y and pageX/Y mixed up with regard to change on page scroll

mabouguerra:
fixed `clientX/Y` and `pageX/Y` mixed up with regard to change on page scroll


iliakan:
It's correct in the tutorial.

PR #2246, opened 2020-10-29
Replace outdated backticks in 1.10.1 (try..catch)

vsemozhetbyt:


iliakan:
Not really needed. Let's not confuse translators =)

PR #2261, opened 2020-11-01
Change `var` to `let` in 7.16 regexp-sticky

### LINKED ISSUE #2247: Confusing outdated explanation at "sticky flag y" article
Arcitle: "sticky flag y"
Direct link: https://javascript.info/regexp-sticky

Here at this article as a first exaple we have a string 

` 'let varName = "value"'`

Farther we'll try to execute a regexp /\w+/

> A call to str.match(/\w+/) will find only the first word in the line (var). That’s not it.

And in this statement pop ups a typo. Seems like we should have received 'let' string instead of 'var'.
It would bring up some confusion to newcomers.

wam:
This should resolve #2247.

Commit 10d1b1f25 added a hint about which word would be found by a certain regexp:

> A call to `str.match(/\w+/)` will find only the first word in the line (`var`). That’s not it.

Unfortunately the result is incorrect: the regexp would find `let`, not `var`.

Updating the hint should resolve any confusion. I've double-checked the example, in the console, to ensure that the regexp does indeed find `let`.

### DIFF
--- a/9-regular-expressions/16-regexp-sticky/article.md
+++ b/9-regular-expressions/16-regexp-sticky/article.md
@@ -13,7 +13,7 @@ E.g. we have a code string `subject:let varName = "value"`, and we need to read 
 
 We'll look for variable name using regexp `pattern:\w+`. Actually, JavaScript variable names need a bit more complex regexp for accurate matching, but here it doesn't matter.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Thanks =)

PR #2264, opened 2020-11-03
Added two forward slashes where there is a need

pushpend3r:
It looks like some places you forgot to add two forward slashes to show a comment


iliakan:
it's the "pseudocode"

PR #2265, opened 2020-11-04
Added comment for better understanding

call0fcode:
For a newbie, it's not clear enough at this point that local variables passed in as arguments to the callback are somehow "transferred" and that they are available to use later on the callback we add as a second parameter on the main function. So, the argument we add to the callback will be the one we've passed in internally on the function that called the callback function.


iliakan:
Sorry, but that seems really obvious.

PR #2267, opened 2020-11-05
declaring reassignable variable for Promise object can be danger

lumosmind:
Declaring `let promise` variable can be dangerous. Because reassignable variables makes asynchronous code difficult to debug. We can't know the promise object is resolved or still pending and  we can reassign the promise variable. So when did the promise resolve and which promise object is owner of the result value in consuming code? There will be a suspicious situation.

### DIFF
--- a/1-js/11-async/02-promise-basics/article.md
+++ b/1-js/11-async/02-promise-basics/article.md
@@ -17,7 +17,7 @@ The analogy isn't terribly accurate, because JavaScript promises are more comple
 The constructor syntax for a promise object is:
 
 ```js
-let promise = new Promise(function(resolve, reject) {
+const promise = new Promise(function(resolve, reject) {
   // executor (the producing code, "singer")
 });
 ```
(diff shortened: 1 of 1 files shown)


iliakan:
Here, in the tutorial, we often use `let` instead of `const`.

A lot of programmers prefer `let` to `const`, as it's shorter, and, frankly, there aren't a lot of mistakes because of that. I'd even say I can't remember a single mistake of mine or my colleagues caused by using `let` instead of `const`.

lumosmind:
With your approach const is useless. I think promise object is best fit with const decleration. There is nearly no reason to use let with promise. Also there is a defensive approach about const and let. 
"if you don't need let , use conts"
"if you don't get error , use conts"

iliakan:
I just say what I see =). That's not an approach. That's a description.

Ultimately, what you use: `let` or `const` depends on the context and the intention. 
In given examples, it's unspecified, so `let` is ok.

lumosmind:
You are more experienced than me. That is clear.
But I think this is about theoretical issue.
Do you know any good example to use let with promise? Because I can not se any reason, except "let is convention of this tutorial".

iliakan:
Let's leave it `let` please, for consistency.

PR #2268, opened 2020-11-05
undefined instead of null for empty argument

lumosmind:
Using `undefined` instead of `null` is more convenient for pass empty argument to a function. For the  example snippet, every thing is okay. it works but in general we must send `undefined` valued argument to trigger **default parameter value** mechanism of JavaScript functions. For the example code we can pass any uncallable value instead of `null`, but `undefined` is more convenient for the purpose.

### DIFF
--- a/1-js/11-async/02-promise-basics/article.md
+++ b/1-js/11-async/02-promise-basics/article.md
@@ -291,7 +291,7 @@ function loadScript(src, callback) {
   let script = document.createElement('script');
   script.src = src;
 
-  script.onload = () => callback(null, script);
+  script.onload = () => callback(undefined, script);
   script.onerror = () => callback(new Error(`Script load error for ${src}`));
 
   document.head.append(script);


iliakan:
Sorry, but no. The convention is `null`.

lumosmind:
Do you have any reference for that?

lumosmind:
```
function multiply(a = 2, b = 1) {
  return a * b;
}

// it uses null as zero
console.log(multiply(null, 3));  //returns 0

// it uses default parameter value as 2
console.log(multiply(undefined, 3)); //returns 6
```

iliakan:
A lot. Just google "callback null convention".

lumosmind:
Thanks for keywords. I found some resources but none of them explain the most important question why? is it because of historical reasons. Default parameter syntax released with chrome-49 in 2016.  Everyone refers to node.js coding style as  reason.

iliakan:
Well, as I believe it, we should not explicitly pass around `undefined`. 

We should pass `null` if we mean "value unknown" or "no value". And leave `undefined` to be the system value, saying "variable not defined".

lumosmind:
Yes it is sensible ,but [default parameter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters) mechanism is not compatible with your words. This mechanism forces us to send `undefined `as argument to say "this argument is empty". I think there is a conflict.

thank you for your time

PR #2274, opened 2020-11-09
Update article.md

haond10adp:


iliakan:
"them" seems correct. We make the calls.

PR #2279, opened 2020-11-10
look like ninja code

lumosmind:
It can be more readable with same line of code. There was a lot of nested parenthesis.
This line deserve a place in literature "layers of hell" :)
`}))));`


iliakan:
Yeah, better, thanks ;)

PR #2284, opened 2020-11-13
Updated the info about VS

damla:
Hi there,

I've updated the information about Visual Studio Code because it is not an IDE according to [Visual Studio Code FAQ](https://code.visualstudio.com/docs/supporting/faq). In addition, I've changed Visual Studio as cross-platform IDE.

### DIFF
--- a/1-js/01-getting-started/3-code-editors/article.md
+++ b/1-js/01-getting-started/3-code-editors/article.md
@@ -15,7 +15,7 @@ If you haven't selected an IDE yet, consider the following options:
 - [Visual Studio](https://visualstudio.microsoft.com) (cross-platform, free/paid).
 - [WebStorm](http://www.jetbrains.com/webstorm/) (cross-platform, paid).
 
-There's also "Visual Studio Code", not to be confused with "Visual Studio". "Visual Studio" is paid and well-suited IDE for the .NET platform. It's also good at JavaScript. Detailed information about their difference can be found in [Visual Studio Code FAQ](https://code.visualstudio.com/docs/supporting/faq) page.
+There's also "Visual Studio Code", not to be confused with "Visual Studio". "Visual Studio" is paid and well-suited IDE for the .NET platform. It's also good for JavaScript. There is also a free edition of "Visual Studio" called [Visual Studio Community](https://www.visualstudio.com/vs/community/), albeit with limited features. Detailed information about the difference between "Visual Studio" and "Visual Studio Code" can be found in [Visual Studio Code FAQ](https://code.visualstudio.com/docs/supporting/faq) page.
 
 Many IDEs are paid, but have a trial period. Their cost is usually negligible compared to a qualified developer's salary, so just choose the best one for you. In addition, student packs are found in most of them.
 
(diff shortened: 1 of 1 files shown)


1-js/01-getting-started/3-code-editors/article.md
```diff
@@ -12,12 +12,12 @@ An IDE loads the project (which can be many files), allows navigation between fi
 
 If you haven't selected an IDE yet, consider the following options:
 
-- [Visual Studio Code](https://code.visualstudio.com/) (cross-platform, free).
+- [Visual Studio](https://visualstudio.microsoft.com) (cross-platform, free/paid).
 - [WebStorm](http://www.jetbrains.com/webstorm/) (cross-platform, paid).
 
-For Windows, there's also "Visual Studio", not to be confused with "Visual Studio Code". "Visual Studio" is a paid and mighty Windows-only editor, well-suited for the .NET platform. It's also good at JavaScript. There's also a free version [Visual Studio Community](https://www.visualstudio.com/vs/community/).
+There's also "Visual Studio Code", not to be confused with "Visual Studio". "Visual Studio" is paid and well-suited IDE for the .NET platform. It's also good at JavaScript. Detailed information about their difference can be found in [Visual Studio Code FAQ](https://code.visualstudio.com/docs/supporting/faq) page.
```
MuhammedZakir, line 18:
```suggestion
There's also "Visual Studio Code", not to be confused with "Visual Studio". "Visual Studio" is paid and well-suited IDE for the .NET platform. It's also good for JavaScript. There is also a free edition of "Visual Studio" called [Visual Studio Community](https://www.visualstudio.com/vs/community/), albeit with limited features. Detailed information about the difference between "Visual Studio" and "Visual Studio Code" can be found in [Visual Studio Code FAQ](https://code.visualstudio.com/docs/supporting/faq) page.
```

I think we should mention about free version as many tend to skip softwares when they see the word "paid".

Is it "at JavaScript" or "for JavaScript"?

MuhammedZakir: Btw, feel free to restructure the paragraph as you see fit! :-)

damla: Hi @MuhammedZakir,

Thanks for the feedback. I totally agree with all of them.

Talking about Visual Studio & VSCode as a paragraph under code editors part seems unrelated for me too. Therefore, It may be better to remove those sentences as you mentioned.

iliakan:
What's IDE depends on the definition of "IDE"

Microsoft has some marketing there. That's why they say "VS Code" isn't IDE.
It actually is IDE, by definition.

iliakan:
Github plugins, debugging, project-level autocompletion, terminal - all these are IDE features.

MuhammedZakir:
Under "lightweight editors", line 28:

```
The main difference between a "lightweight editor" and an "IDE" is that an IDE works on a project-level, so it loads much more data on start, analyzes the project structure if needed and so on. A lightweight editor is much faster if we need only one file.
In practice, lightweight editors may have a lot of plugins including directory-level syntax analyzers and autocompleters, so there's no strict border between a lightweight editor and an IDE.
```

I think VS Code falls in above category described above. It's an "editor", but can be extended to be used as an IDE. Same can be said about Emacs and Vim\*.

>  [...]
> Microsoft has some marketing there. That's why they say "VS Code" isn't IDE.
> It actually is IDE, by definition.

You're probably right about marketing.  But, for consistency, mainly because VS Code is usually introduced as an editor in community (with IDE features), adding it under editors is better IMO.

\* Vim - slightly stretching things here, but...

iliakan:
let's just not argue about that, please please

PR #2288, opened 2020-11-17
Promise.any

lumosmind:
I think `Promise.any()` must be in this article. It is very critical when we work with multiple source content.

### DIFF
--- a/1-js/11-async/05-promise-api/article.md
+++ b/1-js/11-async/05-promise-api/article.md
@@ -302,7 +302,7 @@ There are 5 static methods of `Promise` class:
     - `status`: `"fulfilled"` or `"rejected"`
     - `value` (if fulfilled) or `reason` (if rejected).
 3. `Promise.race(promises)` -- waits for the first promise to settle, and its result/error becomes the outcome.
-4. `Promise.any(promises)` -- waits for the first promise to fulfill, and its result becomes the outcome.If all of the given promises rejects, it becomes the error of `Promise.any`.
+4. `Promise.any(promises)` -- waits for the first promise to fulfill, and its result becomes the outcome. If all of the given promises rejects, it becomes the error of `Promise.any`.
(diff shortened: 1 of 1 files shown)


1-js/11-async/05-promise-api/article.md
```diff
@@ -217,6 +217,29 @@ Promise.race([
...
+
+```js run
+Promise.race([
+  new Promise((resolve, reject) => setTimeout(() => reject(new Error("Whoops!")), 1000)),
+  new Promise((resolve, reject) => setTimeout(() => resolve(1), 2000)),
+  new Promise((resolve, reject) => setTimeout(() => resolve(3), 3000))
+]).then(alert); // 1
+```
+
+The first promise here was fastest, but it rejecets, so the second promise became the result. After the first fulfilled promise "wins the race", all further results are ignored.
```
vsemozhetbyt, line 240:
```suggestion
The first promise here was fastest, but it was rejected, so the second promise became the result. After the first fulfilled promise "wins the race", all further results are ignored.
```

1-js/11-async/05-promise-api/article.md
```diff
@@ -217,6 +217,29 @@ Promise.race([
...
+The syntax is:
+
+```js
+let promise = Promise.any(iterable);
+```
+
+For instance, here the result will be `1`:
+
+```js run
+Promise.race([
```
iliakan, line 233: Promise.race example here?

iliakan (changes requested):
Please fix the example.
And maybe add it to the Summary section?

1-js/11-async/05-promise-api/article.md
lumosmind: I did some changes to meet requirements

iliakan:
Thank you!

PR #2292, opened 2020-11-18
Buffer purpose. And missing "s".

joaquinelio:


joaquinelio:
Just check it 
I don't mean doing the change by my own, maybe I'm just getting slow.

** Rate limiting
It wasn't clear to me until I read the code, 
the buffer starts with the byte count with .send, then decrements.
 
And yes, the missing S

PR #2300, opened 2020-11-19
Rephrased for better clarity.

patrikbego:


iliakan:
Why do we need `ben`?

Also, "Using objects as keys is one of the most notable and important `Map` features. The same does not count for `Object`. String as a key in `Object` is fine, but we can't use another `Object` as a key in `Object`." -- probably, should be rephrased. Or "The same does not count for `Object`" should be removed/replaced.

iliakan (changes requested):
Why do we need `ben`?

Also, "Using objects as keys is one of the most notable and important `Map` features. The same does not count for `Object`. String as a key in `Object` is fine, but we can't use another `Object` as a key in `Object`." -- probably, should be rephrased. Or "The same does not count for `Object`" should be removed/replaced.

patrikbego:
> Why do we need `ben`?
> 
> Also, "Using objects as keys is one of the most notable and important `Map` features. The same does not count for `Object`. String as a key in `Object` is fine, but we can't use another `Object` as a key in `Object`." -- probably, should be rephrased. Or "The same does not count for `Object`" should be removed/replaced.

...

When I have read the sentence: 
"For string keys, `Object` can be fine, but not for object keys."
It was not really clear to me what you mean. And then I looked at the code example and still did not get it. When I played a bit with the code, I actually realized what you mean... 
I think if you add another object "Ben" it makes it a bit more clear. 

Regarding rephrasing, I am open for the idea, if you think it makes sense :)

PR #2304, opened 2020-11-20
Fix solution in data types - string - truncate

johnnycook:
`truncate('Hello',2) // 'He…'` because slice takes substring from the first param to (but not including) second param.


iliakan:
the tests won't pass that way

PR #2305, opened 2020-11-21
chore: update article.md

abhishekjakhar:
Right now while reading this line, it reads like "In this case this is undefined in strict mode", which I think is not correct.


iliakan:
it was correct

PR #2310, opened 2020-11-24
fix nested markdown

joaquinelio:


joaquinelio:
btw,
grammar. Is it ok ?  My English teacher told me that's an horror many decades ago. 

but no. -> but it doesn't.  / but it does not.

`249 It might seem that the line `(*)` would cause an error, but no. 
`

iliakan:
Thank you. Maybe @paroche can tell if that phrase is correct =)

My English grammar is so far from perfection 🤦

paroche:
I made various minor revisions to the punctuation and phrasing of the article and applied them.. W/ respect to "but no", it's a pretty casual phrasing, and its use would depend on context, but for this I decided that "but it does not" would be a bit better. I may go over it again to see if I want to revise my revisions.

paroche:
AOK?

iliakan:
@paroche You made a new PR with these, right?

paroche:
Yes, and merged it. Three times, actually -- after making the changes I went back and re-read the article and made further refinements. Then I did it again. Then I let it be. Mostly just little wording & punctuation changes as it says. A little more filling out for the analogies (piece of paper & cabinet).

iliakan:
Thank you! =)

PR #2315, opened 2020-11-25
Fix possible typos in 4.1 (ArrayBuffer, binary arrays)

vsemozhetbyt:
Especially not sure about the "common discriminator -> common denominator" change. Let me know if I am wrong.


iliakan:
Yeah, you're totally right, `denominator` is the correct term.

PR #2319, opened 2020-11-26
revoked proxies purpose

joaquinelio:
"The target object can be garbage-collected after that."  
  yes... but no   : )  @iliakan  
"object" in this code is still alive, "proxy" disables itself and kills just a target clone inside it.
 
```
...
// the proxy isn't working any more (revoked)
alert(proxy.data); // Error
alert(object.data); //  still working
```
..unless I add right after proxy creation: 
`object = proxy;  //   transparency 
`
so it performs like the rest of the examples, a replacement.   (let {object, revoke} = .... object didnt work)
I would add it to the PR, but I can't see the point of revocable so I don't know what's useful for.

The "distictive error" is the only thing I see different from a plain delete, so I suggested it.

### DIFF
--- a/1-js/99-js-misc/01-proxy/article.md
+++ b/1-js/99-js-misc/01-proxy/article.md
@@ -963,7 +963,7 @@ revoke();
 alert(proxy.data); // Error
 ```
 
-A call to `revoke()` removes all internal references to the target object from the proxy, so they are no longer connected. The target object can be garbage-collected after that.
+A call to `revoke()` removes all internal references to the target object from the proxy, so they are no longer connected. Any further attempt to access it will throw a distinctive error you can catch.
 
 We can also store `revoke` in a `WeakMap`, to be able to easily find it by a proxy object:
 
--- a/1-js/99-js-misc/01-proxy/article.md
+++ b/1-js/99-js-misc/01-proxy/article.md
@@ -963,7 +963,7 @@ revoke();
 alert(proxy.data); // Error
 ```
 
-A call to `revoke()` removes all internal references to the target object from the proxy, so they are no longer connected. Any further attempt to access it will throw a distinctive error you can catch.
+A call to `revoke()` removes all internal references to the target object from the proxy, so they are no longer connected. The target object can be garbage-collected after that. But unlike regular proxies, Proxy.revocable() creates a clon of the original object as a target, so you can still access to object.data.   
 
 We can also store `revoke` in a `WeakMap`, to be able to easily find it by a proxy object:
 
--- a/1-js/99-js-misc/01-proxy/article.md
+++ b/1-js/99-js-misc/01-proxy/article.md
@@ -964,7 +964,7 @@ revoke();
 alert(proxy.data); // Error
 ```
 
-A call to `revoke()` removes all internal references to the target object from the proxy, so they are no longer connected. The target object can be garbage-collected after that.   
+A call to `revoke()` removes all internal references to the target object from the proxy, so they are no longer connected. The target object could be garbage-collected after that.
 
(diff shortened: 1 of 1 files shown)


iliakan:
You mean, the object isn't garbage-collected, because there's a reference to it `object`, right?

joaquinelio:
woa that was fast 
is this a chat room?

yes
just came from plunker
object = proxy  works ok,  revoke kills them both
but I still dont see the point of revocables

iliakan:
Maybe "The target object and proxy can be garbage-collected independently after that." ?

iliakan:
In real life I never saw "revoke" being used. That said, it's still the part of the spec.

joaquinelio:
not very clear
Proxy  internal data maybe garbage, proxy itself seems alive but broken, not garbage 
but I wouldnt mention "target" clon inside, we never access it but via proxy

joaquinelio:
>TypeError: Cannot perform 'get' on a proxy that has been revoked  
object is reachable,
proxy is reachable.   just internal data is not,  

**"target can be garbage collected"  is wrong if  we think it's the same object**

so,
object = proxy ;
works,  
But I dont see the point!

I was hoping you showing me some use for it.
 //

joaquinelio:
1)  "Any further attempt to access it will throw a distinctive error you can catch."  (we can catch)
not a good place. 
I would add it though, line 937.

2) about explaining " proxy internal cloned object being garbage", is it worth? 
I would just remove  "The target object can be garbage-collected after that."

joaquinelio:
Ilya !!!
Being revocable is not the important thing about this proxy.
It' the target clone part.

Next time you need a proxy, you should ask yourself: 
Do I need to overwrite it with the proxied version OR 
a cloned independent one?

For start, revocable doesn't have the danger of using the unproxied version, 
you may have lots of proxies with the same target, all clones, every one with its own data.

Can't figure a use yet, but I think the purpose will be entirely different from regular proxies.

iliakan:
@joaquinelio Here's the spec: https://tc39.es/ecma262/#sec-proxy.revocable

There's nothing like cloning (AKA deep copying or shallow copying) in there.

joaquinelio:
You are right.

I'm really, really sorry.
In my test I put   
`object = {data:"xxxx"}`  instead of  `object.data = "xxxx"` making it an entirely different object
that's embarrasing, all the stuff and waste of your time based on that bug

so, 
yes, it wont be garbage collected because of object {} reference

>"The target object and proxy can be garbage-collected independently after that."

I dont think  so, "independentky". You have to kill both for the garbage to go.

But the change of text can be small.

iliakan:
Please update the PR to reflect your latest suggestion?

joaquinelio:
Choose one, I couldn't think the best:

Fixing the code 

`object = null;   // use proxy from now on     . 
` 
still easier than a revoke if I may add..
or 

`object = proxy; proxy = null;  // transparent proxied   
`    // Is it ok?

Fixing the text 

Removing the garbage collected mention, 

or
Adding explanation.  Couldnt do it short without touching the code. something like 
(...) If the reference to the original object is lost, (by making it null after proxy creation to ensure integrity, or because you are returning only the proxy stuff from the function that creates it, then (finally)
... The target object can be garbage-collected after that.  
 That's insane

joaquinelio:
Done "last suggestion" 
COULD be garbage, "but
>, but in our example `object` variable is still referencing it.

I could add
>We need to dispose it first,  this makes to best practice too:  Once proxied, dont mess up using the unproxied version.

But I would go for  
`object = null;  //  use proxy from now on`  
in the code instead. 

too many silly commits, thats embarrasing.  If you like something, just cut and paste.

Waiting for this resolution,
finished proxy translation, with it the first section. 
For the entire site: 18 to go.

PR #2326, opened 2020-11-29
care the size

joaquinelio:
Of course, I can't "fix" Spanish site if you don't agree. 
I think this is important.

Again, I don't "need" the commit, drop it and replace it with whatever you consider, but:

this is a tutorial, we care best practices, and this is real world example
in real world download progress is best suited for big files...
so, the warning
and the advice next to it

### DIFF
--- a/5-network/03-fetch-progress/article.md
+++ b/5-network/03-fetch-progress/article.md
@@ -109,4 +109,10 @@ Let's explain that step-by-step:
 
 At the end we have the result (as a string or a blob, whatever is convenient), and progress-tracking in the process.
 
+```warn header="Best practices: care the size"
+It is easy to make a blob. That doesn't mean always adviceable. Many apps/OS lags come from poorly managed memory.
+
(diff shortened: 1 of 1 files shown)


iliakan:
I added the relevant comment there. "Stream" processing is extremely rare in client apps. Downloaded size control is a good thing to have =)

joaquinelio:
Ok, you dont have much control client side...  say to save to disk or to send to another app

not convinced yet...
You are right, maybe because "download progress" is not the place.  

Insanely hungry apps is an isue, 
and I found this blob thing a perfect place to warn about it: Usually there is another way.  
again, not the place.
But many will go to node (or whatever is cooking right now) and will be tempted to manage may blobs, maybe small docs but hundreds of them in memory instead of just passing the chunks to client and release. 

When in the near future internet breaks, 
you'll know is your fault.

iliakan:
We mention here about the memory and the possibility of huge downloads. Hope it's good enough.

joaquinelio:
yeah,
Im only a maniac

The article is more than ok now,  (not just good enough)
but when an issue traps me is not easy to let it go. 

time to sync, thanks

PR #2341, opened 2020-12-03
Update article.md in section 2.8

vsoni101:
In the section "String concatenation with binary +", the example of 2 + 2 + "1" was given, but there was no example for when the first operand is a string.

I have added another example of "1" + 2 + 2, which is similar to the example stated before it making it easier for the reader to understand. It's just a small thing but it's kind of important.

### DIFF
--- a/1-js/02-first-steps/08-operators/article.md
+++ b/1-js/02-first-steps/08-operators/article.md
@@ -109,7 +109,7 @@ Here, operators work one after another. The first `+` sums two numbers, so it re
 ```js run
 alert('1' + 2 + 2); // "122" and not "14"
 ```
-Here, the first operand is a string, the compiler treats the other two operands as strings too. The `2` gets concatenated to `'1'`, so it's like `'1' + 2 = '12'` and `'12' + 2 = '122'`.
+Here, the first operand is a string, the compiler treats the other two operands as strings too. The `2` gets concatenated to `'1'`, so it's like `'1' + 2 = "12"` and `"12" + 2 = "122"`.
 
 The binary `+` is the only operator that supports strings in such a way. Other arithmetic operators work only with numbers and always convert their operands to numbers.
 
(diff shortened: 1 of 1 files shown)


1-js/02-first-steps/08-operators/article.md
```diff
@@ -106,6 +106,11 @@ alert(2 + 2 + '1' ); // "41" and not "221"
 
 Here, operators work one after another. The first `+` sums two numbers, so it returns `4`, then the next `+` adds the string `1` to it, so it's like `4 + '1' = 41`.
 
+```js run
+alert('1' + 2 + 2); // "122" and not "14"
+```
+Here, the first operand is a string, the compiller treats the other two operands as string too. The `2` gets concatinated to `'1'`, so it's like `'1' + 2 = "12"` and `"12" + 2 = "122"`
```
vsemozhetbyt, line 112: Not sure, but maybe (correct possible typos and unify quotes):

```suggestion
Here, the first operand is a string, the compiler treats the other two operands as strings too. The `2` gets concatenated to `'1'`, so it's like `'1' + 2 = '12'` and `'12' + 2 = '122'`.
```

MuhammedZakir: For consistency in quotes with above paragraphs, it's better to change

```
`'1' + 2 = '12'` and `'12' + 2 = '122'`.
```
to 

```
`'1' + 2 = "12"` and `"12" + 2 = "122"`.
```

This is not a must, and more like a personal preference.

vsoni101: Yes, that's why I put single quotes there in the initial commit, but a suggestion was made by a contributor, so, I thought he must be right because of more experience in this project.
Should I change those quotes now or not?

vsemozhetbyt: Not so much experience, sorry) It was just a suggestion, so feel free to stick to your initial thought)

iliakan:
Isn't it obious?

vsoni101:
> Isn't it obvious?

It wasn't obvious to me. I came from C++, when I first learned JS, maybe that's the reason.

PR #2346, opened 2020-12-04
Discourage use of automatic variable creation

orionrush:
I've added explicit declarations of variables and not relied on automatic variable creation as it is [not the best of practices] (https://stackoverflow.com/a/6853267/362445) and not possible in IE5 strict mode. It also has different behaviour in IE vs other browsers. 

I've also added `<lable>` elements where it makes sense to do so, for semantic purposes, and as it is also considered best practice, and improves accessibility.

### DIFF
--- a/2-ui/4-forms-controls/2-focus-blur/article.md
+++ b/2-ui/4-forms-controls/2-focus-blur/article.md
@@ -27,19 +27,21 @@ In the example below:
   #error { color: red }
 </style>
 
-Your email please: <input type="email" id="input">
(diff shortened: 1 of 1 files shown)


iliakan:
This is wrong: https://stackoverflow.com/questions/6852883/can-i-use-an-id-as-a-variable-name/6853267#6853267

Automatic vars creation was added to HTML5, it's a part of the standard AFAIC.

Let's keep using it, to make examples shorter.

Labels can be added, though, as a separate PR =)

PR #2351, opened 2020-12-05
Fix typos in 7.3 (JavaScript animations)

vsemozhetbyt:


iliakan:
I'm ashamed. How can I write with that many mistakes!

vsemozhetbyt:
For such an amount of wonderful data, these tiny mistakes are just something that proves you are a real human being)

PR #2361, opened 2020-12-07
Fixed small typo

RapTho:
2-ui / 2-events / 5-dispatch-events

In subchapter "Event constructor": first word use "Built-in" instead of "Build-in"


RapTho:
fixed 2 small typos

PR #2375, opened 2020-12-09
Correct example in 9.14 (Lookahead and lookbehind)

vsemozhetbyt:
Without the `g` flag, we cannot say that the price is skipped.
Without the `\b` assertion, we will have the part of the price in the result.


iliakan:
Good catch! 😊

PR #2389, opened 2020-12-14
Add information about virtual/mobile keyboards and an additional keyboard event inspection tool.

Danziger:
There's a comment from a user on this page that says "This works well on desktop but fails on Android".

I replied with what might be the issue (it's usually not possible to identify most pressed keys on mobile keyboards), but it looks like that message needs to be approved first and might be easy to miss, so I added an additional section on this page to explain this fairly common issue.

### DIFF
--- a/2-ui/3-event-details/7-keyboard-events/article.md
+++ b/2-ui/3-event-details/7-keyboard-events/article.md
@@ -174,7 +174,7 @@ There were so many browser incompatibilities while working with them, that devel
 
 ## Mobile Keyboards
 
-When using virtual/mobile keyboards, formally know as IME (Input-Method Editor), the W3C standard states that a KeyboardEvent's [`e.keyCode` should be `229`](https://www.w3.org/TR/uievents/#determine-keydown-keyup-keyCode) and [`e.key` should be `"Unidentified"`](https://www.w3.org/TR/uievents-key/#key-attr-values).
+When using virtual/mobile keyboards, formally known as IME (Input-Method Editor), the W3C standard states that a KeyboardEvent's [`e.keyCode` should be `229`](https://www.w3.org/TR/uievents/#determine-keydown-keyup-keyCode) and [`e.key` should be `"Unidentified"`](https://www.w3.org/TR/uievents-key/#key-attr-values).
 
 While some of these keyboards might still use the right values for `e.key`, `e.code`, `e.keyCode`... when pressing certain keys such as arrows or backspace, there's no guarantee, so your keyboard logic might not always work on mobile devices.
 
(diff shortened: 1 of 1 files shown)


2-ui/3-event-details/7-keyboard-events/article.md
```diff
@@ -170,6 +172,12 @@ In the past, there was a `keypress` event, and also `keyCode`, `charCode`, `whic
 
 There were so many browser incompatibilities while working with them, that developers of the specification had no way, other than deprecating all of them and creating new, modern events (described above in this chapter). The old code still works, as browsers keep supporting them, but there's totally no need to use those any more.
 
+## Mobile Keyboards
+
+When using virtual/mobile keyboards, formally know as IME (Input-Method Editor), the W3C standard states that a KeyboardEvent's [`e.keyCode` should be `229`](https://www.w3.org/TR/uievents/#determine-keydown-keyup-keyCode) and [`e.key` should be `"Unidentified"`](https://www.w3.org/TR/uievents-key/#key-attr-values).
```
vsemozhetbyt, line 177: Maybe:

```suggestion
When using virtual/mobile keyboards, formally known as IME (Input-Method Editor), the W3C standard states that a KeyboardEvent's [`e.keyCode` should be `229`](https://www.w3.org/TR/uievents/#determine-keydown-keyup-keyCode) and [`e.key` should be `"Unidentified"`](https://www.w3.org/TR/uievents-key/#key-attr-values).
```

Danziger: Good catch 👌

iliakan:
Thanks, it's a great addition, about the mobile keyboard.

What's good in https://keyjs.dev/, that doesn't exist on the keyboard stand here? Looks like it shows all the same.

Danziger:
@iliakan It shows `which` and `keyCode` as well and includes `keypress`, which might be relevant if you need to support old legacy browsers or work with outdated libraries that still rely on those. Also, if you click "Show More" you'll see some other KeyboardEvent properties such as `repeat`, `location` or the 4 meta keys (which are referenced on that page), as well as some additional deprecated ones.

If you enter an Alt Code or use it on a mobile keyboard that doesn't use the right key-specific codes, it will display an additional explanation for that (similar to the one I've included in the PR).

I'll soon add a table of keycodes that includes a "reliability" column so that you know what values are safe to use, which ones would usually work on English keyboard layouts and which other ones are browser or layout specific and you should try to avoid.

iliakan:
Happy new Year!

These seem to be minor improvements. It seems better to add them to the in-page keyboard test stand, rather than send a visitor away.

Danziger:
@iliakan I've just removed the link so that you can merge the other change. If you still want to modify the snippet, I think it would be better to create a separate issue, as you might want to first figure out what you want to display and how (other than just logging it into the textarea) so that it doesn't look too overwhelming for someone that is just getting started.

iliakan:
Thank you!

PR #2393, opened 2020-12-16
Add link to Kotlin/JS

tokou:


iliakan:
Is it popular enough?

tokou:
Well it depends on your definition of popularity and what to add or not to add to the list :)

- Kotlin as a programming language is more popular than many already listed (like Dart, Flow or CoffeeScript) [Google Trends](https://trends.google.com/trends/explore?q=Brython,Kotlin,%2Fm%2F0hjc5m0,%2Fm%2F0h52xr1,Flow%20JS)
- Kotlin JS is not a toy project or an external effort, it is part of the direction that Kotlin is taking to support multiple platforms
- Should it be in the list? Well it's your call of what to add and not to add, but personally, as I see something like Brython on the list I think Kotlin JS is similar and should be on it as well

iliakan:
Okay.

PR #2397, opened 2020-12-18
Modify shortened indexOf algorithm

TaylorClay:
While iterating over the string, the algorithm can jump by the length of the target rather than simply incrementing by 1

### DIFF
--- a/1-js/05-data-types/03-string/article.md
+++ b/1-js/05-data-types/03-string/article.md
@@ -274,7 +274,7 @@ let target = "as";
 
 *!*
 let pos = -1;
-while ((pos = str.indexOf(target, pos += target.length)) != -1) {
+while ((pos = str.indexOf(target, pos + target.length)) != -1) {
   alert( pos );
 }
 */!*
--- a/1-js/05-data-types/03-string/article.md
+++ b/1-js/05-data-types/03-string/article.md
@@ -266,7 +266,7 @@ while (true) {
 }
 ```
 
-The same algorithm can be layed out shorter:
+The same algorithm can be layed out shorter and more efficient:
 
 ```js run
 let str = "As sly as a fox, as strong as an ox";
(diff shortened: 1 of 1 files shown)


1-js/05-data-types/03-string/article.md
```diff
@@ -266,15 +266,15 @@ while (true) {
...
+The same algorithm can be layed out shorter and more efficient:
 
 ```js run
 let str = "As sly as a fox, as strong as an ox";
 let target = "as";
 
 *!*
 let pos = -1;
-while ((pos = str.indexOf(target, pos + 1)) != -1) {
+while ((pos = str.indexOf(target, pos += target.length)) != -1) {
```
vsemozhetbyt, line 277: Maybe:

```suggestion
while ((pos = str.indexOf(target, pos + target.length)) != -1) {
```

It seems there is no reason to reassign `pos` twice here.

TaylorClay: You're absolutely right @vsemozhetbyt. I'm so used to `+=`, it was basically a reflex. Thanks!

vsemozhetbyt:
I think this would be a different algorithm. Compare:

```js
let str = "aaa";
let target = "aa";

let pos = -1;

while ((pos = str.indexOf(target, pos + 1)) != -1) {
  console.log( pos );
}

// 0
// 1
```
```js
let str = "aaa";
let target = "aa";

let pos = -1;

while ((pos = str.indexOf(target, pos + target.length)) != -1) {
  console.log( pos );
}

// 1
```

TaylorClay:
Touché @vsemozhetbyt, I did not put enough thought into this. It occurred to me while skimming, so I thought it would be nice to contribute as it worked in two simple cases I tried. I absolutely should have thought through a few more test cases before tossing up a PR. I will close this. Thanks again!

vsemozhetbyt:
Do not give up and good luck with new contributions!

PR #2405, opened 2020-12-24
removed async keyword

yashdesoi:
In fourth example we had async keyword in front of the function declaration, which was unnecessary since author is showing how we can explicitly return a promise without async keyword.


iliakan:
We're talking about async functions here, that's why =)

PR #2411, opened 2020-12-30
Add sample for split limit argument

behnammodi:
- added a sample for limit 
- fix some output

### DIFF
--- a/9-regular-expressions/17-regexp-methods/article.md
+++ b/9-regular-expressions/17-regexp-methods/article.md
@@ -95,13 +95,13 @@ Splits the string using the regexp (or a substring) as a delimiter.
 We can use `split` with strings, like this:
 
 ```js run
-alert('12-34-56'.split('-')) // array of [12, 34, 56]
+alert('12-34-56'.split('-')) // array of ['12', '34', '56']
 ```
 
 But we can split by a regular expression, the same way:
 
 ```js run
-alert('12, 34, 56'.split(/,\s*/)) // array of [12, 34, 56]
+alert('12, 34, 56'.split(/,\s*/)) // array of ['12', '34', '56']
 ```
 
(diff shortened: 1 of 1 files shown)


9-regular-expressions/17-regexp-methods/article.md
```diff
@@ -95,13 +95,19 @@ Splits the string using the regexp (or a substring) as a delimiter.
...
 ```
 
 But we can split by a regular expression, the same way:
 
 ```js run
-alert('12, 34, 56'.split(/,\s*/)) // array of [12, 34, 56]
+alert('12, 34, 56'.split(/,\s*/)) // array of ['12', '34', '56']
+```
+
+Also we can use `limit` for limit output, like this:
```
vsemozhetbyt, line 107: Maybe: 'for limited output' or 'to limit output'.

behnammodi: @vsemozhetbyt Thanks, I'll fix it

behnammodi:
@iliakan Why? it's not polite, close PR without explanation.

iliakan:
First, thanks for the PR.

This section is about regular expressions. So more details about `str.split` seems extraneous.

Usually when I ask people to fix their PRs, they don't do that.
That's why I took the first change (array quotes) and applied it in the commit https://github.com/javascript-tutorial/en.javascript.info/commit/6418344d594b08069968f1715472ccfc1dbb4f86.

Sorry for any offence, and thanks for the contribution!

behnammodi:
Thanks for the response, but I think we can add a back-link for `str.split`

iliakan:
We could, but this method is so obvious...

behnammodi:
@iliakan Ok man 😀 , you're right

PR #2413, opened 2020-12-31
Change source of the term 'microtask'

### LINKED ISSUE #2412: Source of the term 'microtask'
Source of the term `microtask` was changed to `ES8` from `V8` in [this](https://github.com/javascript-tutorial/en.javascript.info/commit/38f04689104a1a393d1f06b3784cb9ec4d5a827b) commit.

I think it should be changed to `V8` instead of `ES8`.

raycon:
Fix #2412

This modification changes the source of the term from `ES8` to `V8`.


iliakan:
Thanks!

PR #2422, opened 2021-01-09
Update currying function to use bind instead of wrapper pass

lapstjup:
I think this change will help us to dodge concatenating new arguments with lexical ones and make the currying function much simpler.

P.S. - Using ternary instead of if and else will shorten this more but didn't do it for the sake of readability for now.

### DIFF
--- a/1-js/99-js-misc/03-currying-partials/article.md
+++ b/1-js/99-js-misc/03-currying-partials/article.md
@@ -150,7 +150,7 @@ The result of `curry(func)` call is the wrapper `curried` that looks like this:
 function curried(...args) {
   if (args.length >= func.length) { // (1)
     return func.apply(this, args);
-  } else return curried.bind(this, ...args);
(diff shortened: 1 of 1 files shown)


iliakan:
Are you sure? Did you test it well?

lapstjup:
I did test it with an object as this and without it for like 4 - 5 currying level of arguments. 
Do you have any specific edge cases to test for ? I can perform those.

iliakan:
I merged and changed the text a bit, made it shorter and simpler, hope that works.

Thank you!

lapstjup:
@iliakan Hey man. I shared this approach with others and @bendtherules pointed out that bound function will not work when we need to change the context later on. 

So if you have, curriedSum(1,2).call({},3) - Here the `this` will still point to globalThis and not take the new one since that's how bind works. 

Sorry for the trouble !!

You can revert this to what it was.

iliakan:
Well, that's probably the reason why it was the way it was.

PR #2424, opened 2021-01-09
confusion between arguments and parameters

lumosmind:
when we call a function with some values in braces these values are called as "arguments" instead of "parameters".


iliakan:
Function parameters are also called arguments.

iliakan:
I'm closing this unless there's more argumentation =)

MuhammedZakir:
Although both of them are used interchangeably, function parameters and arguments are indeed different. [1] I am in favor of making that distinction, but I think it might create more confusion.

[1] https://stackoverflow.com/questions/1788923/parameter-vs-argument

lumosmind:
These terms are generally used wrongly. And if this wrong usage get place in this tutorial it can be shame perspective of engineering.

If we loose difference between **parameters** and **arguments** we can not make technical conversations properly.

[MDN](https://developer.mozilla.org/en-US/docs/Glossary/Parameter)
[microsoft docs](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/parameters-and-arguments)

iliakan:
So:
- we use "parameters" when talking about what's written inside the function declaration,
- we use "arguments" when talking about values, passed to a function.

Correct?

lumosmind:
> So:
> 
> * we use "parameters" when talking about what's written inside the function declaration,
> * we use "arguments" when talking about values, passed to a function.
> 
> Correct?

absolutely right.

iliakan:
Could you please look for parameters/arguments in the text then and update? So that we'll have all fixed appropriately.

lumosmind:
> Could you please look for parameters/arguments in the text then and update? So that we'll have all fixed appropriately.

I found only one conflict in rest of the article and fixed it.

MuhammedZakir (approved):

iliakan:
Modified SVG manually in the figures.sketch file.

iliakan:
Thanks!

PR #2426, opened 2021-01-11
"bigint" is also numeric type

lumosmind:
To keep question simple and precise we could change "numeric" with "number typed". Because term numeric involves both "number" type and "bigint" type. But the solution and test cases want us to solve this question for "number" type.

### DIFF
--- a/1-js/04-object-basics/01-object/8-multiply-numeric/task.md
+++ b/1-js/04-object-basics/01-object/8-multiply-numeric/task.md
@@ -2,9 +2,9 @@ importance: 3
 
 ---
 
-# Multiply numeric property values by 2
+# Multiply number typed property values by 2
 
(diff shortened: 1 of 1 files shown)


iliakan:
There's a type "number". And we'd better not confuse people using the same word for bigints, as they are "bigint".

Both are numerical types. Similar to many other languages, where there are "float" for decimals, "int" for integers, both are numeric types.

lumosmind:
I think I couldn't express my ideas. Because we are talking about same things.  If we use "numeric" instead of "number typed", we need to expand solution to cover bigint type. we need to emphasize number type or change the solution.

iliakan:
Oh, I see now, sorry.

You are correct. Although, "number typed" feels a bit awkward..

lumosmind:
> Although, "number typed" feels a bit awkward..

I just try to create solution but English is not my main language :)

PR #2427, opened 2021-01-12
declarations vs calls

lumosmind:
I think article talks about "closures". If we talk about closures, calling a function in another function does not create links if arguments are not objects.
So if the subject is "closures", it should be "declarations" instead of "calls".

### DIFF
--- a/1-js/04-object-basics/03-garbage-collection/article.md
+++ b/1-js/04-object-basics/03-garbage-collection/article.md
@@ -15,7 +15,7 @@ Simply put, "reachable" values are those that are accessible or usable somehow. 
     For instance:
 
     - Local variables and parameters of the current function.
-    - Variables and parameters for other functions on the current chain of nested calls.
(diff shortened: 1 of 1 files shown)


iliakan:
Actually, "calls" is correct here. Variables that are "outside" are kept due to closures mechanics, not because they are roots.

lumosmind:
Can you give an example? I couldn't understand what you mean.
Can nested called function reach outer lexical environment?

iliakan:
Just a nested call, like here:

```js
f();

function f() {
  g()
}

function g() {
   // <-- here the calls chain is f → g
}
```

lumosmind:
ahh I got it now g() can reach local variables of f(). Thank you

iliakan:
No, it can't (how could it?) ;)

They are "roots" meaning that they are retained in memory (not thrown away). So that after `g()` is done, the execution of `f()` can continue.

PR #2428, opened 2021-01-12
use property value shorthand to add as method

lumosmind:
We can also use **property value shorthand** concept for creating method.


iliakan:
Are you sure that such example is needed in that exactly place?

lumosmind:
The example uses  property value shorthand concept  which is discussed previous article and shows another way to declare a method. It is good for reinforcing learned concepts. And it shows we can manage functions like normal variables.

iliakan:
Yes, but here we concentrate on another topic. I try not to dilute the user attention.

lumosmind:
yes it makes sense. Subjective situation.

PR #2430, opened 2021-01-12
"exception" vs "new feature" ?

joaquinelio:
I think it is not an exception but a feature.

But there is another problem 
you define "module" as "just a file" and it was ok until now,  

but as I understand await won't work outside functions in the "main" script, so the change is not enough
should we add "inside imported modules" ? 

edit;
just realized I dont know shit
How do you access "top level code" inside modules?

### DIFF
--- a/1-js/11-async/08-async-await/article.md
+++ b/1-js/11-async/08-async-await/article.md
@@ -140,7 +140,7 @@ But we can wrap it into an anonymous async function, like this:
 })();
 ```
 
-P.S. The notable exception is V8 engine version 8.9+, where top-level await works in [modules](info:modules).
+P.S. New feature: starting from V8 engine version 8.9+, top-level await works in [modules](info:modules).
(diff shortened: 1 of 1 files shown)


iliakan:
The "main" script can be `<script type="module">`, then top-level async/await should work there.

What'sthe issue with accessing "top level code"?

joaquinelio:
mm
When I read your line "work in modules" I figured it all wrong, so forget it.

The suggestion "new way" instead  "exception" stands (not that important)
the rest is just me voicing my ignorance too soon.


TLDR, 
I thought the improvement was more flexibility inside (imported) modules 
and still not working on "main" module, it seemed ok to me. Then I realized the conceptual error.

I don't feel comfortable with it anyway.  
Can't say way. It doesnt sound "js way".  
Like the thread you never block in other languajes. Keep forking, never let it die.

PR #2435, opened 2021-01-13
Numeric Separators

lumosmind:
there is another way to write numbers more human readable with Numeric Separators.


lumosmind:
there is [caniuse page](https://caniuse.com/mdn-javascript_grammar_numeric_separators) for numeric separators
It works with decimal numbers , hexadecimals etc. but I didn't want to make article too complicate.

iliakan:
Nice, please see the updated article, is it good enough?

lumosmind:
yes it is very nice but simple typo in there. I open PR for that  #2437

iliakan:
Thanks!

PR #2436, opened 2021-01-13
more descriptive boundary value for Math.round

lumosmind:
3.5 is a boundary value for Math.round() so use 3.5 as example can be more descriptive.

### DIFF
--- a/1-js/05-data-types/02-number/article.md
+++ b/1-js/05-data-types/02-number/article.md
@@ -125,7 +125,7 @@ There are several built-in functions for rounding:
 : Rounds up: `3.1` becomes `4`, and `-1.1` becomes `-1`.
 
 `Math.round`
-: Rounds to the nearest integer: `3.1` becomes `3`, `3.6` becomes `4` and `-1.1` becomes `-1`.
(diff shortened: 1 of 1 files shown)


iliakan:
Sorry, but quite the opposite: 3.5 is the edge case, while 3.6 is more obvious.

iliakan:
We describe the main scenario here, not edge cases.

lumosmind:
but 3.5 shows that how does round() work. 3.6 does not give a clue about that.

iliakan:
Well, 3.2 shows rounding down, while 3.6 shows rounding up.

If you consider `3.5` an important edge case, we can add it.

lumosmind:
> Well, 3.2 shows rounding down, while 3.6 shows rounding up.
> 
> If you consider `3.5` an important edge case, we can add it.

yes it could be very efficient

iliakan:
Done:
```
Rounds to the nearest integer: `3.1` becomes `3`, `3.6` becomes `4`, the middle case: `3.5` rounds up to `4` too.
```

lumosmind:
thank you

PR #2438, opened 2021-01-13
example code has some problems

lumosmind:
example code has several problems
1 - it overuse the optional chaining 
user1?.[key]
user1 should be checked with different way not with optional chaining (the article says as that)

2 - example does not provide real reason to use square bracket notation

3 - example does not show nested object which is the real use case of optional chaining

I tried to create an example as a solution. My solution may be insufficient but current example must be changed I think.


iliakan:
Made minor fixes there.

The current example uses `[key]`, accessing a property by a variable name, stored in `key`.

That's common, opposite to `["property"]`. So it's important to use the variable here.

lumosmind:
I did my best actually. Maybe you can find another way to improve the example.

iliakan:
Thank you, I made it a little bit shorter and better I hope.

PR #2448, opened 2021-01-19
shorten if condition in line 14

tomasdelaveau:
Both "Cancel" and empty string ("") cases can be handled with `!value`, because it returns true in both cases (and false if input is `"0"`).


iliakan:
Let's be more explicit here. Better for people.

PR #2450, opened 2021-01-21
a more concise solution

tomasdelaveau:
I propose this alternative solution as it's shorter and it makes more use of destructuring assignment.


iliakan:
There's no need for more, more destructions here. Doesn't make the code better.

PR #2454, opened 2021-01-21
Update solution.md

shubhampaliwal:


iliakan:
You forgot about `use strict`, so it seems.

PR #2456, opened 2021-01-21
Add description that function in strict mode

emek:
It is not specified that the function is in strict mode. Added 'in strict mode' texts to prevent confusion.


lumosmind:
all example snippets and solutions in this project are written in strict mode, unless expressly provided otherwise.

PR #2459, opened 2021-01-22
Fix explanation of fetch's 'redirect: manual'

### LINKED ISSUE #2394: Fetch API: wrong information about `redirect`?
In the Fetch API page, there is a section about `redirect` (https://javascript.info/fetch-api#redirect):
https://github.com/javascript-tutorial/en.javascript.info/blob/a43c18ff18eb6a770d996b5ba699c872c6861fce/5-network/06-fetch-api/article.md#L166-L174

From my own understanding of the Fetch API, I think the details about how the `manual` value works are wrong: `response.url` doesn't hold the new URL and `response.redirected` is not `true` but `false`.

I think it's actually the Fetch API that is misleading in that regard, this has been reported here: https://github.com/whatwg/fetch/issues/763
And it's described as follows in the spec:
image
https://fetch.spec.whatwg.org/#requests

I'd be happy to update that but I'd like other point of views just to be sure I'm not missing something :)

ClementParis016:
Fixes #2394

Let me know if the explanation is good enough or not :)

Also I'm not sure what the process it to update that for other languages too?

### DIFF
--- a/5-network/06-fetch-api/article.md
+++ b/5-network/06-fetch-api/article.md
@@ -171,7 +171,7 @@ The `redirect` option allows to change that:
 
 - **`"follow"`** -- the default, follow HTTP-redirects,
 - **`"error"`** -- error in case of HTTP-redirect,
-- **`"manual"`** -- don't follow HTTP-redirect, but `response.url` will be the new URL, and `response.redirected` will be `true`, so that we can perform the redirect manually to the new URL (if needed).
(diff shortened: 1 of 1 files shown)


iliakan:
Oh. Please pity over people. Make it simpler!

What's an "opaque-redirect"? Not explained here (and don't make them read the spec please) =)

PR #2460, opened 2021-01-22
Add that both with or without new keyword causes same result

emek:
In 'new.target' section of "Constructor, operator 'new", I added that result of function call without new keyword is the same.


iliakan:
You added a single-line example that totally copied two lines above it.

Issue #2461, opened 2021-01-23
Alternatives in the example code of `1-js/09-classes/07-mixins` may be wrong


AmosChenYQ:
```javascript
let sayMixin = {
  say(phrase) {
    alert(phrase);
  }
};

let sayHiMixin = {
  __proto__: sayMixin, // (or we could use Object.create to set the prototype here) !!! this may be wrong

  sayHi() {
    // call parent method
    super.say(`Hello ${this.name}`); // (*)
  },
  sayBye() {
    super.say(`Bye ${this.name}`); // (*)
  }
};

class User {
  constructor(name) {
    this.name = name;
  }
}

// copy the methods
Object.assign(User.prototype, sayHiMixin);

// now User can say hi
new User("Dude").sayHi(); // Hello Dude!
```

Object.create can set __proto__ of sayHiMixin, but its method can be only function properties by
```javascript
let sayHiMixin = Object.create(sayMixin, {
    sayHi: {
        value: function() {
            super.say(`Hello ${this.name}`);
        },
        enumerable: true
    }
});
```
or 
```javascript
sayHiMixin.sayHi = function() {
 // ....
}
```
Calling super later will cause an error becasue both in these way there;s no [[HomeObject]], super keyword unexpected here.

iliakan:
Indeed. Merged!

PR #2462, opened 2021-01-23
modify alternative method in comment of mixins' code example

AmosChenYQ:
https://github.com/javascript-tutorial/en.javascript.info/issues/2461

Pls see the issues


iliakan:
A good observation, thanks!

PR #2464, opened 2021-01-25
Unify content with cosmetic changes

VibingCreator:


iliakan:
Could you please explain, why `try..catch` is bad? Is there an agreement out there for 3 dots?

Replacement of `e` with `err` seems ok.

VibingCreator:
There isn't any rule for using three dots, although it feels more natural after spending some time in MDN docs.

PR #2468, opened 2021-01-27
Fix Typo

Hasssan-Ramadan:
'single' -> 'song'


iliakan:
Is "single" really difficult to understand here and should be replaced?

Hasssan-Ramadan:
I'm not an English native speaker, I didn't know that "single" has the same meaning as "song". I thought it is a typo, so I opened this PR. I'm sorry for that.

iliakan:
I think, it's a valid point. I'd probably merge it.

iliakan:
Not only you may have trouble understanding this word.

Issue #2472, opened 2021-01-28
Spread Operator could be included in this article: https://javascript.info/object-copy


BM-BrunoMarques:
This article is great but I noticed that the spread operator is missing and could be introduced here as an alternative for object cloning.

PR #2473, opened 2021-01-28
Included Spread Operator in Objects Copy Article

BM-BrunoMarques:
When I was going through this article I felt that the spread operator could be included


BM-BrunoMarques:
I created this push to solve issue #2472

iliakan:
I added a note, still not sure it's needed though.

PR #2474, opened 2021-01-29
use conditional assignment in first solution

tomasdelaveau:
if we just do this.stomach = [food], for further calls of speedy.eat(food) the array would be re-assigned instead of having food pushed to it

### DIFF
--- a/1-js/08-prototypes/01-prototype-inheritance/4-hamster-proto/solution.md
+++ b/1-js/08-prototypes/01-prototype-inheritance/4-hamster-proto/solution.md
@@ -20,8 +20,8 @@ let hamster = {
 
   eat(food) {
 *!*
-    // assign to this.stomach instead of this.stomach.push
-    this.stomach = [food];
+    // conditional assignment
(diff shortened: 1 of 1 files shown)


1-js/08-prototypes/01-prototype-inheritance/4-hamster-proto/solution.md
```diff
@@ -20,8 +20,8 @@ let hamster = {
 
   eat(food) {
 *!*
-    // assign to this.stomach instead of this.stomach.push
-    this.stomach = [food];
+    // conditional assignment
+    this.stomach ? this.stomach = [food] : this.stomach.push(food);
```
MuhammedZakir, line 24: This won't work. `this.stomach.push(food)` won't be executed because property `stomach` is in the prototype chain. You have to use `hasOwnProperty` method instead.

```
this.hasOwnProperty('stomach') ? this.stomach.push(food) : this.stomach = [food];
```

iliakan:
No need to make things more complex.

PR #2478, opened 2021-01-30
fix text not corresponding to the code

seamissu:
Some text was still for the code example in Russian. (https://learn.javascript.ru/optional-chaining)
Here I just edited the text. Maybe changing code back to Russian version is another good option to fix the error.

### DIFF
--- a/1-js/04-object-basics/07-optional-chaining/article.md
+++ b/1-js/04-object-basics/07-optional-chaining/article.md
@@ -166,9 +166,9 @@ userGuest.admin?.(); // nothing (no such method)
 */!*
 ```
 
-Here, in both lines we first use the dot (`user1.admin`) to get `admin` property, because the user object must exist, so it's safe read from it.
+Here, in both lines we first use the dot notion `userAdmin.admin\userGuest.admin` to get `admin` property, because we assume the `userAdmin\userGuest` object must exist, so it's safe read from it.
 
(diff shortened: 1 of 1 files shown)


1-js/04-object-basics/07-optional-chaining/article.md
```diff
@@ -166,9 +166,9 @@ userGuest.admin?.(); // nothing (no such method)
 */!*
 ```
 
-Here, in both lines we first use the dot (`user1.admin`) to get `admin` property, because the user object must exist, so it's safe read from it.
+Here, in both lines we first use the dot notion `userAdmin.admin\userGuest.admin` to get `admin` property, because we assume the `userAdmin\userGuest` object must exist, so it's safe read from it.
```
MuhammedZakir, line 169: A suggestion:

```suggestion
Here, in both lines we first use the dot notation (`userAdmin.admin`) to get `admin` property, because we assume  `userAdmin` and `userGuest` objects must exist, so it's safe to read from it.
```

MuhammedZakir: We are mentioning `userAdmin.admin` after "dot notation" to remind users what dot notation is. So there is no need to mention `userGuest`.

iliakan:
Thanks, fixed user variable names.

PR #2481, opened 2021-02-01
there is not type conversion

lumosmind:
`7/0` does not require type conversion


iliakan:
Ok, removed, changed `task.md` too.

lumosmind:
did you delete all question? I can not see old question.

PR #2482, opened 2021-02-01
there are four logical operators in JS

lumosmind:
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator)


MuhammedZakir:
We also need to add a brief explanation about `??`.

lumosmind:
> We also need to add a brief explanation about `??`.

Yes but I couldn't decide to put brief explanation where and how. I just leave this job to @iliakan or somebody more experienced than me.

PR #2488, opened 2021-02-02
Re-exported symbols not available in current module

tredondo:
From https://developer.mozilla.org/en-US/docs/web/javascript/reference/statements/export#re-exporting_aggregating, "But where function1 and function2 do not become available inside the current module."


iliakan:
Thanks, added.

PR #2498, opened 2021-02-12
we can use label and break with simple JavaScript blocks and functions

lumosmind:
Changed line make me think "we can only use label and break/continue with loops". But this is not totally true. We can use label and break with simple JavaScript blocks and function declarations. So, I tried to make it more clear.

[MDN ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label#using_a_labeled_block_with_break)

### DIFF
--- a/1-js/02-first-steps/13-while-for/article.md
+++ b/1-js/02-first-steps/13-while-for/article.md
@@ -368,7 +368,7 @@ break label; // doesn't jumps to the label below
 label: for (...)
 ```
 
-If we work with labeled loops, a call to `break/continue` is only possible from inside the loop and the label must be somewhere above the directive.
+When working with a labelled loop, a call to `break/continue` is only possible from inside that loop and the label must be somewhere above the directive.
 ````
 
 ## Summary
--- a/1-js/02-first-steps/13-while-for/article.md
+++ b/1-js/02-first-steps/13-while-for/article.md
@@ -368,7 +368,7 @@ break label; // doesn't jumps to the label below
 label: for (...)
 ```
 
-A call to `break/continue` is only possible from inside a loop and the label must be somewhere above the directive.
+If we work with labeled loops, a call to `break/continue` is only possible from inside the loop and the label must be somewhere above the directive.
 ````
 
 ## Summary


1-js/02-first-steps/13-while-for/article.md
```diff
@@ -368,7 +368,7 @@ break label; // doesn't jumps to the label below
 label: for (...)
 ```
 
-A call to `break/continue` is only possible from inside a loop and the label must be somewhere above the directive.
+If we work with labeled loops, a call to `break/continue` is only possible from inside the loop and the label must be somewhere above the directive.
```
joaquinelio, line 371: I would agree
just a suggestion, I'm not native English speaker:

this article is about loops, I would follow the example context "Here the call..."

```suggestion
The call to `break/continue` is only possible from inside the loop block, and the label must be somewhere above the directive.
```

Maybe "label" should be mentioned inside "structure" article.

lumosmind: > I would agree
> just a suggestion, I'm not native English speaker:
> 
> this article is about loops, I would follow the example context "Here the call..."
> 
> Maybe "label" should be mentioned inside "structure" article.

I am not native English speaker too. But your suggestion is not clear enough. Because this concept is used very rarely. Yes, the article is about loops but this line can be more **explicit**. Or someone can add little extra explanation/warning about other usages of labels.

joaquinelio: I suggested to fix only the "must be inside a loop"

But I see your point, a comprehensive turorial should note 
"break can break any block not just loops" somewhere

MuhammedZakir, line 371:
```suggestion
A call to `break/continue` for a labelled loop is only possible from inside that loop and the label must be somewhere above the directive.
```

OR

```suggestion
When working with a labelled loop, a call to `break/continue` is only possible from inside that loop and the label must be somewhere above the directive.
```

iliakan:
@joaquinelio Wow, I never ever saw `break` used this way.

Maybe it's worth a better explanation?

iliakan:
Personally, that looks really weird.

If `break` is allowed to work with blocks, then why non-labelled break doesn't work?

This works:
```js
foo: {
  console.log('face');
  break foo;
  console.log('this will not be executed');
}
console.log('swap');
```

...But this doesn't (no label):

```js
{
  console.log('face');
  break;
  console.log('this will not be executed');
}
console.log('swap');
```

lumosmind:
> Imho, that looks really weird.
> 
> If `break` is allowed to work with blocks, then why non-labelled break doesn't work?
> 
> This works:
> 
> ```js
> foo: {
>   console.log('face');
>   break foo;
>   console.log('this will not be executed');
> }
> console.log('swap');
> ```
> 
> ...But this doesn't (no label):
> 
> ```js
> {
>   console.log('face');
>   break;
>   console.log('this will not be executed');
> }
> console.log('swap');
> ```

You are right that's my mistake but it does not effect the changed line. functions and blocks must be labeled when they contain break keyword. I just couldn't write good explanation as MDN.

iliakan:
Thanks for the note, I added some words about that.

iliakan:
...Without going much in-detail of that really weird feature =)

lumosmind:
> ...Without going much in-detail of that really weird feature =)

yes really weird :) it feels like `goto` keyword.

joaquinelio:
I guess more a side effect than a "feature"
they didn't bother

joaquinelio:
it's not weird, it is beautiful.  
Classical break feels like goto.  
The concept is cleaner and universal:  **You label any block you are breaking out.**  (*)

we can forget the "goto but no goto" thing - and its warnings.
 
...
(*) Is grammar ok?
break jail, like the tool inside the cake

joaquinelio:
not weird but cool

```
invoice: {

    for invoiceItems {

        processItem 
        if inPain   break invoice     // truly
    } 
    processPayment     // can break invoice here too
    printing 
    etc

} 
```
pretty clean

PR #2499, opened 2021-02-13
Fixed grammatical typo

neeraj1bh:


1-js/04-object-basics/06-constructor-new/article.md
```diff
@@ -185,7 +185,7 @@ Omitting parentheses here is not considered a "good style", but the syntax is pe
 
 Using constructor functions to create objects gives a great deal of flexibility. The constructor function may have parameters that define how to construct the object, and what to put in it.
 
-Of course, we can add to `this` not only properties, but methods as well.
+Of course, we can add `this` to, not only properties, but methods as well.
```
joaquinelio, line 188: but we add the stuff "to this" 
just asking, not native speaker, I dont know where to put "as well"

```suggestion
Of course, we can add not only properties, but methods as well, to `this` .
```
?

joaquinelio, line 188: but we add the stuff "to this"
just asking, not native speaker 

```suggestion
Of course, we can add not only properties, but also methods to `this`.
```

neeraj1bh: Yes the edit you made would work fine as well.

neeraj1bh: Fixed grammatical typo

joaquinelio:
I mean it's a matter of meaning, 
In spanish is ok to say *"add to this the props" meaning "add the props to this"   

here we don¿t add "this" to the property,  
we do add the property  to "this"

neeraj1bh:
Yes you are right but in English we write it as " we can add `this` to "

MuhammedZakir:
Apologies, but like @joaquinelio said, yours is wrong. "Add a to b" and "add b to a" are totally two different things. Here, we are adding properties/methods to `this` and not adding `this` to properties.

joaquinelio:
If my grammar is ok, I stand on my first suggestion...
stand with ?
uf, grammar

I mean I think my suggestion is the way to go,  drop it if flawed.
Agreed,  original grammar is wrong 
but your suggested fix changes the original meaning

iliakan:
The English seems wrong in the PR. I barely can understand that.

joaquinelio:
"...we can also add methods to `this`."
cleaner

PR #2502, opened 2021-02-20
What local really means

joaquinelio:
I think you should add this as a characteristic, here or somewhere else, your own words if you don't like this.
But I think this is important.
 
Maybe It has an implicit warning about the lack of backup option, (if you want it you have to implement it) but it s a digression maybe already covered with the mention of the serviceworker. 
Personally, If I ever use indexeddb, I will do it only as an offline cache

### DIFF
--- a/6-data-storage/03-indexeddb/article.md
+++ b/6-data-storage/03-indexeddb/article.md
@@ -18,6 +18,8 @@ The native interface to IndexedDB, described in the specification <https://www.w
 
 We can also use `async/await` with the help of a promise-based wrapper, like <https://github.com/jakearchibald/idb>. That's pretty convenient, but the wrapper is not perfect, it can't replace events for all cases. So we'll start with events, and then, after we gain an understanding of IndexedDb, we'll use the wrapper.
 
+Please note the data stored is browser/user profile/site dependent. The local data is not meant to be shared among local users nor browser vendors. If the user switches to another browser, it will work on its own database.
+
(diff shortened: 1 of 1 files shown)


iliakan:
This is a common thing to all storage options: cookies, localstorage, indexeddb.

If we are to explicitly mention this, then we should do it elsewhere.

P.S. It's probably obvious that different browsers don't share the storage. They're just different programs. And incognito is, well, incognito...

joaquinelio:
ok...  I'm a maniac (that helped me sometimes tough)

But I don't see it that obvious

This db is so powerful and can store so much data, I would expect more robustness after I close the tab.   
we know how browsers work and treat local resources,
it's not the browser's responsibility but developer's...  so the warning. 

TLDR
beginner's eyes.
after a beginner overwriting a customer's Access db with his update
I asked him: "so what did you learn"
(angry with himself): "I have to pay more attention and check every step...!"
"so you learn nothing. Backups! "
so...

sometimes I see through the user's eyes too, I would relate the data to the site, not the program.
again, if dev does a good job, the user shouldn't care.

iliakan:
I added this:
```
Technically, the data is usually stored in the visitor's home directory, along with browser settings, extensions, etc.

Different browsers and OS-level users have each their own independant storage.
```

joaquinelio:
I like it!
You solved your way, with straight technical info instead of empty warnings, it's perfect.

PR #2508, opened 2021-03-04
Add clarity to number data type rounding examples

Brennvo:
This adds some clarity and expected behavior in the rounding examples supplied in the number data type in section 5.2

When using `Math.floor` in the multiply-and-divide example, the result could potentially be different than that of the native `toFixed` method on the `Number` object. For instance, if I provided a number that would round up with `toFixed` method (such as rounding to two decimal places in 1.2385), it would not round up in the current implementation of the multiply-and-divide example. I went about this by using `Math.round` instead of `Math.floor`. I also decided to update example 1 to round up, just to further clarify that the multiply-and-divide behaves similarly to the `toFixed` method in example 2.

Additionally, with the update to  example 1, I also went ahead and updated  example 2 to be in sync, as well as provide an example for rounding down since the initial example will now round up.

### DIFF
--- a/1-js/05-data-types/02-number/article.md
+++ b/1-js/05-data-types/02-number/article.md
@@ -150,7 +150,7 @@ Here's the table to summarize the differences between them:
 
 These functions cover all of the possible ways to deal with the decimal part of a number. But what if we'd like to round the number to `n-th` digit after the decimal?
 
-For instance, we have `1.2345` and want to round it to 2 digits, getting only `1.23`.
+For instance, we have `1.2385` and want to round it to 2 digits, getting only `1.24`.
 
 There are two ways to do so:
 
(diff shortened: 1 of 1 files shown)


iliakan:
This PR breaks a nice explanation of basic `Math` function. 

Even I have difficulty understanding what you're trying to explain. Really don't get it, sorry.

And yes, `toFixed` rounds like `Math.round`, we can add this info if it's not there.

Brennvo:
> Even I have difficulty understanding what you're trying to explain. Really don't get it, sorry.

I am addressing the two options supplies for the following text under [the rounding section of 5.2](https://javascript.info/number#rounding):

> But what if we'd like to round the number to `n-th` digit after the decimal?

The text lists two approaches for doing so:

1. Multiply-and-divide.

2. `toFixed(n)`

The formula currently used in multiply-and-divide does not match the behavior of `toFixed(n)`. For example, instead of rounding `1.2345` to two decimals (which is being used in the text), let's use `1.2385`. I would expect this number to round to `1.24`. However, if we do this under both the options listed:

Multiply-and-divide -> `Math.floor(1.2385 * 100) / 100` -> ` 1.23`

`toFixed(n)` -> `1.2385.toFixed(2)` -> ` "1.24"`

Data types aside, they compute differently. I am suggesting that the multiply-and-divide formula use `Math.round(n)` in its formula instead. If we do that:

`Math.round(1.2385 * 100) / 100` -> `1.24`

iliakan:
Sounds good, done, thanks!

PR #2512, opened 2021-03-08
class syntax vs object literal

lumosmind:
> To tell the truth, the notations are not fully identical. There are subtle differences related to object inheritance (to be covered later), but for now they do not matter. In almost all cases the shorter syntax is preferred.

I think this paragraph talks about prototypal inheritance but when we work with object literal there is no difference between  `sayHi: function() {` and `sayHi() {`  booth syntaxes add property as own property of object itself, does not add property to object's `prototype` like es6 class syntax.

### DIFF
--- a/1-js/04-object-basics/04-object-methods/article.md
+++ b/1-js/04-object-basics/04-object-methods/article.md
@@ -90,8 +90,6 @@ user = {
 
 As demonstrated, we can omit `"function"` and just write `sayHi()`.
 
-To tell the truth, the notations are not fully identical. There are subtle differences related to object inheritance (to be covered later), but for now they do not matter. In almost all cases the shorter syntax is preferred.
-
 ## "this" in methods
 
 It's common that an object method needs to access the information stored in the object to do its job.


iliakan:
What is `user.__proto__` is changed? Then, `super` keyword inside these two notations behaves differently.

lumosmind:
I couldn't understand. Doesn't `super` keyword belong to es6 class syntax? But These paragraph is about object literal. What do I miss?
Please throw some example snippets. I am very confused.

iliakan:
Sure, here you are.

This works (shows 1)
```js
let user = {
  f() { 
    super.f();
  }
};

Object.setPrototypeOf(user, {
  f() { alert(1) }
});

user.f();
```

This doesn't (shows error):
```js
let user = {
  f: function() { 
    super.f();
  }
};

Object.setPrototypeOf(user, {
  f() { alert(1) }
});

user.f();
```

iliakan:
Closing it meanwhile, let me know if should be open.

lumosmind:
thanks for examples I will mediate on them.

PR #2514, opened 2021-03-11
Class setter/getter

zxymike93:
At [class setter/getter sample code](https://javascript.info/class#getters-setters)

It seems that setter/getter is referencing a "private" property.


iliakan:
No mistake.

Sebastianokunkelo:
C:\Users\fyfloki\Documents\GitHub\vscode-eslint\package.json

PR #2515, opened 2021-03-11
minor suggestions

joaquinelio:


joaquinelio:
another thing

Here we talk about searches
1) "by key" and 
2) "by other fields" using ad-hoc indexes... 
but here we use the word "key" too. Shouldn't we talk about master/main/primary keys in (1)? With maybe a mention regarding the mandatory "unique" prop.

iliakan:
Why should we replace self-descriptive normal-sized links with names? ;)

iliakan:
Integrated few other changes, closing the PR, as totally avoiding links isn't justified.

joaquinelio:
got it.  you put it on purpose
I thought link looked  too large

About spec links, I believe you used it that way in many places, looks cleaner.

joaquinelio:
@iliakan 
There are other "minor" things 

"...TWO MAIN types of searches..."
472 ## Searching by keys
525 ## Searching by any field with an index

but "by field" is"by key" too, it's a real key, over an index.
the word key is used in "keypath" and "index is ordered by key"
should we use "primary/main/ " ( store key, or something)  to avoid ambiguity?

iliakan:
Here, "by key" is "by key of the collection", the difference versus fields looks obvious, no?

In indexeddb, unlike SQL, there's a single obligatory primary key, the word "key" means exactly that...

Is there anything else to fix? Maybe I'm missing something.

joaquinelio:
The diff may be obvious.  But it doesn't sound good to me.

a) by key
b) by field via index
 then you mention here keypath, key for index, array key

Could be "by field via index key"?  so key is clearly "main" and "index key" sounds different then keypath doesn't shock. 
Don't pay me attention, it's ok.

---

136, silly change 
"... it closes the old connection"
more like
"... closes this connection because it got old / was notified of being old kill yourself" pffffff, not like this.
not that important, at this point is clear how it works


I love this article, 
it should be the "fundamentals" reading for any db

iliakan:
Added more explanations about keys and more headers, should be enough now.

136 looks good to me, submit a PR if you feel otherwise =)

Thanks a lot for making the tutorial much much better!

joaquinelio:
> 136 looks good to me, submit a PR if you feel otherwise =)

subtle:
"old connection" vs "THIs connection", but it is old. maybe "this old connection" : )  that's ugly, sorry.

> Thanks a lot for making the tutorial much much better!

yes, I know.
that's easy, I just complain and you do the work.
oh... my ENG repo ranking should be higher...  You plagiarist!  XD

iliakan:
PR, please?

PR #2519, opened 2021-03-13
rephrase

joaquinelio:


joaquinelio:
ha ha !
cant explain why

I wanted "closes the conn that became old" 
it IS old, but I erased that too

I think it reads better, 
and it is still evident,  event -> closes the connection, 
the reader goes to code, it reads "closed  because obsolete, I think it's ok

joaquinelio:
oops you just did it again

it s ok

iliakan:
updated

iliakan:
hopefully better now

PR #2525, opened 2021-03-17
locate vs add?

joaquinelio:
more a "shows" or locate/allocate/put than add


iliakan:
maybe puts? not locates

joaquinelio:
from the dictionary  "locate" is "to put" but also "find" 

so ok, put is better

PR #2533, opened 2021-03-19
No polyfill needed any longer

joaquinelio:
should we keep polyfill link?
or just drop it


iliakan:
We can drop it now, or just later.

PR #2535, opened 2021-03-21
:memo: Added links for TODO documentation tools

konsalex:
I added links for TODO documentation tools like JSdoc `@todo` param and Better Comments for VScode.

Those tools feel essential for new large teams.


iliakan:
AFAIC, Imdone isn't essential for large teams. 
VS Code is also not essential.

I'm not sure this is relevant.

Also, why do you replace *italic* with _underline_? We never use the underline in the tutorial.

PR #2543, opened 2021-03-24
one more declaration in 06-advanced-functions/07-new-function/article.md

LLyaudet:


iliakan:
There's no such thing in the spec: https://tc39.es/ecma262/#sec-createdynamicfunction

I believe, that's because an array is converted to String.

PR #2555, opened 2021-03-28
transpiler def

joaquinelio:
*Not English native speaker*

"transpiler, unlike compilers, translate to still high-level code..."  would be a more accurate definition  
is not only for outdated stuff but to add functionality they don't plan to add 

but maybe the shorter the better.

### DIFF
--- a/1-js/03-code-quality/06-polyfills/article.md
+++ b/1-js/03-code-quality/06-polyfills/article.md
@@ -22,7 +22,7 @@ Here, in this chapter, our purpose is to get the gist of how they work, and thei
 
 ## Transpilers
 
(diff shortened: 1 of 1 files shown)


lawki:
> result would be the same

This, the older text, makes me question what would be the role of transpilation then?

>target engine (even oudated ones) can run

This makes more sense and overall improves on existing text. But I feel like removing `so, ` makes it harder to understand.

joaquinelio:
My complain was about transpiler definition.
I don like my version either, it makes it harder to understand. 

Im changing it.

lawki:
I think the updated text is a good balance between skipping over crucial details and being too complicated.

PR #2562, opened 2021-04-01
Grammar Fix Update in article.md

PhilipKirner:
Potential Grammar Correction


PhilipKirner:
For instance, in `childNodes` we can see both text nodes, element nodes, and even comment nodes if **there** exist.
To
For instance, in `childNodes` we can see both text nodes, element nodes, and even comment nodes if **they** exist.

iliakan:
Thanks!

PR #2564, opened 2021-04-03
domain cookie option

joaquinelio:
line 111
"no way to let a cookie be accessible from another 2nd-level domain"
you mean
"no way to let a cookie be accessible from another domain" ??

if so, I would also rewrite the "domain" topic something like this.

Moved the most important lines to the top  "other.com never access to..."  
once stated that, no need to refer to "other site" anymore 
and can focus on subdomain tricky part

### DIFF
--- a/6-data-storage/01-cookie/article.md
+++ b/6-data-storage/01-cookie/article.md
@@ -98,7 +98,7 @@ A domain defines where the cookie is accessible. In practice though, there are l
 
 **There's no way to let a cookie be accessible from another domain, so `other.com` will never receive a cookie set at `site.com`.**
 
-It's a safety restriction, to allow us to store sensitive data in cookies, that should be available only on one site.
+It's a safety restriction, to allow us to store sensitive data in cookies that should be available only on one site.
 
 By default, a cookie is accessible only at the domain level that set it.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Is it really better ?

joaquinelio:
I think so.  git shows more diff than actually are, though.

step by step, you may agree with some:

1) line 111 is wrong: 
**There's no way to let a cookie be accessible from another 2nd-level domain, so `other.com` will never receive a cookie set at `site.com`.**
to
**There's no way to let a cookie be accessible from another XXXXXXX domain, so `other.com` will never receive a cookie set at `site.com`.**

2) moved that line (and related line 113)  to the top (makes sense)

3) then we can simplify former line 101
...But what's more tricky, we also won't get the cookie at a subdomain `forum.site.com`!
to
...What's tricky, we won't get the cookie at a subdomain `forum.site.com`!

joaquinelio:
hey, why is this hanging. 
No big deal.

The part that troubled me 
>111 **There's no way to let a cookie be accessible from another 2nd-level domain, so

I just deleted "2nd-level", if I'm wrong I can close it.

iliakan:
Well, the thing is: a cookie set by `a.site.com` can be accessible by `b.site.com` -- a different domain.

iliakan:
Thanks, let's merge this one! ;)

joaquinelio:
Right!
Feeling fool -again-

1, I didnt realize "com" is 1st level domain.  May others have same mistake when reading the first time?
2, PR is a minor change now. I prefer: What it does, what it cant do, THEN  the tricky part.

oops
merged while  writing this comment
nice, thanks

Issue #1961, opened 2020-06-16
optional chaining


dagolinuxoid:
> The optional chaining ?. is not an operator, but a special syntax construct, that also works with functions and square brackets.

But according to MDN it's an operator. 

> The optional chaining operator (?.) permits reading the value of a property located deep within a chain of connected objects without having to expressly validate that each reference in the chain is valid. The ?. operator functions similarly to the . chaining operator, except that instead of causing an error if a reference is nullish (null or undefined), the expression short-circuits with a return value of undefined. When used with function calls, it returns undefined if the given function does not exist.

Are there any JS spec evangelists who can claim precisely what the *thing* is ? =)

iliakan:
Read the spec, ok? And tell us! I have my opinion, but mistakes are always possible.

dagolinuxoid:
Folks at *tc39* refer to it as an operator as well.

> The Optional Chaining operator is spelled ?.

https://github.com/tc39/proposal-optional-chaining#syntax

iliakan:
It's the proposal, not the spec.

iliakan:
You know where to find the spec, right? ;)

dagolinuxoid:
sure :)

This specification, the 11th edition, introduces adding two new syntax features to improve working with “nullish” values (null or undefined): nullish coalescing, a value selection operator; and optional chaining, a property access and function invocation operator that short-circuits if the value to access/invoke is nullish.

It's a bit confusing and I'm not 100% convinced but to me it seems like an operator ( now I'm regretting that I opened the issue and struggling with the spec rather than spending time to build react app ) 
btw. congrats upon reaching 10k stars!

iliakan:
@dagolinuxoid maybe ask TC39 directly? In their issues or alike?

iliakan:
@dagolinuxoid they might just answer and save our time =)

iliakan:
Here: https://es.discourse.group/t/optional-chaining-the-syntax-or-the-operator/366

dagolinuxoid:
You did the work for me. And I glad you did it. You asked right questions even though yet it turns out there is no 'strict right' answer to point that the `?.` either an operator or not.  Your term syntax is more general so I'm convinced it's appropriate to use, but probably the strong claiming that `?.` it's not an operator is a bit reckless I would say.

iliakan:
[...]

PR #2567, opened 2021-04-05
[discussion] The optional chaining `?.` is an operator

lumosmind:
reference : [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

### DIFF
--- a/1-js/04-object-basics/07-optional-chaining/article.md
+++ b/1-js/04-object-basics/07-optional-chaining/article.md
@@ -142,7 +142,7 @@ alert(x); // 0, value not incremented
 
 ## Other variants: ?.(), ?.[]
 
-The optional chaining `?.` is not an operator, but a special syntax construct, that also works with functions and square brackets.
+The optional chaining `?.` is an operator, but also works with functions and square brackets.
 
 For example, `?.()` is used to call a function that may not exist.
 


joaquinelio:
!!
"not a math operator"? no, no need.
 
The article should mention it has a precedence value too.

lumosmind:
> !!
> interesting.
> The article should mention it has a precedence value too.

actually I realized ?? is an operator when I study on operator precedence table

joaquinelio:
> > !!
> > interesting.
> > The article should mention it has a precedence value too.
> 
> actually I realized ?? is an operator when I study on operator precedence table

yes, that's what convinced me, more than mdn.

iliakan:
?. isn't an operator. Discussed many times.

joaquinelio:
oh...  
#1961 

Then 
1 "despite what MDN stubbornly says many times right now, it is not an operator"

2 we (you) should explain why it has an "operator precedence" value
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table

eppur si muove
the operands are blocks,

iliakan:
@joaquinelio please take a look at the spec as the main source of truth: https://tc39.es/ecma262/

This is the ONLY source of truth. 

MSN just isn't. MDN is a compilation for people who want things simpler. Sometimes MDN authors have their own interpretation, that may be compatible with the spec, and we can even pick it up as well.

About `?.` specifically, here's the discussion with one of the spec authors: https://es.discourse.group/t/optional-chaining-the-syntax-or-the-operator/366/2

Personally, I take words from a spec author as a higher authority than MDN.

Also, we don't call the dot as in "user.name" a "member access operator", like MDN does. There's no such operator in the spec, it's something invented in MDN. So it makes little sense to call `?.` an operator, as it's a "modifed" dot.

joaquinelio:
ok, your kid.
still the best tutorial.

as this doesn't have a "discussion page" like wiki, 
a small note, maybe at the end of the article, would prevent recurring discussions...

iliakan:
Leaving it open for discussions, if there are other opinions.

joaquinelio:
dot as ugly "member access operator" is strong. It IS better for the tutorial to consider it as syntax.
still, why "operator precedence" value

iliakan:
Please, note that the "operator precedence" table is an MDN invention. There's nothing like that in the spec. 

So they give numbers and change them, as they like. 

That's their interpretation of the spec, and I guess it's good for people. Makes it easy to understand why some operator works before another.

We also use this table in the tutorial, when talking about operators. But a dot isn't an operator, at least in JS. Likewise, the `?.` also isn't.

joaquinelio:
ah, that's why I couldn't find it in the spec

NOW we agree.
I guess...  😄

PR #2569, opened 2021-04-06
it also work as Map value

lumosmind:
```
let john = { name: "John" };

let map = new Map();
//map.set(john, "...");
map.set("x",john);

john = null;
map.keys().next().value
```

### DIFF
--- a/1-js/05-data-types/08-weakmap-weakset/article.md
+++ b/1-js/05-data-types/08-weakmap-weakset/article.md
@@ -36,7 +36,7 @@ john = null; // overwrite the reference
 */!*
 ```
 
-Similar to that, if we use an object as the key in a regular `Map`, then while the `Map` exists, that object exists as well. It occupies memory and may not be garbage collected.
+Similar to that, if we use an object as the key or the value in a regular `Map`, then while the `Map` exists, that object exists as well. It occupies memory and may not be garbage collected.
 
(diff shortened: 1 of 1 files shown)


joaquinelio:
not sure...   
Right, the sentence may be more accurate I guess

But 
it is not exactly wrong without
it should be obvious at that moment (same as array item I guess)
and 
it is trying to do a point about "key", 
the object as a value is not going to change anything, it feels... a stowaway

lumosmind:
> not sure...
> Right, the sentence may be more accurate I guess
> 
> But
> it is not exactly wrong without
> it should be obvious at that moment (same as array item I guess)
> and
> it is trying to do a point about "key",
> the object as a value is not going to change anything, it feels... a stowaway

this commit aim at completeness. the subject is garbage collector behavior so  object as a value is absolutely going to change something.

joaquinelio:
Got it, garbage collection is the important point here
the author put it only as the introduction to weakmap 

If you add "or value"  you should add the consequence:  the item stays. 
As expected, so I wouldn't add it.  (the item & the object stay, map or weakmap)  

and following 
"  `WeakMap` is fundamentally different in this aspect."  seems odd now
you'll need to add you are referring to key only

iliakan:
The main point here is about keys. Values can be anything.

PR #2570, opened 2021-04-06
It can be more clear to read

lumosmind:
nested function call is make this snippet hard to understand. And array destructuring is not mentioned yet. So it can be hard for beginners.

### DIFF
--- a/1-js/05-data-types/09-keys-values-entries/article.md
+++ b/1-js/05-data-types/09-keys-values-entries/article.md
@@ -88,12 +88,10 @@ let prices = {
   orange: 2,
   meat: 4,
 };
-
 *!*
-let doublePrices = Object.fromEntries(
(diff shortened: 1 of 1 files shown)


joaquinelio:
mm...
Agreed it may be too early, but the chaining is a point

lumosmind:
> mm...
> Agreed it may be too early for a beginner, but here the chaining is a point too.

chaining and nesting different things. Chaining is protected if it is the point.

joaquinelio:
ouch! right. 
I meant the line below
"...We can make powerful chains of transforms this way. "

maybe both? 

...your code 
add "better yet"
...original code
so "...We can make powerful chains of transforms this way. " makes sense. I like being introduced to this here.

iliakan:
There are detailed explanations, the reader may need to wrap their head about it, but that's the point.

PR #2578, opened 2021-04-12
store vs keeps stored

joaquinelio:


joaquinelio:
"to store" data,  vs  
"to keep" it there while reachable

Reinforce "reachable" quoted or another remark, the word is key to garbage collection 

I think it is more clear and make it easier to follow the next lines

PR #2580, opened 2021-04-14
Mention scrollX and scrollY

paroche:
They are how pageXOffset and pageYOffset are listed in MDN. For completeness, and so a search for one of those terms will bring up this page.

### DIFF
--- a/2-ui/1-document/10-size-and-scroll-window/article.md
+++ b/2-ui/1-document/10-size-and-scroll-window/article.md
@@ -73,7 +73,7 @@ alert('Current scroll from the left: ' + window.pageXOffset);
 
 These properties are read-only.
 
-(note: `pageXOffset` and `pageYOffset` are aliases for `scrollX` and `scrollY`, but provide greater cross-browser compatibility.)
+(note: `pageXOffset` and `pageYOffset` are aliases for `scrollX` and `scrollY`, but may provide somewhat greater cross-browser compatibility.)
 
 ## Scrolling: scrollTo, scrollBy, scrollIntoView [#window-scroll]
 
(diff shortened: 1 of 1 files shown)


iliakan:
Can we safely say that `scrollY` is an alias of `pageYOffset`? (IE9 doesn't exist for us, IE11 is the most recent version we may take into account).

paroche:
I based my phrasing on the MDN Web Docs:
[https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollY](url)
[https://developer.mozilla.org/en-US/docs/Web/API/Window/pageYOffset](url)

I figured that the term "alias" was probably being used there in a formal way, so that `pageYOffset` was technically an alias for `scrollY`, not the other way around. But I don't know for sure.

`caniuseit.com` also describes `pageYOffset` as "non-standard":
[https://caniuse.com/?search=scrolly](url)

(even though it's apparently the only syntax recognized by IE11 (and IE9-IE10)), and, based on a cursory examination of the site, seems to be recognized everywhere `scrollY` is recognized).

iliakan:
The spec https://drafts.csswg.org/cssom-view/#dom-window-scrolly mentions both.

It also says "The pageYOffset attribute must return the value returned by the scrollY attribute."

Both attributes are read-only.

It makes them literally aliases, or do I miss something?

paroche:
It seems to be listing `scrollY` as primary -- it's defined as having to return a certain value, whereas `pageYOffset` is defined in terms of `scrollY`: 

"_The scrollY attribute must return the y-coordinate, relative to the initial containing block origin, of the top of the viewport, or zero if there is no viewport._

_The pageYOffset attribute must return the value returned by the scrollY attribute._"

So it looks to me like `scrollY` is the "real" name and `pageYOffset` is the "alias". 

(Typically, an "alias" is not a real name: [https://www.merriam-webster.com/dictionary/alias](url))

If that still doesn't sit well, I think the main thing is just to mention `scrollY` and `scrollX` so people looking for those functions will be able to find them (and `pageYOffset` et al).

iliakan:
Your suggestion is great, just exploring options how to put it better.

iliakan:
Added, thanks!

paroche:
Looks good.

PR #2584, opened 2021-04-18
typos? easier reading

joaquinelio:
review pls


iliakan:
Sure. Thanks for fixing these. I definitely need to put more articles.

PR #2589, opened 2021-04-23
Update article.md

Jiapan-Yu:
should be 'iterable' precisely speaking


iliakan:
it was correct. it's an example. 

I removed confusing comments.

PR #2591, opened 2021-04-23
Update article.md

GulfamAnsari:
The iteration goes in the same order as the values were inserted in the regular Object as well. Check the below code snippet. The iteration order in regular will be the same like map has.
```
var obj = {};
obj['a'] = 'a';
obj['b'] = 'b';
obj['d'] = 'd';
obj['c'] = 'c';
for ( var t in obj) {
  console.log(t)
}
```

### DIFF
--- a/1-js/05-data-types/07-map-set/article.md
+++ b/1-js/05-data-types/07-map-set/article.md
@@ -134,9 +134,6 @@ for (let entry of recipeMap) { // the same as of recipeMap.entries()
 }
 ```
 
-```smart header="The insertion order is used"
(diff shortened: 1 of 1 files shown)


vsemozhetbyt:
Try this:

```js
var obj = {};

obj['a'] = 'a';
obj['b'] = 'b';

obj['1'] = '1';
obj['2'] = '2';

for ( var t in obj) {
  console.log(t)
}
```

GulfamAnsari:
Ohh yes.

Thanks.

PR #2596, opened 2021-05-01
Move section about undefined this inside modules

leo-diehl:
Move the section about "undefined global this" within the "inbuilt strict mode" one.
This characteristic is not specific to modules but related to the inbuilt use-strict that modules have,
and the way the information was placed could mislead the user to think that it was module-exclusive.

### DIFF
--- a/1-js/13-modules/01-modules-intro/article.md
+++ b/1-js/13-modules/01-modules-intro/article.md
@@ -77,6 +77,19 @@ Modules always `use strict`, by default. E.g. assigning to an undeclared variabl
 </script>
 ```
 
+Because of this inbuilt rule, it's useful to mention that the top-level `this` will be undefined.
+
+Compare it to non-module scripts, where `this` is a global object:
+
+```html run height=0
+<script>
+  alert(this); // window
+</script>
+<script type="module">
+  alert(this); // undefined
+</script>
+```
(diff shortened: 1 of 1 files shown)


joaquinelio:
I wouldnt delete it...
maybe rephrase it  adding the why 

the point is the difference,  window obj  (even with strict) vs undefined.

I guess no-strict outside the browser it should return system obj

iliakan:
@leo-diehl please elaborate.

Looks like you think that 'use strict' implies undefined `this`. It's not so. Modules do that.

Please check this example: http://plnkr.co/edit/caiRRgRbfqsc84Yr?open=lib%2Fscript.js

leo-diehl:
@iliakan, thanks for the code example, it gave me an insight that I didn't have had before: that the functions created inside a strict environment have their `this` undefined, not the environment itself.

I would like to keep my suggestion, please check this example:
http://plnkr.co/edit/4XvG7xorHd3CUf8T?open=lib%2Fscript.js&preview

iliakan:
These are different rules.

1. A function in `use strict` mode has `this=undefined`.
2. Modules in global scope has `this=undefined`.
3.

leo-diehl:
Thanks for the clarification!

PR #2599, opened 2021-05-11
fix(classes): replace [[FunctionKind]] with [[IsClassConstructor]]

H2rmone:
The latest ECMAScript Language Specification has replace `[[FunctionKind]]` with `[[IsClassConstructor]]` in favor of indicating whether the function is a class constructor.


iliakan:
Yeah, right =)

PR #2613, opened 2021-06-02
ECMAScript Engine for Safari is also known as "JavaScriptCore"

Suraj7879:
https://en.wikipedia.org/wiki/List_of_ECMAScript_engines


iliakan:
Come on, we can't list all names :(

PR #2615, opened 2021-06-04
not a handler -pls check gramm

joaquinelio:
"DOMContentLoaded handler" suggests "handler named DOMContentLoaded".
Is allowed  "DOMContentLoaded's handler " ?
maybe  " the handler of DOMContentLoaded "

### DIFF
--- a/2-ui/5-loading/01-onload-ondomcontentloaded/article.md
+++ b/2-ui/5-loading/01-onload-ondomcontentloaded/article.md
@@ -45,7 +45,7 @@ For instance:
 <img id="img" src="https://en.js.cx/clipart/train.gif?speed=1&cache=0">
 ```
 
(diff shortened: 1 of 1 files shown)


iliakan:
It's the handler of the event "DOMContentLoaded". Looks good to me, as it is now.

joaquinelio:
mmm... ok...

48 ...  the `DOMContentLoaded` handler runs ...
52  ... the `DOMContentLoaded` event is very simple,

it lead to a wrong translation in sp.
but I dont think anybody will get it wrong

iliakan:
The long phrase would be: "the handler of the DOMContentLoaded event".

joaquinelio:
yes
loooong,
mm.
sounds more precise. I'll fix sp again. Hope it doesnt distract readers

PR #2620, opened 2021-06-07
Optimization solution of hamster-proto

leviding:
When feeding multiple times, keep the food that has been fed:

```js
eat(food) {
- this.stomach = [food];
+ this.stomach = [...this.stomach,food];
}
```


iliakan:
Thanks, may be more correct, but kind of diverts attention from important stuff =) 

I'd better not make this code even a tiny bit more complex here.

leviding:
Thanks!

PR #2629, opened 2021-06-13
small update to 1-js/04-object-basics/09-object-toprimitive

ZYinMD:


iliakan:
The article was updated, please see now (from the beginning)

iliakan:
Good idea, to specify that clearly.

PR #2631, opened 2021-06-14
fix bug in a task solution in weakmap-weakset

ZYinMD:
Hi, in my humble opinion a global symbol should be used here, because when we check the isRead status for all messages, they should exist under the same secret key.

I could be totally wrong because I'm not very good at symbols, feel free to reject.


iliakan:
Only if the symbol must be app-scoped, not always the case. 
Usually, such symbols are module-scoped.

PR #2632, opened 2021-06-14
add one sentence in 1-js/04-object-basics/06-constructor-new

ZYinMD:
Add explanation to some magic.

### DIFF
--- a/1-js/04-object-basics/06-constructor-new/article.md
+++ b/1-js/04-object-basics/06-constructor-new/article.md
@@ -81,6 +81,8 @@ let user = new function() {
 ```
 
 This constructor can't be called again, because it is not saved anywhere, just created and called. So this trick aims to encapsulate the code that constructs the single object, without future reuse.
+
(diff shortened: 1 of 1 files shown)


iliakan:
Why is this logically needed? A total surprise.

ZYinMD:
Because logically the `new` keyword should be followed by a function **_call_**. For instance `let now = new Date()`. It is indeed valid to simply write `let now = new Date`, which works fine, but that's the magic of the `new` keyword, it doesn't mean it's logically correct. If we want to take advantage of the magic, we should at least mention it is magic.

If you use prettier, and write something like 
```js
let foo = new function() {
  //...
}
```
then prettier will automatically change it to 
```js
let foo = new (function () {
  //...
})();
```

iliakan:
That will be more confusing to people than the existing variant.

It's clear enough what's going on in a `new function() { ... }`. I added a comment there to show it.

PR #2633, opened 2021-06-16
"inserts into" instead of "appends to"

joaquinelio:


joaquinelio:
The other option  "appends to the document's head"
but I preffer the more vague option wo HTML struct

PR #2640, opened 2021-06-19
slightly improve clarity of Promise.all()

ZYinMD:


iliakan:
Sorry, but the fix is wrong. When 1st promise dies, the `.all` dies too.

ZYinMD:
Thanks, I'll submit a different fix to explain my self better.

iliakan:
Sorry, I messed up. Your fix is perfectly correct.

iliakan:
Thank you!

PR #2642, opened 2021-06-19
(2nd try) improve clarity of Promise.all()

ZYinMD:
The original version was inaccurate because:
1. "settle" means "either resolve or reject" (think`Promise.allSettled`)
2. `Promise.all()` does not necessarily _resolve_ when all promises are _settled_.


iliakan:
Your fix in https://github.com/javascript-tutorial/en.javascript.info/pull/2640 was correct, thanks for the 2nd attempt. Merged it!

PR #2645, opened 2021-06-20
Added a missing explanation

law27:
It is mentioned that we should not use `setAttribute` for registering handlers. But it is possible to use `setAttribute` for registering handlers if we pass the function expression as a string. Requesting proper explanation for the statement **Don't use `setAttribute` for handlers** in the tutorial if the claim was flawless.


iliakan:
There's a general rule: use DOM properties unless an attribute is required. Hopefully mentioned somewhere.

PR #2647, opened 2021-06-22
chore: replace with spread syntax

109149:
It would be much more user-friendly to write `Object.assign(dest, [src1, src2, src3...])` with spread syntax instead.


iliakan:
The current syntax is more obvious.

PR #2651, opened 2021-06-23
Update prop descriptors article

ArtemTropanets:
Deletes info that it's possible to assign get/set for non-configurable accessor property if absent.

Maybe I just misunderstood. Could you please add some example about it if I'm wrong.

Please check this example.
https://codepen.io/artemtropanets/pen/bGqPKYx
If we have only a getter before setting 'configurable' to 'false' - then setter is 'undefined' and it's not possible to overwrite it later.
If we don't have get/set before changing 'configurable' then it's data property and it's not possible to change it to accessor property.


iliakan:
Thanks! I'm not exactly sure, what was the thing there, but looks like your example contradicts the current text.

Anyway, there's no need in the explicit list, I removed it in https://github.com/javascript-tutorial/en.javascript.info/commit/614c85ad5de601d941ea72292a3c3e927d43bafc.

PR #2654, opened 2021-06-24
add notes about CSS animation performance

### LINKED ISSUE #2652: Mention performance concerns of CSS animations on properties other than opacity and transform
The CSS `opacity` and `transform` properties are the only 2 best candidates for animation, because they don't trigger repaint. All other properties are less performant. Many articles talk about this fact, for example [this one](https://web.dev/animations-guide/#triggers).

In many examples and tasks in `7-animation/2-css-animations`, properties like `left` or `height` are being animated. They're fine being examples, but I think it'd be great if we could mention that they can usually be replaced by `transform: translate` and `transform: scale`.

ZYinMD:
Closes #2652.

### DIFF
--- a/7-animation/2-css-animations/article.md
+++ b/7-animation/2-css-animations/article.md
@@ -407,6 +407,53 @@ There are many articles about `@keyframes` and a [detailed specification](https:
 
 You probably won't need `@keyframes` often, unless everything is in constant motion on your sites.
 
+## Performance of CSS animations
+
+Most CSS properties can be animated, because most of them are numeric values. For instance, `width`, `color`, `font-size` are all numbers. When you animate them, the browser gradually changes these numbers frame by frame, creating a smooth effect.
+
+However, not all animations will look as smooth as you'd like, because different CSS properties cost differently to change. Among all the properties, `transform` and `opacity` are the cheapest to animate, and produce the highest framerate, so in real life projects, you should try to only use these two to achieve your desired effects.
+
+The reason for this is when something is being _transformed_, the browser engine is able to only worry about changing the look of this one thing, without touching other existing things on the page. Same for opacity change.
+
+In more technical details, when there's a style change, the browser goes through 3 steps to render the new look:
+
+1. **Layout**: re-compute the geometry and position of each element, then
+2. **Paint**: re-compute how everything should look like, including "layers" which is a big deal, then
+3. **Composite**: render the final results into pixels on screen.
+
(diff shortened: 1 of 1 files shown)


iliakan:
Thanks! I changed some phrases and rearranged them a little bit, please see https://javascript.info/css-animations#performance

ZYinMD:
Thanks. I want to point out that I have once seen a sentence [here](https://web.dev/animations-overview/#paint) that mentioned that `paint` is often the slowest step, not `layout`.

iliakan:
Better now?

joaquinelio:
Nice job, doc
👍

joaquinelio:
layout vs paint
I guess it depends on cpu/gpu/screen hardware

joaquinelio:
Hi Ilya,
is 440 grammar ok?  sounds odd
I would also write "mozilla gecko" or just "firefox"  so it doesnt sound like a stowaway

iliakan:
@joaquinelio what's wrong? Is 440 a line number?

joaquinelio:
oops sorry
 Not sure 
your Engl is better thn mine

440 The `opacity` property also never causes Layout (also skips Paint in Gecko). We can...

joaquinelio:
my bad "also never" seems correct. 
apologies

but the 2nd silly suggestion may stand...

iliakan:
@joaquinelio Changed to Mozilla Firefox.

ZYinMD:
Yes 440 is a line number. It is now [442](https://github.com/javascript-tutorial/en.javascript.info/blame/4ed51aa48cb667c87e27bfbf05450d1985c80a9d/7-animation/2-css-animations/article.md#L442). 

> The opacity property also never causes Layout (also skips Paint in Gecko). We can...

Yes I agree it sounds weird to say somethings "causes Layout". It'd be better to say it "triggers Layout", even better if we say it "triggers the Layout step".

Line [438](https://github.com/javascript-tutorial/en.javascript.info/blame/4ed51aa48cb667c87e27bfbf05450d1985c80a9d/7-animation/2-css-animations/article.md#L438) also uses "cause" which could be changed to "trigger".

I [originally](https://github.com/ZYinMD/en.javascript.info/blame/b855f6c9db21daa8fee761b79ceb879befb75932/7-animation/2-css-animations/article.md#L424) used `1` and `1` `2` and `1` `2` `3` when talking about these triggers, which was good too.

iliakan:
@ZYinMD changed to trigger.

joaquinelio:
Uhm...
@iliakan 
you added a redundant line 46 (link to bezier is in line 40 too)

sounds to me
 "...you might want to check the *previous article* you *just read* in *this very same chapter*  :)

I would combine it with line 40, or place it before 40 below the heading.

iliakan:
done

PR #2656, opened 2021-06-26
Added rest parameter

Slogghykitten5:
I was learning with javascript.info and freecodecamp together and noticed the rest parameter missing in the functions article.
I added it, the paragraph is copied from freecodecamp as it think it is easy to understand.

### DIFF
--- a/1-js/02-first-steps/15-function-basics/article.md
+++ b/1-js/02-first-steps/15-function-basics/article.md
@@ -270,6 +270,21 @@ showCount(null); // unknown
 showCount(); // unknown
 ```
 
+## Rest Parameter
+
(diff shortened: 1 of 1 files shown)


joaquinelio:
//javascript.info/rest-parameters-spread

Slogghykitten5:
Shouldn't it be in the basics? or is it because the learner doesn't know about array?

iliakan:
That comes later in the tutorial.

And you're absolutely correct, that's because the learner doesn't know about array yet.

joaquinelio:
Maybe a mention, a vague "other ways" "later in this tutorial"...

Slogghykitten5:
I also think this will be good, so people like me won't get confused. 
Like a link saying "there are [other advanced methods](https://javascript.info/rest-parameters-spread) which we will later in the tutorial"

PR #2657, opened 2021-06-27
 Add Generator.return

Manik2375:
I think adding `generator.return()` will be nice.


Manik2375:
@iliakan hey! what you think about this ?? :yum:

iliakan:
Yes, sure! Thanks!

PR #2661, opened 2021-07-03
Response header name is changed to the right one

WebSavva:
According to Wikipedia, there doesn't exist a header with a name "Accept-Control-Allow-Credentials". There is only one header that has similar name "Access-Control-Allow-Credentials". It seems to be exactly what was meant initally. By the way, it is mentioned in previous chapters as well. So it appears to be quite reasonable to fix that typo if I'm not mistaken.


iliakan:
Good catch, thanks!

PR #2663, opened 2021-07-05
Case and filesystems

joaquinelio:
* not comfortable in adding this so early in the tutorial but there is no "filesystem" topic.
You may want to drop it if it is too off-topic or distracting, or put it somewhere else.

* This is not THE reason for using all-lowercase. May have a pretty offtopic one.

### DIFF
--- a/1-js/02-first-steps/01-hello-world/article.md
+++ b/1-js/02-first-steps/01-hello-world/article.md
@@ -122,6 +122,18 @@ The example above can be split into two scripts to work:
 ```
 ````
 
+````warn header="Case and filesystems"
+The browser relies on the OS to find a file, so
+
+```html
+<script src="./myscript.js"></script>
+```
+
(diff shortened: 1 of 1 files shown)


kkanekii:
Another section for this would be nice

joaquinelio:
problem is,
there is no filesystem for js

node has its own libs to work with paths

kkanekii:
Hmm, then theres no need for a seperate section, a small addition should do, I can help out if required

joaquinelio:
any idea welcomed, 

I dont like this PR.  It s not a real js issue

but beginners working on windows may find 
their local code not working when they move it to the server or other clients.

"CamelCase not good for filenames"

iliakan:
That's an issue. But not quite a JS thing.

Linux guys usually understand this well, so they have no problem.

Mac/windows have this issue when going to Linux - that's true, but again not JS.

PR #2664, opened 2021-07-06
add finally return info

ArtemTropanets:


1-js/10-error-handling/1-try-catch/article.md
```diff
@@ -561,6 +561,28 @@ function func() {
 
 alert( func() ); // first works alert from finally, and then this one
 ```
+
+If there is a `return` in both `then` and `finally`, function returns the value from `finally`.
```
vsemozhetbyt, line 565: Maybe:
```suggestion
If there is a `return` in both `try` and `finally`, function returns the value from `finally`.
```

iliakan:
You mean "try", right? and then it's kind of obvious :/ Because `return` works after `try`.

PR #2674, opened 2021-07-19
Update article.md

Joelmmm:
It uses SyntaxError for the example while the text refers to ValidationError and the possibility to extend it.


iliakan:
The line `(*)` is about `SyntaxError`.

The fix appears to be wrong.

PR #2675, opened 2021-07-20
change defenition of absolute and relative path

masoudtabatabaei:
I think a part of content is against with absolute and relative path definition that I modify this

### DIFF
--- a/1-js/02-first-steps/01-hello-world/article.md
+++ b/1-js/02-first-steps/01-hello-world/article.md
@@ -73,7 +73,7 @@ Script files are attached to HTML with the `src` attribute:
 <script src="/path/to/script.js"></script>
 ```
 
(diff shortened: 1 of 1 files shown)


joaquinelio:
careful! 
nobody is able to change that definition!
`/path` is absolute from ancient times

relative, without the root /, means your path moves with your file

masoudtabatabaei:
> careful!
> nobody is able to change that definition!
> `/path` is absolute from ancient times
> 
> relative, without the root /, means your path moves with your file

I commented based on this link [https://www.w3schools.com/html/html_filepaths.asp](url)

iliakan:
The article is correct to my opinion.
Please note, I'm speaking about the absolute *path*, not absolute *URI*.

Also, w3schools has never been the trusted source.

PR #2679, opened 2021-07-24
Fixing the unintentional behavior of the slider

WebSavva:
When reading the article "Pointer events", I came across just a small bug.  When the pointer capture is set to the thumb, all consequent events including "pointermove" are retargeted back to the thumb. However, when the capture is lost, event listener for "pointermove" remains installed on the thumb element.  As a result, we can move inadvertently that thumb when we simply moving our mouse over it. To prevent its unintentional behavior, I suggest that we store capture pointer's id in a separate variable as long as capture persists.

### DIFF
--- a/2-ui/3-event-details/6-pointer-events/slider.view/index.html
+++ b/2-ui/3-event-details/6-pointer-events/slider.view/index.html
@@ -8,16 +8,20 @@
 <script>
   let thumb = slider.querySelector('.thumb');
   let shiftX;
-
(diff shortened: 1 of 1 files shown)


joaquinelio:
!!!
amateur opinion:
I didnt test the fix, but I see the bug.
I would add  a single comment, line 23 maybe:  // Prevents... etc

iliakan:
Thanks!

Fixed, I hope. Code changed.

PR #2680, opened 2021-07-26
old and new

joaquinelio:


joaquinelio:
We can't open an **newer** version of the database

PR #2684, opened 2021-08-03
current origin meaning

joaquinelio:
or something

I remember my first dumb idea of having a local html backup to access my db while offline and sync with origin when online (I learn the impossibility when I realized the security issue. Btw, pinning the webpage in chrome works.)

here I read this line "current origin" and "different sites cant access each other dbs" 
with "current" I see an open door "not now". 
I thought it's a good place to make a strong statement, it is not that evident if you miss cross-origin articles...


joaquinelio:
"different sites cant access each other dbs"
clear as it is

PR #2685, opened 2021-08-03
Update article.md

kkanekii:
made the language better, removed the code and added normal text


iliakan:
The new text is bad, sorry. Also has mistakes.

PR #2687, opened 2021-08-08
fix: minor typo error

leviding:
As title


iliakan:
The ; there was added by a native English speaker. Are you absolutely sure it should be removed?

PR #2688, opened 2021-08-08
Nothing to fix

joaquinelio:
Why this is not an issue? The site shows it ok, as a single line.

should I translate to ES in 2 lines to keep the numbering?


iliakan:
Doesn't matter, you can translate as 1 line.

PR #2689, opened 2021-08-08
slider not working on touch devices

joaquinelio:
I did only a quick Test on plunker, I guess this is the file it uses.

if not ok, consider it an issue


iliakan:
What's no working? What is movile? What's the purpose of the suggested fix?

joaquinelio:
I meant cellphone

the slider example works ok with the mouse,
bur keeps losing control on touch devices, you cant move it far.

iliakan:
Is it better now?

PR #2690, opened 2021-08-09
Added word "hoisting"

yeadamk:
This will help the readers to familiarize with the word "hoisting" while learning new JavaScript concepts.

As a beginner, I would love to have this word added to the text!


iliakan:
Adding a new opaque term doesn't make the text better, sorry.

PR #2691, opened 2021-08-09
Added a small explanation in article.md

olegbarabanov:
Added an explanation, since constants in JS do not make objects immutable


iliakan:
A variable stores a reference to an object, so it's immutable. The reference can't be changed, no contradiction. 
(the object issue is explained in further chapters).

olegbarabanov:
@iliakan , thank you for your the explanation and I completely agree with you. But I think that at this point it is still worth adding something like "... but a constant **referring** to an object does not make it immutable ..." with a link to the section with an explanation (link to "Const objects can be modified" within [/1-js/04-object-basics/02-object-copy/article.md](https://github.com/javascript-tutorial/en.javascript.info/blob/master/1-js/04-object-basics/02-object-copy/article.md))
If you agree that such a thing is necessary, i can try to offer a more correct fix (with or without issue).

PR #2697, opened 2021-08-15
promise chain flow

joaquinelio:
not to merge as it is

* changes, 
 if the handler doesnt return a promise,  .then returns a resolved one
am I correct?

* It feels like the async stuff is an issue to deal with.  It may worth a mention to beauty of concurrency here (or in basics) , before the cold API chapter.
Like: "why async?"


iliakan:
Yes, that's correct. I'll merge and fix the typo.

PR #2698, opened 2021-08-16
edge

joaquinelio:


joaquinelio:
btw,
aren't you tired of "but IE" all over the turorial? 

I would only let a warning here in this article
"why IE" 

TLDR
even new MS products dont support it . and old products are going to drop it too

I worked for Atlas Copco 5 years ago, they had IE for a few legacy apps but they let users install Edge and Chrome, and their new corporate apps were written for the new platform

neelkarma:
you could also say that many chromium-based browsers use v8, instead of just saying chrome, opera and edge

iliakan:
Yes, let's remove all IE stuff.

PR #2707, opened 2021-08-29
Adding scrollTop to make text area scroll down.

Manik2375:
When adding keys in textarea, we need to scroll it down again & again. So a little fix for it :yum:


Manik2375:
I added a little text also, that `beforeunload` may not work as intended with `event.preventDefault()`

What do you think @iliakan ^^ I think it's important because some people may get confused with this

PR #2708, opened 2021-09-01
after transitioned event button click is removed

arvind248:
I have  added code to remove click event of button instead of removing transitioned event to stop the user to interact again.


iliakan:
Please elaborate, what exactly is wrong?

arvind248:
actually, when button is clicked again , the circle is rendered again on the earlier circle it seems to be nothing happened on click. 
if the intention is to stop user clicking more then one, I have added ``` btn.removeEventListener("click",go);   ```  to remove click event of button and removed this line ```div.removeEventListener('transitioned',handler)  ```

PR #2728, opened 2021-09-30
Remove a whitespace

mahdiHashemi14:
When the code runs, there is an unwanted whitespace at the beginning of the second line. 
It's removed so it looks nicer


iliakan:
Why "unwanted"? :)

Looks good 🤷

PR #2748, opened 2021-10-17
"for" block

joaquinelio:
I dont like my writting, but.

why. Consider 
```
for ( let i=1;;)  i++ ;     // no brackets, still hidden from the outside
for ( let i ; i<1  ;  i++, j="" ) {  let j ; ... }   //  err, block inside block (couldnt test wo brackts  )
```
nothing "special" about it. It behaves 
{for ( ){ } }

something similar in while for / label article
you dont label a line to jump to,
you label the for block you want to escape from.
label: { for ( ){ } }

### DIFF
--- a/1-js/06-advanced-functions/03-closure/article.md
+++ b/1-js/06-advanced-functions/03-closure/article.md
@@ -69,7 +69,7 @@ alert(message);
 ```
 ````
 
-For `if`, `for`, `while` and so on, variables declared in `{...}` are also only visible inside:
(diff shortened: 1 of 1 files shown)


iliakan:
Please take a look at the specification: https://tc39.es/ecma262/#sec-runtime-semantics-forloopevaluation.

image

Actions 6-7 literally describe the special actions for the "for" statement, saying that the `LexicalDeclaration` in `for(let...)` should be evaluated in the context of `loopEnv`.

This is described apart from other loop constructs, and apart from other blocks. So the words "special case" can be used here, because it's just that - a specific "for" algorithm.

PR #2758, opened 2021-10-23
fix sentence

KennethKinLum:


essilfie00 (approved):
Yes, this sentence fix is correct.

PR #2769, opened 2021-10-31
Update article.md

kk-source:
Added missing word in a the article.


iliakan:
No no, it was correct.

PR #2777, opened 2021-11-16
Included missing showStep methods

Tofpu:
I have seen plenty of people in the comments asking why there was a need for the showStep method to be chainable as it was never shown neither in the task nor in the solution, although it was still mentioned anyway. this PR fixes that!


iliakan:
Thanks =)

PR #2778, opened 2021-11-30
Fix : Added the difference between Let and Const

ra-kesh:
It is mentioned above that `const` also behaved same as `let` , then gave the example of how `let` can be initialised with `undefined` . Which can be confusing to some people ,because I believe `const` can't get initialised to `undefined` , as it will be against it's fundamental purpose .  

let me know if I am thinking this in a wrong way or what I need to change in my PR 

thank you...

### DIFF
--- a/1-js/06-advanced-functions/03-closure/article.md
+++ b/1-js/06-advanced-functions/03-closure/article.md
@@ -15,7 +15,7 @@ Let's expand our knowledge to understand these scenarios and more complex ones.
 In JavaScript, there are 3 ways to declare a variable: `let`, `const` (the modern ones), and `var` (the remnant of the past).
 
 - In this article we'll use `let` variables in examples.
-- Variables, declared with `const`, behave the same, so this article is about `const` too.
+- Variables, declared with `const`, behave the same except it does not get initialised with `undefined`, so this article is about `const` too.
 - The old `var` has some notable differences, they will be covered in the article <info:var>.
 ```
 


Aleksey28:
Hi, 
I beg to differ, it isn't illegal by the JS engine. If you do it you won't get any exception. And this is exactly what is meant here, not about a reason why we should do it.  Initialize const like undefined is senselessness, but anyway we can do it.

ra-kesh:
> Hi, I beg to differ, it isn't illegal by the JS engine. If you do it you won't get any exception. And this is exactly what is meant here, not about a reason why we should do it. Initialize const like undefined is senselessness, but anyway we can do it.

Hello , I may have not explained clearly enough . I know we can initialise `const` with `undefined` . All I was saying `const` needed to be initialised during the declaration, unlike `let` which can be declared without being initialised. 

that's why the below example can be confusing for people as `const` doesnot behave same as `let` here. 

chrome_sCiv61Oj7k

thank you..

joaquinelio:
I didnt check it, 
but It is harder to read and I don't think it worths it

rajat-mehra05 (approved):

ra-kesh:
> I didnt check it, but It is harder to read and I don't think it worths it

There must be a way to mention this subtle difference somewhere without comprising the readability . That's the whole point of this guide , isn't it ???

iliakan:
I'm not sure we need to explicitly say that `const x` is illegal.

If someone tries it, there'll be an obvious error.

PR #2780, opened 2021-11-30
as of now, from now on

joaquinelio:


iliakan:
Please see https://www.quora.com/What-is-the-difference-between-as-for-now-and-as-of-now#:~:text='As%20for%20now'%20means%20',your%20mother's%20condition%20is%20stable.

joaquinelio:
Done.
300 votes nearly convinced me!  
Still...

TLDR, nobody will fail the task because of this.

It says
>“as of now”  "what’s happening or exists right now in your domain."
 
But other references I read point to the present AND the future.
Like "from now on"
>"as for now"  "about the past and are leading into the present". 

it looks like the right option to me.
Like "until now"

It's ok; even when I don't agree, this is still the best tutorial and that s because of a reason.

PR #2781, opened 2021-12-02
fix the broken RFC link

chanjsq:
The original URL http://tools.ietf.org/html/rfc6455 is broken, return a 404 status code. Refresh it to the new URL https://datatracker.ietf.org/doc/html/rfc6455.


iliakan:
Thanks!

PR #2784, opened 2021-12-07
Clarify that generators are both iterator and iterable

zzJinux:
The fact that a generator object is iterable, has nothing to do with it
having the `next()` method. It is according to the [spec][1]:
A generator object is *both* iterator and iterable.

I try to introduce change as small as possible, but I wish that
iteration protocols are explicitly mentioned somewhere.

[1]: https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-generator-objects


iliakan:
I don't mind additional explanations, but your fixes seem to make the article more difficult to understand.

PR #2790, opened 2021-12-11
fix typo

yangsong97:


joaquinelio:
<html>
<body>
<p>Do we want/need accuracy?</p>
<p>From 
https://developer.chrome.com/docs/devtools/shortcuts/
</p>


Action | Mac | Windows / Linux
-- | -- | --
Open whatever panel you used last | Command+Option+I | F12 or Control+Shift+I
Open the Console panel | Command+Option+J | Control+Shift+J


</body>
</html>

iliakan:
The text specifically explains that the command opens the Console.

Issue #2787, opened 2021-12-10
[Pointer-Events] touch-events should be touch-action


justforuse:
In the pointer-events section, it is said 
> set touch-events: none in CSS for elements that we engage.

 I think it should be `touch-action: none`, because there is no `touch-events` in css, and have mentioned `touch-action` before

PR #2792, opened 2021-12-13
Typo: touch-events should be touch-action

justforuse:


justforuse:
related issue: https://github.com/javascript-tutorial/en.javascript.info/issues/2787

PR #2799, opened 2021-12-16
Added reference for `for...of`

akxavier:
A reference link to mdn docs for `for...of` has been added for better clarity.


joaquinelio:
If so, link to abetter source
 /iterable

iliakan:
At this stage of the tutorial, we shouldn't point to this "advanced" manual article. The user will go there, see unknown terms (iterable) and get confused.

PR #2801, opened 2021-12-20
fix the broken RFC link in the JSON article

tburette:
In the JSON article, a link leads to an 404 page. Changing http to https makes the link redirect to the correct page instead.
**http**://tools.ietf.org/html/rfc4627 => **https**://tools.ietf.org/html/rfc4627

Same problem as https://github.com/javascript-tutorial/en.javascript.info/pull/2781 that was applied on the websocket page.  
I looked for all the pages with a link to ietf.org and they will all be OK now : 
https://javascript.info/json
https://javascript.info/cookie
https://javascript.info/websocket


iliakan:
Thank you!

PR #2804, opened 2021-12-24
fix: changed 'within' to 'without'

Aleksey28:
I think here was meant 'without' instead of 'within'. Because `documentElement.clientHeight/clientWidth` methods get size without scrollbars.

### DIFF
--- a/2-ui/1-document/10-size-and-scroll-window/article.md
+++ b/2-ui/1-document/10-size-and-scroll-window/article.md
@@ -29,7 +29,7 @@ alert( window.innerWidth ); // full window width
 alert( document.documentElement.clientWidth ); // window width minus the scrollbar
 ```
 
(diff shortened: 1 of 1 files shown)


iliakan:
I hope, within is ok. It's as intended.

Aleksey28:
I think it may be confused and using 'without' is more suitable.

iliakan:
I really mean "within" here ;)

iliakan:
"something without scrollbars" may lead to another confusion.

Aleksey28:
On top of this topic was written 'clientWidth/clientHeight provide the width/height without it'. But here was written 'within scrollbars (if there are any), so we should use documentElement.clientHeight/clientWidth'. Maybe it should be identical to avoid confusion.

PR #2808, opened 2021-12-30
Correction to Part I - 14.2 "Eval"

696f7a:
I think there's a mistake in a code snippet which illustrates usage of `eval` in strict mode. As `let` creates a block scope variable, such a variable won't be visible neither in strict mode, nor in sloppy mode. A `var` should be used instead.


696f7a:
P.S. Is it also worth mentioning in the article that `eval` creates a block?

iliakan:
Thanks, I rewrote a large section of eval, to add these changes.

PR #2812, opened 2022-01-06
:books: fixed destructuring statement and made semantically correct

rahulk31:
The statement that stated, "unpack arrays or objects into a bunch of variables" is somewhat semantically wrong because of the word "bunch" (as it means 12). It may not necessarily unpack value of an array into 12 variables, if the array itself has 2 elements contained. 

This edit was done to make it semantically correct as many devs refer this project who are not native to english and find it difficult to understand.

Thank you! :)

### DIFF
--- a/1-js/05-data-types/10-destructuring-assignment/article.md
+++ b/1-js/05-data-types/10-destructuring-assignment/article.md
@@ -7,7 +7,7 @@ The two most used data structures in JavaScript are `Object` and `Array`.
 
 Although, when we pass those to a function, it may need not an object/array as a whole. It may need individual pieces.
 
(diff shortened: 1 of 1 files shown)


joaquinelio:
I'm not a native Eng speaker, but I see it harder to read

1st part
 I understand you don't unpack individual elements but the entire thing
...to "unpack" an array or object into... 
sounds ok to me.

2nd part
"Bunch" sounds more flexible (I dont know about grammar).

Just my opinion.

iliakan:
It's more confusing after your fix. As if you're unpacking elements, not the array itself.

rahulk31:
Alright, thanks.

PR #2815, opened 2022-01-08
Update article.md

### LINKED ISSUE #2811: adding a note in regex lookahead and lookbehind for browser compatibility
It seems that look-behind is not supported for IE and Safari.It would be helpful to add the same in the article. https://javascript.info/regexp-lookahead-lookbehind

prachurjya15:
Added a note for browser compatibility for regex lookbehind, hopefully will save a lot of time for someone referencing this article. Fixes #2811


iliakan:
Thanks, added!

PR #2817, opened 2022-01-09
cleaner with a dot.

joaquinelio:


iliakan:
There's no real benefit in replacing `(...)` with `,...,`

joaquinelio:
I have my own issues with commas ;)

Here, it is very clear because of context. 

but when I read
the character (`xxxxx), for non-character keys,
I tend to stick 
the character (xxxxx) for non-character keys, 

maybe a semicolon...

PR #2818, opened 2022-01-09
cleaner reading

joaquinelio:


joaquinelio:
too many fragments, the () keeps it clean
im not sure about the other change.

not very important, the article is easy to understand if you followed the path

joachimklug (approved):

iliakan:
This clause is essential to understanding, better to  leave it as is.

joaquinelio:
I think it's a little hard to follow (too many commas) and it shouldn't be. 

But I didn't like my "fix" either.

PR #2827, opened 2022-01-15
Fix edge case in propogation of NaN in mathematical expressions

arun-abraham-108:
This PR fixes an error in [Data types](https://javascript.info/types#number) chapter of the tutorial. In this chapter, it is written that if there’s a NaN somewhere in a mathematical expression, it propagates to the whole result. However, this is not completely true for [exponentiation operator](https://262.ecma-international.org/12.0/#sec-numeric-types-number-exponentiate). NaN ** 0 is defined as 1.

### DIFF
--- a/1-js/02-first-steps/05-types/article.md
+++ b/1-js/02-first-steps/05-types/article.md
@@ -52,7 +52,7 @@ Besides regular numbers, there are so-called "special numeric values" which also
     alert( "not a number" / 2 + 5 ); // NaN
     ```
 
-    So, if there's a `NaN` somewhere in a mathematical expression, it propagates to the whole result.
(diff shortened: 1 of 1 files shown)


iliakan:
Feels like a bug in the specification, doesn't it?

joachimklug:
Not sure it is a bug. Looks they did place it intentionally in between the exponent and base check for NaN.
I wouldn't however mention this edge case here as it is far too detailed and will confuse the average reader.

iliakan:
Rephrased a bit. Mentioned that now, thanks!

joaquinelio:
Now I wonder why.
politeness ** 0 == 1
metaphysical spec

iliakan:
The spec is definitely bugged. `0 ** 0` is `1`, but  "Zero to the power of zero is a mathematical expression with no agreed-upon valu".

PR #2832, opened 2022-01-17
Update solution.js

ManuMathewVarghese:
A better general solution.


iliakan:
According to the task formulation, the current solution will suffice.

PR #2834, opened 2022-01-21
kill IEs

joaquinelio:


joaquinelio:
No PRs left, congrats.

I just got covid and I'm MAD so this PR is a negative one.

IE is not even supported in new MS soft 

every kill may need some review

You may want to add a note, maybe in polyfills 
"why you may still ear about IE" or something and put the link to poltf  & libs there
but I think IE developm is a too confined niche, if any.

iliakan:
I'm sorry to say so, but IE isn't fully dead yet. In some countries, such as China, it's still actual.

joaquinelio:
Oh, I didnt know that.

Why? 
I would warn instead, "refuse to run in old browsers... security issues"

ok, you may close it.
I dont have the courage to do so.

joaquinelio:
I would still keep the tutorial clean
and put all the IE traps in a separate chapter
ugh
dirty

iliakan:
There are not many :-)

Best regards,
Ilya Kantor

iliakan:
Maybe you could add something new?
A Chapter about something, not yet covered?

Best regards,
Ilya Kantor

joaquinelio:
Im not that good.

"not covered"?  javascript.info is my bible, nothing exists outside.

I would add a task here and there, but I cannot think of any.

iliakan:
There're so many. Like, Canvas.

Issue #2563, opened 2021-04-02
How does ?? produce mistakes?


lumosmind:
> The limitation is surely debatable, it was added to the language specification with the purpose to avoid programming mistakes, when people start to switch from || to ??.
from: [nullish-coalescing-operator](https://javascript.info/nullish-coalescing-operator#comparison-with)

How can it be possible to produce mistake while changing `||` to `??` . There is not any other operator between their precedence order.

> 

Precedence | Operator type | Associativity | Individual operators
-- | -- | -- | --
7 | Logical AND (&&) | left-to-right | … && …
-- | -- | -- | --
6 | Logical OR (\|\|) | left-to-right | … \|\| …
5 | Nullish coalescing operator (??) | left-to-right | … ?? …

from : [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table)

iliakan:
Maybe you'd better look for reasons and ask at the forum https://es.discourse.group/ ?

I have my guesses, but may be wrong :/

joaquinelio:
better to mention it here:
#2836 ?

PR #2836, opened 2022-01-22
Correction to precedence levels

erolaliyev:
Nullish coalescing and logical OR both equal to 4 in MDN table now.


joachimklug (approved):
Good catch!

joaquinelio:
#2563 ?

Issue #2826, opened 2022-01-14
Object methods, "this": Missing appropriate tests for chaining exercise


Omid-Heidarzadeh:
### Article
[Object methods, "this"](https://javascript.info/object-methods#chaining)

There are missing tests for the "chaining" exercise on the "Object methods, this" article to ensure that the showStep method returns this. 

Suggested tests:
 ```javascript
  it('showStep() should return this', function() {
    assert(ladder.showStep(), ladder);
  });

  it('up().up().down().showStep().down().showStep()', function () {
    assert.equal(ladder.up().up().down().showStep().down().showStep().step, 0)
  })
```

iliakan:
Might be good. Make a PR, please?

PR #2837, opened 2022-01-22
fix:Add missing test issue:#2826 & fix related solution

Omid-Heidarzadeh:
Added required tests that ensure "showStep()" method in article [Object methods, "this", chain-calls](https://javascript.info/object-methods#chaining) returns this, thus the result will remain chainable after using this method.


iliakan:
Thanks!

Omid-Heidarzadeh:
You're welcome.

PR #2839, opened 2022-01-25
fixes operator precedence

ruslauz:
@iliakan pr copy from [here](https://github.com/javascript-tutorial/ru.javascript.info/pull/1478)


jovton:
If you change "17" to "16", then line 208 must also be changed to reflect this change, i.e. ``"unary plus" has a priority of `17` which is higher`` becomes ``"unary plus" has a priority of `16` which is higher``.

ruslauz:
@jovton Done, thank you!

iliakan:
Thanks. MDN ever renames something.

PR #2844, opened 2022-01-27
Add alternative solution for throttle decorator task on article "Decorators and forwarding, call/apply"

Omid-Heidarzadeh:
This alternative solution for the ["throttle decorator"](https://javascript.info/call-apply-decorators#throttle-decorator) has similar look to the solution for the "Debounce" task, and probably will be easier to grasp, especially for new learners.


iliakan:
Are you 100% sure, that your solution is better and easier to grasp?

Omid-Heidarzadeh:
In the comment section, there was some argument about the whole article and its difficulties, so I decided to help learners by suggesting a similar solution for the last task which seems to be the most complicated one.

This solution uses the same technique to manage calls (using `setTimeout` alongside `clearTimeout`) just like the solution for [debounce-decorator](https://javascript.info/call-apply-decorators#debounce-decorator) and it doesn't need to save `this` and `arguments` explicitly, so it could be a good alternative solution. Also, I think the current solution can inspire learners to see problems from different perspectives, and thus it should remain and users can use both.

Besides, I tried my best to clearly explain the algorithm if it's not that easy for some users. Also, feel free to change anything that you think needs to be changed to make it easier or cleaner.

PR #2852, opened 2022-02-04
remove deprecated `document.domain` setter

ZYinMD:
`document.domain` is becoming immutable in future chrome releases: 
https://developer.chrome.com/blog/immutable-document-domain/


iliakan:
Is there a replacement?

ZYinMD:
I guess "Replacement" will depend on your goal. If the goal is to communicate between iframes, then `postMessage` still works if you control both pages. If you want to share indexedDB or localStorage or BroadcastChannel, I think it never worked in the first place. If you want to access the window object of an iframe, e.g. `iframe.contentWindow.fetch()` for CORS purposes, then I'm not sure if there's any replacement.

iliakan:
"Replacement" is something that has the same functionality.
As of now, there's none. And the support is still everywhere (and will be for a long time).

I'll add a warning.

PR #2853, opened 2022-02-04
Update operator precedence

leviding:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table


iliakan:
Will that renumbering ever stop?

PR #2866, opened 2022-02-07
GB unit symbol

joaquinelio:


joaquinelio:
bytheway, I know I mentioned this before.
I would encourage "stream first", chunks wherever possible as best practice
instead of warnings "if it's too big"

iliakan:
Sorry, not getting, what do you mean? Should we change something?

joaquinelio:
The text is more than ok

What I'd like to add is some advice: do always prefer chunks over monolithic blocks (if you can) to process,
treat them like burning match game
keep lightweight
dont hold breath
free resources quickly

but I dont know how to add it without distracting from the important concepts.
Just forget it, the article is perfect and I am insane, more than yesterday.

PR #2867, opened 2022-02-09
Added a scheme reflecting the difference in approaches

Rnbsov:
I think it will help newbies understand this intricate theme


iliakan:
Thanks, we considered it, but not sure it's needed there.

PR #2869, opened 2022-02-09
Fix misleading statement

vabushkevich:
From the current wording it may seem that the historical feature is the ability of `focus/blur` events to propagate down on the capturing phase. Since all JavaScript events go through the capturing phase, I want to suggest using a more unambiguous wording.


iliakan:
You totally remove the important sensible phrase, it's unacceptable, sorry. Can rephrase, though.

PR #2871, opened 2022-02-10
Check input to be an Array before using Array.map

Omid-Heidarzadeh:
As the beginning of [Promise API](https://javascript.info/promise-api) article mentions that technically the input should be iterable, then in the polyfill section, we should assume the same. So, we need to check the input type using `Array.isArray(promises)` or `promises instanceof Array` before using `Array.map()` or use it with `call` or `apply` method.


iliakan:
We're not making a 100% compatible polyfill here. 

If you want, there are options in the web, more close to the standard.

PR #2873, opened 2022-02-13
fix: remove extra blank lines and spaces

leviding:


iliakan:
We can, but that'll trigger syncs for all other translations. Not worth that.

PR #2884, opened 2022-02-19
Update Optional Chaining article.md

sbkrish:
`obj` instead of `obj.property` in Optional chaining syntax explanation


iliakan:
Are you sure?

PR #2887, opened 2022-02-21
added a little piece of text

Rnbsov:


iliakan:
See above: "So it has access to these functions."

PR #2888, opened 2022-02-22
propose additional context for removeEventListener

branlok:
Added a clarification and example for removeEventListener matching on eventListeners initialized with object handler. 
Previously a section mentions 'Removal requires the same **function**'. In the case of object handling, it is the object reference matters here.

demo:
https://jsfiddle.net/op5hjw47/20/


iliakan:
It's obvious... Isn't it?

Reopen if I'm mistaken =)

PR #2889, opened 2022-02-23
Line 181 moved "["

JPachecoZ:
Line 181 moved "[" according to the examples below


Rnbsov:
hmm...

In the official documentation like so

image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

joaquinelio:
it means optional

` = JSON.stringify  ( value )`   leaving the cooma out, 
or
` = JSON.stringify  (value , replacer, space )`

iliakan:
[ .. ] means optional

PR #2896, opened 2022-02-24
replacement the word with more appropriate term

Rnbsov:
Perhaps it would be more correct

A Wikipedia article could be attached to the term `Pseudorandom number` ( this is not necessary but as an option ) 

https://en.wikipedia.org/wiki/Pseudorandomness


iliakan:
The word "pseudorandom" gives more details about how the random value is generated, not sure we need them here.

Issue #2895, opened 2022-02-24
imperceptible link


Rnbsov:
In the article Data types in block [Tests: isFinite and isNaN](https://javascript.info/number#tests-isfinite-and-isnan)

There is a link that's easily can confuse with just a code

![image](https://user-images.githubusercontent.com/73550760/155566353-8fcb5442-1518-465e-a9a8-26c9146c8125.png)

What about make other highlighting for such cases
For instance, make such links always glowing, like when u hover mouse

![image](https://user-images.githubusercontent.com/73550760/155567874-b3b977ab-a68f-4442-b149-3877478de591.png)
I think that would solve the problem

PR #2897, opened 2022-02-26
mdn link

joaquinelio:


Rnbsov:
> #2895? kind of.
> mdn md limks look different

I don't think this will solve the problem

I don’t remember exactly in which article, but there is the same problem

It's a problem specifically when you wrap `links` like code

joaquinelio:
what I fix here is the link

"mdn:" markup was a good idea, but it fails with some transalated mdn pages
because someone also translated the link path.

About #2895,
maybe ...
There is [a special built-in method `Object.is`](https://developer.mozilla...

Rnbsov:
hmm...

yeah, this makes the link more visible

but it not solve problem globally, it's still worth making another highlight for such links

PR #2902, opened 2022-03-01
Date task solution: use standard date time string

jeffchase:
The Date article only explains standard strings for date parsing.
Non-standard strings are implementation dependent and should
not be used.

### DIFF
--- a/1-js/05-data-types/11-date/1-new-date/solution.md
+++ b/1-js/05-data-types/11-date/1-new-date/solution.md
@@ -13,6 +13,6 @@ We could also create a date from a string, like this:
 
 ```js run
 //new Date(datastring)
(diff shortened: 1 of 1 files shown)


iliakan:
What's the actual support of date strings?

It makes sense to mention some compatible formats, that are supported everywhere.

jeffchase:
@iliakan The [tutorial](https://javascript.info/date#date-parse-from-a-string) already describes the [standard format](https://tc39.es/ecma262/#sec-date-time-string-format) which looks to be [supported](https://caniuse.com/mdn-javascript_builtins_date_parse_iso_8601) by every major browser in the past decade.

However the solution to the first exercise uses a non-standard string. WIth this change I am suggesting updating the solution to use the standard format already described in the tutorial.

PR #2911, opened 2022-03-09
added fetch get example for non-utf8 pages

ProjectEli:
Still, some webpages do not follow standard UTF-8 encoding. As far as I know, `fetch()` does not support encoding options by default. Therefore, I added an example with error handling part (not already shown).

### DIFF
--- a/5-network/01-fetch/article.md
+++ b/5-network/01-fetch/article.md
@@ -98,6 +98,26 @@ let text = await response.text(); // read response body as text
 
 alert(text.slice(0, 80) + '...');
 ```
+If you are dealing with non UTF-8 charset pages, use `.arrayBuffer()` to get binary text and `TextDecoder()` to decode it into a specific encoding:
+
+```js run
+fetch('/article/fetch/non-utf8-page.html')
+  .then( response => response.arrayBuffer() )
+  .then( bytes => {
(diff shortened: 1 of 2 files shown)


saitharunsai (approved):

Rnbsov (approved):

iliakan:
Are there many pages in non-UTF in  Korean (or Chinese)?

Is it a common way of using `fetch`, then?

ProjectEli:
> Are there many pages in non-UTF in Korean (or Chinese)?

Yes. At least there are many Korean and Japanese websites that are not in UTF8, especially for some government websites. For Chinese, I don't have much experience. In practice, many people struggle with encodings when accessing those websites programatically. 

Ideally, the owners of those websites should follow international standards. However, in most cases they don't want to change, or just give up management. From the user point of view, we want to deal with what's inside rather than the charset encoding. Therefore, this example will be meaningful for beginners to utilize without a deeper understanding on non-UTF8 encoding.

> Is it a common way of using `fetch`, then?

I agree that this is not the intended usage of `fetch`. But we cannot decode those websites correctly through given methods in the document. In my opinion, it is worth including as a tutorial for practical use unless we find a better way of decoding.

iliakan:
up )

PR #2912, opened 2022-03-09
fix apparently an extra link

Rnbsov:


joaquinelio:
The link is useful if you want to change any tittle without worryng  "is there a link pointing to this anchor?"
And NEEDED for translated sites.

PR #2919, opened 2022-03-15
Minor grammar fix

Omid-Heidarzadeh:


Rnbsov (approved):
according to [StackExchange](https://english.stackexchange.com/questions/407810/what-is-correct-will-be-not-or-will-not-be) answers yeah, it's more correct 😃👍

but personally I would place `won't` instead

Omid-Heidarzadeh:
It seems you also caught another one 😉👍, so I'll close this PR.

PR #2920, opened 2022-03-15
Create for of loop

thisRuan:
hi, can anyone add the for of loop in the javascript info files?


Rnbsov (commented):
The explanation of this loop is already present in the guide
https://javascript.info/iterable

At such an early stage of learning, it is still too early to explain the `for...of` loop designed to iterate over objects

And next time, raise for such stuff a `issue`, not `PR`

PR #2923, opened 2022-03-17
add wikipedia link

Rnbsov:
I just thought it might be helpful


ShahriarKh (approved):

iliakan:
If you read the link, it talks about OOP and classes. Will be confusing, cause there's no class here.

Should we change a word "encapsulate" to something else instead?

iliakan:
Closed for the lack of feedback ;(

PR #2926, opened 2022-03-20
Update article.md

apoorvachikara:
- updated incorrect spellings
- Updated the grammar-related errors
- Updated punctuations

### DIFF
--- a/1-js/04-object-basics/01-object/article.md
+++ b/1-js/04-object-basics/01-object/article.md
@@ -40,11 +40,11 @@ In the `user` object, there are two properties:
 1. The first property has the name `"name"` and the value `"John"`.
 2. The second one has the name `"age"` and the value `30`.
 
-The resulting `user` object can be imagined as a cabinet with two signed files labeled "name" and "age".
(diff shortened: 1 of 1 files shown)


joaquinelio:
I prefer "labeled" but I'm not native English speaker.

-123 Should it be quotation marks intead of quotes? I dont like "quote" as "citation"

-328 sound broken now.   original is ok, it does mean a variable. (containing the name...)

apoorvachikara:
- I prefer "labeled" but I'm not a native English speaker  
       That is fine It won't change the meaning of the statement. I can revert it.
- 123 Should it be quotation marks instead of quotes? I don't like "quote" as "citation" 
        Yes, it could be more descriptive if we use "quotation marks" instead of just quotes or quote
- 328 sound broken now. original is ok, it does mean a variable. (containing the name...)
        The better way- If we omit values inside the quotes, it should contain the actual name to be tested. For instance:
       
If you are fine with the above changes, I can make these changes and request PR.

iliakan:
Merged, with "labeled" and "quotes" left "as is".

PR #2930, opened 2022-03-22
Fixed the display on the site

Rnbsov:
Most likely it was expected that it would look like this:

image

But on the website it looks like this: 

image

these lines are glued together


Rnbsov:
In this commit I also added missing line breaks

Issue #2932, opened 2022-03-24
Wrong units @ https://javascript.info/js-animation


adamnar:
It says: "The callback gets one argument – the time passed from the beginning of the page load in **microseconds**. This time can also be obtained by calling performance.now()."

https://developer.mozilla.org/en-US/docs/Web/API/Performance/now says: "The performance.now() method returns a [DOMHighResTimeStamp](https://developer.mozilla.org/en-US/docs/Web/API/DOMHighResTimeStamp), measured in milliseconds."

The units are milliseconds, not microseconds.

Rnbsov:
thank you for your report 👾

iliakan:
Thanks, fixed!

PR #2933, opened 2022-03-28
fix units

Rnbsov:
ref: https://github.com/javascript-tutorial/en.javascript.info/issues/2932


iliakan:
Microseconds is correct.

PR #2935, opened 2022-03-29
fix typo in the summary section

gnopor:


iliakan:
Maybe "have branches" is easier to understand =)

PR #2944, opened 2022-04-11
add italic for more emphasis

Rnbsov:


iliakan:
Must is a strong word enough =)

PR #2949, opened 2022-04-13
Added method Array.prototype.at

security-somanos:
I noticed this method was not in the docs. At least not in this section.


iliakan:
Nice catch, added to the main `array` article!

PR #2953, opened 2022-04-15
change some occurances of (sayHi()) => (getName())

Amir-nik:
In order to make the article more beginner friendly , readable & maintain clean code principles for naming functions, using 'getName()' is more appropriate than 'sayHi()' in this instance!


iliakan:
Never ever.

Please note: `get*` function should never ever do something like this. They should return a value only. No modifications, no side effects !!111

PR #2961, opened 2022-04-18
Make text unambiguous

danipoma:
It was possible to interpret text in two possible ways:
in case you need them - var declarations
in case you need them (to know) - subtle differences

Since var is not that much used (as said in the article and in following text concerning this), I incline towards second option (difference). Which is also what is discussed in referenced article.


iliakan:
Isn't the text obvious now? :/

PR #2967, opened 2022-04-22
semantics

joaquinelio:


joaquinelio:
respectively to what

"first and second, named a and b respectively"

iliakan:
The first one becomes `a`, the second one becomes `b`?

PR #2970, opened 2022-04-24
Removing unnecessary and confusing language

MondoBurrito:
The line: " not a "magical language structure", " is confusing and unnecessary. It is important to point out that functions are values, but these extra words only serve to muddy the point the author is trying to make.  We need to make this material as clear and concise as possible.

### DIFF
--- a/1-js/02-first-steps/16-function-expressions/article.md
+++ b/1-js/02-first-steps/16-function-expressions/article.md
@@ -1,6 +1,6 @@
 # Function expressions
 
-In JavaScript, a function is not a "magical language structure", but a special kind of value.
+In JavaScript, a function is a special kind of value.
 
 The syntax that we used before is called a *Function Declaration*:
 


Rnbsov:
Personally, I don't think there is anything superfluous in this

leviding:
I think it doesn't matter.

MondoBurrito:
> I think it doesn't matter.

Probably because you are already really good at coding ;)  From my perspective,  as someone who is literally learning the very first steps of JS and struggling to make sense of it all, any extra, confusing and unnecessary words should not be in the tutorials.

joaquinelio:
"magic" is metaphoric, It means there is nothing overcomplicated.
You may have a different style of writing (eg "just another value"), but the point is not superfluous.

smith558:
Well, in some languages functions are kind of magical structures. To me, it well conveyed the point that they are equal to objects, as everything else in JS.

PR #2980, opened 2022-05-01
add parentheses for the method

Rnbsov:


iliakan:
Other methods in this article don't have `()`.

PR #2992, opened 2022-05-07
remove extra break lines and add missed

Rnbsov:


Rnbsov:
period* i mean

iliakan:
Are you sure about the comma after "instead of"? Please note: English syntax rules are different from Russian.

Rnbsov:
Okay, I get you 😃👍

PR #2995, opened 2022-05-07
Fix: Edit code in line 243

hootan-rocky:
Added an underscore to the start of the private variable "name".
Which was missing from that line.


hootan-rocky (commented):
as you can see here, "_name" is used in the getter and setter

PR #3000, opened 2022-05-10
warning about copied methods

webFardin:
A warning about that user can't access methods of the parent object of the object that copied its methods to class, from the class.


iliakan:
Sorry, I'm not sure it's needed for the understanding.

webFardin:
Maybe someone thinks wrong!

PR #3006, opened 2022-05-13
return in sortBy, closes... closed #2965

### LINKED ISSUE #2965: Missing "return" keyword in sort-objects solution
This function misses the `return` keyword.

The solution.md is located [data types -> array methods -> sort-objects](https://github.com/javascript-tutorial/en.javascript.info/blob/master/1-js/05-data-types/05-array-methods/8-sort-objects/solution.md).

```js
function sortByAge(arr) {
  arr.sort((a, b) => a.age - b.age);
}
```

It should be:

```js
function sortByAge(arr) {
  return arr.sort((a, b) => a.age - b.age);
}
```

joaquinelio:
#2965 made a point.  `console.log(sortByAge(arr));`  undefined

I would always add a return even if dont needed. You need to learn to think ahead

not sure about the comment:
// good practice
...so behaves same as sort(), func name meaning as good practice
...so `console.log(sortByAge(arr));` works as expected
...so it can be chainable too

### DIFF
--- a/1-js/05-data-types/05-array-methods/8-sort-objects/solution.md
+++ b/1-js/05-data-types/05-array-methods/8-sort-objects/solution.md
@@ -1,6 +1,6 @@
 ```js run no-beautify
 function sortByAge(arr) {
-  arr.sort((a, b) => a.age - b.age);
+  return arr.sort((a, b) => a.age - b.age);  // return is not needed for this task, but it's ok to expect sortBy() behaves the same as sort()
 }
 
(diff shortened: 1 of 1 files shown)


iliakan:
It would add more confusion and an additional need to explain.

We also do not require the function to be chainable here. It's not an array method, after all, but a separate function.

joaquinelio:
ok!
It... still itches, naming & behavior are important no matter if you don't use it "right now"
but the need for extra explanation troubled me more.

PR #3009, opened 2022-05-14
✅ just add missed period and not only

Rnbsov:


iliakan:
Sorry, but are these changes really needed? =) They don't appear to add anything.

The current variant also isn't a mistake. Small letters are ok, I hope.

Rnbsov:
I did this based on the example above, to preserve the consistent style

image

PR #3012, opened 2022-05-16
Update about.md (change "data" field to "value")

AmirHasanzade:
The object returned from "await stream.read()" has two fields: "done" and "value", and not the "data" field destructured here.


joaquinelio:
cant we name them as we want?

AmirHasanzade:
To rename the "value" field to "data" you should write something like:
const { done, value: data } = result;
// Here we are destructuring the "value" field from "result" object and renaming it to "data"

iliakan:
Thank you! Good catch!

PR #3014, opened 2022-05-17
Using `__proto__` in object literals is not deprecated

bakkot:
The `__proto__` _syntax_ (as opposed to the getter/setter) was moved out of Annex B in https://github.com/tc39/ecma262/pull/2125. It's fine to use, and in fact is generally preferable to `Object.create()` (because it makes it easier to set up other normal data properties, and because it's syntax rather than an API call).

### DIFF
--- a/1-js/08-prototypes/04-prototype-methods/article.md
+++ b/1-js/08-prototypes/04-prototype-methods/article.md
@@ -3,15 +3,14 @@
 
 In the first chapter of this section, we mentioned that there are modern methods to setup a prototype.
 
-The `__proto__` is considered outdated and somewhat deprecated (in browser-only part of the JavaScript standard).
+Setting or reading the prototype with `obj.__proto__` is considered outdated and somewhat deprecated (in browser-only part of the JavaScript standard).
 
 The modern methods are:
 
-- [Object.create(proto, [descriptors])](mdn:js/Object/create) -- creates an empty object with given `proto` as `[[Prototype]]` and optional property descriptors.
 - [Object.getPrototypeOf(obj)](mdn:js/Object/getPrototypeOf) -- returns the `[[Prototype]]` of `obj`.
 - [Object.setPrototypeOf(obj, proto)](mdn:js/Object/setPrototypeOf) -- sets the `[[Prototype]]` of `obj` to `proto`.
 
-These should be used instead of `__proto__`.
+These should be used instead of `.__proto__`.
 
 For instance:
 
@@ -22,7 +21,9 @@ let animal = {
 
 // create a new object with animal as a prototype
 *!*
-let rabbit = Object.create(animal);
+let rabbit = {
+  __proto__: animal,
+};
 */!*
 
 alert(rabbit.eats); // true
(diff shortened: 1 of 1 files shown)


smith558:
@bakkot Is this not the consensus of the discussion you are referring to? 

image

bakkot:
@smith558 as that screenshot says, the use of the `__proto__` _syntax_ is not discouraged. It's only the use of the `__proto__` _accessor_ which is problematic.

smith558:
> @smith558 as that screenshot says, the use of the `__proto__` _syntax_ is not discouraged. It's only the use of the `__proto__` _accessor_ which is problematic.

Yeah, Could you explain to me what they mean by "proto syntax"? I understand the other 2 points but for me, those would be included under "proto syntax".

bakkot:
`{ __proto__: null }` is [special syntax](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-runtime-semantics-propertydefinitionevaluation) (search for "isProtoSetter" in that section); it's a different thing than `{ foo: null }`. That specific syntax is given meaning by the spec.

By contrast, `obj.__proto__` is a normal property access, just like `obj.foo`. But there happens to be a getter/setter named `__proto__` on `Object.prototype`, so if `obj` inherits from `Object.prototype` then this triggers the behavior of that getter/setter (just like how every other getter/setter works).

1-js/08-prototypes/04-prototype-methods/article.md
```diff
@@ -3,15 +3,14 @@
...
+Setting or reading the prototype with `obj.__proto__` is considered outdated and somewhat deprecated (in browser-only part of the JavaScript standard).
 
 The modern methods are:
 
-- [Object.create(proto, [descriptors])](mdn:js/Object/create) -- creates an empty object with given `proto` as `[[Prototype]]` and optional property descriptors.
 - [Object.getPrototypeOf(obj)](mdn:js/Object/getPrototypeOf) -- returns the `[[Prototype]]` of `obj`.
 - [Object.setPrototypeOf(obj, proto)](mdn:js/Object/setPrototypeOf) -- sets the `[[Prototype]]` of `obj` to `proto`.
 
-These should be used instead of `__proto__`.
+These should be used instead of `.__proto__`.
```
smith558, line 13: Change to `obj.__proto__`? Clashes with above paragraph when it's referenced as `obj.__proto__` rather.

smith558:
> `{ __proto__: null }` is [special syntax](https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-runtime-semantics-propertydefinitionevaluation) (search for "isProtoSetter" in that section); it's a different thing than `{ foo: null }`. That specific syntax is given meaning by the spec.
> 
> By contrast, `obj.__proto__` is a normal property access, just like `obj.foo`. But there happens to be a getter/setter named `__proto__` on `Object.prototype`, so if `obj` inherits from `Object.prototype` then this triggers the behavior of that getter/setter (just like how every other getter/setter works).

I see. Thanks! I feel like Objects and `this` are very important concepts and a first-time reader may feel confused by seeing suggestions to "use" `__proto__` and "not use" `__proto__` (tho not in the same way) at once and the article changes proposed as it at this time do not fully convey the meaning and the differences in such way so as not to confuse the reader or teach them to use it correctly. (feedback)

iliakan:
You mean that using `__proto__` in `{...}` is ok, but get/set `.__proto__` is not, right?

iliakan:
@bakkot

PR #3025, opened 2022-05-23
👾 add clarifying example

Rnbsov:
I think it's worth adding this example, because the first time it confused me a little

I also think it's worth writing it down better somehow. Because the introduction of a new array in the example, which besides stand out of the whole article, is not good decide

Maybe it would look better:

```js
let arr = [1, 0, false, 1];

alert( arr.indexOf(0) ); // 1
alert( arr.indexOf(false) ); // 2
alert( arr.indexOf(null) ); // -1

// note that the lastIndexOf method looks for from the end, but index counted from the beginning
alert( arr.lastIndexOf(1)) // 3

alert( arr.includes(1) ); // true
```


iliakan:
What else we should change?

PR #3036, opened 2022-06-03
update article

binarydevelop:
In Number article the number of zero's count is 5 whereas it's written as 6 in the comment. Total number of digits should be 6 rather than zero's.


joaquinelio:
but   " 5, so naturally 1e-6 " does not make sense.  

I never liked the "number of  0s " explanation.  ( should I count the zero before the decimal point ?) 

I always thought, and explained it, as the **" number of spaces the decimal point moves"**  
because it's how it actually works.  

It's a universal explanation and you can forget the zeroes, it works with any digit.

iliakan:
I'm sorry, but `0.000001` has exactly 6 zeroes, not 5!

iliakan:
Added a few words about the decimal point move.

PR #3039, opened 2022-06-06
👾 smth

Rnbsov:
What about add it to the article ?

 Screenshot_4

### DIFF
--- a/7-animation/2-css-animations/article.md
+++ b/7-animation/2-css-animations/article.md
@@ -256,9 +256,18 @@ That's a "soft" variant for sure. If we put `y` values like `-99` and `99` then 
 
 But how do we make a Bezier curve for a specific task? There are many tools. For instance, we can do it on the site <http://cubic-bezier.com/>.
 
-Or you can do it right in the developer tools dashboard in Chrome:
+Or you can do it right in the developer tools dashboard (most browsers supports editing Bezier curve):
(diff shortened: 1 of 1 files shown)


joaquinelio:
if so,
check firefox too,
I think it was the first with those nice tools - I never tried

joaquinelio:
if so,
check firefox too,
I think it was the first with those nice tools - I never tried

Rnbsov:
Yes, this function is there.

I can even assume that she appeared there even earlier than in chrome

Rnbsov:
I can't check if it's in safari

So it makes sense to add information like:

> If you use Chrome, Firefox (or any other chromium based browser) except Safari. You can try it right in the developer tools dashboard:

joaquinelio:
ok!

about safari, we could ask here in this pr

I bet this ends in a more generic statement
"... main browsers dev tools"

Rnbsov:
> I bet this ends in a more generic statement "... main browsers dev tools"

Yeah, me too

iliakan:
Can you please change the text, make it more generic ("most browsers") and add hints how to reach this panel instead of a screenshot?

Rnbsov:
I left the screenshots unchanged, I think then need to make svgs out of them

Perhaps some adjustments in the text are still needed

PR #3044, opened 2022-06-09
docs: fix the syntax copy in splice() array method

sohamsshah:
Previous copy:
`arr.splice(start[, deleteCount, elem1, ..., elemN])`

Updated copy:
`arr.splice(start, deleteCount, elem1, ..., elemN)`

Referred:
[MDN Docs: Array.prototype.splice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice#syntax)

### DIFF
--- a/1-js/05-data-types/05-array-methods/article.md
+++ b/1-js/05-data-types/05-array-methods/article.md
@@ -41,7 +41,7 @@ The [arr.splice](mdn:js/Array/splice) method is a swiss army knife for arrays. I
 The syntax is:
 
 ```js
-arr.splice(start[, deleteCount, elem1, ..., elemN])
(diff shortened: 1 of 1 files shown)


joaquinelio:
I can understand it.
I'm ok with the use of [ ] as markers for optional (since 1980... wow)

But it may be confusing, and if you want precision it should be
f (a [, b [ , c [, ..., ]]] )
Not nice.

MDN is clear
```
splice(start)
splice(start, deleteCount)
splice(start, deleteCount, item1)
splice(start, deleteCount, item1, item2, itemN)
```

Not sure if I like the extended version everywhere.

iliakan:
[] markers should be ok. Some people not familiar =(

PR #3045, opened 2022-06-09
Update resume of check spam exercice

Geminii:
To avoid a misinterpretation about this exercice, I'm suggesting to put in lowercase the constraint about 'XXX'.

Looking closer at the solution, the function will be purposed at : `return lowerStr.includes('viagra') || lowerStr.includes('XXX');`. 
But if we do, the `checkSpam('free xxxxx')` can't be true.


iliakan:
It's important that XXX should be caught too.

PR #3046, opened 2022-06-09
typo?

joaquinelio:
I cant understand the last commit...


Rnbsov:
> I cant understand the last commit...

It seemed a little strange to me too

the word sane was perfectly suited there, but fine is also not bad

PR #3047, opened 2022-06-10
👾 just add missed space and as it turned out, not only

Rnbsov:
image

### DIFF
--- a/1-js/05-data-types/05-array-methods/article.md
+++ b/1-js/05-data-types/05-array-methods/article.md
@@ -510,7 +510,7 @@ The `split` method has an optional second numeric argument -- a limit on the arr
 ```js run
 let arr = 'Bilbo, Gandalf, Nazgul, Saruman'.split(', ', 2);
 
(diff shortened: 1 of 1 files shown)


joaquinelio:
exmple seems clear...
You may want to add spaces after comma too. line below.

Rnbsov:
> exmple seems clear... You may want to add spaces after comma too. line below.

not sure it neccesery

joaquinelio:
the comments show the actual output
"alert" doesnt show brackets, quotes nor spaces

Rnbsov:
what about the Numbers article?  There, when using the toFixed() method, there are quotes that emphasize that the method returns exactly the string 

image

joaquinelio:
Right!
Then I agree with the quotes now, but no with the brackets, the result is string.

iliakan:
Usually there are no quotes when we put output after //.

The `toFixed` is an exception, because it's important that there's a string there.

We can replace alerts by console.logs, but it's another story...

PR #3053, opened 2022-06-12
Fix link to MDN bitwise operator docs in (2.18)Javascript specials

nintendoit:
The permalink to bitwise operator was incorrect.The link before took to the correct page but not to bitwise operator docs subheading.


nintendoit:
closes #3053

PR #3061, opened 2022-06-17
Clarify finally only "passing through"

smith558:


iliakan:
This looks kind of obvious. Isn't it?

smith558:
> This looks kind of obvious. Isn't it?

Not from the article as it is right now. If consulting other sources, yes.

PR #3068, opened 2022-06-19
👾 update link

Rnbsov:


iliakan:
Do extra line breaks affect the site in any way? Why do you introduce them?

Rnbsov:
I don't know, I think it's all because of my perfectionism

iliakan:
Please don't add line breaks. Remove your PRs.

This is maybe a bit better, but the problem is that all such changes propagate to all language translations.

Translators manually examine each change, they just won't like extra work without any added value.

PR #3069, opened 2022-06-19
Fix language errors and fix broken link

peterhauke:
differ should be differentiate

https://developer.mozilla.org/en-US/search is broken so I changed it to  https://developer.mozilla.org/en-US/

Other language changes are to improve readability and correct minor language errors or improve punctuation


iliakan:
Thank you! Much better wordings here and there.

PR #3071, opened 2022-06-20
Fix language errors and improve style

peterhauke:
Technologies don't really "play" together, but they can "work" together.
It is best to avoid words such as "crippled", when "restricted" or "limited" are perfectly acceptable.
Change HTTP to HTTPS in the URLs in the paragraph above the image as HTTPS is in the image. I thought of changing anysite.com to javascript.info in this paragraph to match the image, but I decided not to. I may raise this later as a PR for you to decide.

### DIFF
--- a/1-js/01-getting-started/1-intro/article.md
+++ b/1-js/01-getting-started/1-intro/article.md
@@ -1,41 +1,41 @@
 # An Introduction to JavaScript
 
-Let's see what's so special about JavaScript, what we can achieve with it, and what other technologies play well with it.
+What's so special about JavaScript, what can we achieve with it, and what other technologies work well with it.
 
 ## What is JavaScript?
 
 *JavaScript* was initially created to "make web pages alive".
 
-The programs in this language are called *scripts*. They can be written right in a web page's HTML and run automatically as the page loads.
+The programs in this language are called *scripts*. They can be written within a web page's HTML and run automatically as the page loads.
 
 Scripts are provided and executed as plain text. They don't need special preparation or compilation to run.
 
 In this aspect, JavaScript is very different from another language called [Java](https://en.wikipedia.org/wiki/Java_(programming_language)).
 
 ```smart header="Why is it called <u>Java</u>Script?"
-When JavaScript was created, it initially had another name: "LiveScript". But Java was very popular at that time, so it was decided that positioning a new language as a "younger brother" of Java would help.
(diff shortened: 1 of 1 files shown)


peterhauke:
Thanks, @joaquinelio 

Searching for "avoid ableist language" in google brings up many reasons why using "crippled" is not recommended in English.

Apart from that, I am not sure that the word "crippled" would properly describe the situation here. Nothing is malfunctioning or is being damaged. So, we could make a case that "crippled" is wrong in this particular case anyway, from a purely technical point of view.

Are you saying that it is easier to translate read/update/delete into another language? That's interesting.

joaquinelio:
> Searching for "avoid ableist language" in google brings up many reasons why using "crippled" is not recommended in English.
> 
> Apart from that, I am not sure that the word "crippled" would properly describe the situation here. Nothing is malfunctioning or is being damaged. So, we could make a case that "crippled" is wrong in this particular case anyway, from a purely technical point of view.
> 

Just read, thanks!
Two strong arguments, for any language.

> Are you saying that it is easier to translate read/update/delete into another language when / is used? That's interesting.

It IS hard to read as a normal reader.
I still fix it when it doesn't fit, Spanish doesn't allow me to translate it that way.
But when it is related to explaining code or structs, sometimes it makes it easier for me programmer to catch. 
so I treat it as a special markdown to feel less guilty.
Just like to avoid writing "The force equals the mass times the acceleration", I'll never understand that way.

iliakan:
And what's the wrong with the "Let's see" introductory phrase?

peterhauke:
English is very forgiving and most people will understand what "technologies play well" means in this context, that's why I wrote 'Technologies don't really "play" together'.

"what's so special about JavaScript" is a very nice phrase which at the same time questions the importance of JavaScript and suggests that some people don't think it is important, at least that's the way that I understand it. "Let's see" at the beginning just weakens the strength of that message, but it is not wrong. Swapping "we can" in the second phrase becomes necessary if "Let's see" is removed.

I suggest that we just close this PR. When I have some time, I will read it again and correct just the language errors and not try to improve the style.

iliakan:
Dear @peterhauke,

As a person whose English is non-native, I surely lack the understanding of intricate style aspects and undertones of the language.

Although, I'm trying to learn the reason of changes before blindly accepting them.

Thank you very much for your improvements! ❤️

PR #3077, opened 2022-06-21
WebSocket, some English grammar

joaquinelio:


joaquinelio:
...from a non-native :(  Not sure about all the changes

Love these articles, but the lack of some pronouns and articles makes it unnecessarily harder to read

line 3
still dont like it, but the original sounds odd:
`...without breaking the connection and additional HTTP-requests`
like 
...The new model, without spots and additional colors.
The ambiguity is resolved not by context but from the reader's knowledge of how http works.
so the proposed
`...without breaking the connection and (((implicit without?))) the need of additional HTTP-requests.
`
or 
`...without breaking the connection, thus the need of additional HTTP-requests.
`

joaquinelio:
line 324
I think it's "set" or "set up" (verb) instead of "setup" (noun)

still issues with commas, pr s rejected before. 
"Not now but tomorrow" 
different grammar than 
"Not now, but a with a good reason"

PR #3078, opened 2022-06-22
"and" instead of "/" just to match previous line

joaquinelio:


joaquinelio:
This time is ok to put "respectively" 
but I dont know where.

PR #3079, opened 2022-06-22
Improve English language/correct grammar

peterhauke:
In the first paragraph, both of the following are correct:

you'll need other resources.
you'll need access to other sources.

I propose the first version

There is no hyphen in In-between, so you could write here: in between or between - but between is by far better.


iliakan:
Are these changes really needed, I wonder?

iliakan:
Are these changes really needed, I wonder?

Alexandre887 (approved):

iliakan:
Thanks!

PR #3081, opened 2022-06-23
Fixed typo

josephharkins:
Fixed 'referrer' typo


josephharkins:
Hm. I fixed a typo but it appears to be a typo in the actual language: https://en.wikipedia.org/wiki/HTTP_referer

But there is no note on the tutorial page and it mixes the different spellings, could lead to confusion (like me)

joaquinelio:
did a quick search too, found that's a known typo
maybe they decided to invent a new word to save one letter. 
  
"HTTP REFERER", header as described
but
"document.referrer", prop that we use
so,  
referrer.

iliakan:
Yeah, known glitch. Not only me who made this error! 🥳

PR #3083, opened 2022-06-26
promise minor grammar suggestions

joaquinelio:


joaquinelio:
everyones fav I won't dare suggest style.
should I keep doing this kind of minor changes?

maybe you feel forced to merge if they are correct 
but you do it... reluctant bcause they  are irrelevant

iliakan:
These changes are very important, as my language is somewhat lacking in this aspect. And the correct language makes the book easier to read. 

Thank you very much for suggesting them @joaquinelio!

PR #3086, opened 2022-06-28
👾 fix typo

Rnbsov:


iliakan:
The first change is definitely wrong: "prototypal inheritance" is correct.

The 2nd one - not sure. Closing for now, but maybe 2nd change is good...

PR #3088, opened 2022-07-01
fix: 1-js/06-advanced-functions/05-global-object

printfinn:
This sounds weird, maybe we should change the order?


joaquinelio:
👍 
more versatile langs sound ok here,
but I think you're right

PR #3093, opened 2022-07-10
👾 fix typo

Rnbsov:
I believe that the word `error ` should be in the singular

image


iliakan:
"errors" means something like "exits with an error"

PR #3098, opened 2022-07-12
Removed Atom (Sunsetted) From List Of Code Editors

### LINKED ISSUE #3097: Remove Atom (Sunsetted) From Recommended Code Editors
While Atom was a useful and popular lightweight code editor for many years, GitHub have announced that they are sunsetting Atom on December 15, 2022 ([link](https://github.blog/2022-06-08-sunsetting-atom/)). Given a key audience of the code editors page (1-js/01-getting-started/3-code-editors/article.md) are new programmers, it may reduce confusion and the chance that new programmers will start their journey using a deprecated code editor. 

Therefore, Atom should be removed from the list of recommended code editors.

nayanadasgupta:
Resolves #3097 by removing Atom (sunsetted) from the list of recommended code editors.

### DIFF
--- a/1-js/01-getting-started/3-code-editors/article.md
+++ b/1-js/01-getting-started/3-code-editors/article.md
@@ -31,7 +31,6 @@ In practice, lightweight editors may have a lot of plugins including directory-l
 
 The following options deserve your attention:
 
-- [Atom](https://atom.io/) (cross-platform, free).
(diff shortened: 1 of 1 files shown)


iliakan:
These are all somewhat doubtful choices. I stick to VS code mostly.

PR #3106, opened 2022-07-16
Update article.md

himalayasingh:
Added the missing "use strict" directive in the snippet.


iliakan:
It's by default everywhere.

PR #3111, opened 2022-07-19
Adding clarification for undefined arguments

JeremyIglehart:
Consider the following code:

```javascript
let someFunction = (someString = "nothing passed") => console.log(someString);

someFunction("hello") // hello
someFunction() // nothing passed
someFunction(undefined) // nothing passed
someFunction(null) // null
someFunction(0) // 0
someFunction((() => {return undefined})()) // nothing passed

/* We learned here that default values are assigned to anything undefined,
   not just parameters that do not receive a value.  If you pass in undefined, you're going to get
   a default assignment as per the parameter definition
*/
```

I'd like to clarify for future readers that even any argument getting passed in that evaluates as `undefined` we'll receive the default value - not *just* parameters/arguments that are not passed.  If you want `undefined` to pass through to your function, you need to take this into account when designing your function signature.

### DIFF
--- a/1-js/02-first-steps/15-function-basics/article.md
+++ b/1-js/02-first-steps/15-function-basics/article.md
@@ -206,7 +206,7 @@ function showMessage(from, *!*text = "no text given"*/!*) {
 showMessage("Ann"); // Ann: no text given
 ```
 
-Now if the `text` parameter is not passed, it will get the value `"no text given"`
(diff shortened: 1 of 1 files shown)


joaquinelio:
You're right. Maybe it worths the mention.

But.
I never liked undefined type. 
Its only purpose (I think) is to detect if a parameter was provided to the function,
you shouldnt assign "undefined", 
"null" is meant for that.

just my opinion

PR #3112, opened 2022-07-19
Fixed Typo in line 192

Kurineko:
AND returns the first falsly or the last truthy, or so i've learned in the text above, I'm sure this is a typo

### DIFF
--- a/1-js/02-first-steps/11-logical-operators/article.md
+++ b/1-js/02-first-steps/11-logical-operators/article.md
@@ -189,7 +189,7 @@ The AND `&&` operator does the following:
 
 In other words, AND returns the first falsy value or the last value if none were found.
 
-The rules above are similar to OR. The difference is that AND returns the first *falsy* value while OR returns the first *truthy* one.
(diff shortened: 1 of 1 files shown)


Kurineko:
im not sure if this is how you push for a change, how do i know if its done correctly?

joaquinelio:
first.

true || unknown

If the first is true, it won't bother in evaluate the second

Kurineko:
As far as i understand. That would be for the OR, but here it is talking about the AND. they give this as an example: alert( 1 && 2 && null && 3 ); // null, the answer is null because it finds the first falsy one, if it stopped at the first truthy one it would stop at 1. that said if the example was alert( 1 && 2 && 5 && 3 );, it would give the number 3 out, because none are falsy it gives the _last_ truthy, which  is the change i made above. I hope im not being a nuisance, since im just learning this for the first time, but as far as i can tell this is correct

Kurineko:
Nevermind, i realized that it ment to say OR as the comparison, not "or" the word, in that case i would suggest repharsing that in a more clear way if possible as it can be confusing. thanks for taking your time in answering my... Confusion

PR #3115, opened 2022-07-21
Update article.md

Archsx:
for the example about `cache.js` using `WeakMap`, the statement `let result = /* calculate the result for  */ obj;` is the same as `let result = obj` , if so , the weakmap would use `obj` as key and **value** at the same time , if you set `obj` to `null` later, the weakmap won't remove the key 'obj'. just want to make the example more clear.


iliakan:
Thanks, but I hope it's kind of obvious what we mean here =)

PR #3123, opened 2022-07-29
add another solution

webFardin:
this solution is unrelated to the article, but i think it can be helpful


iliakan:
not sure how that'll react on `<style-me>` or `<style_me>`

PR #3137, opened 2022-08-08
Update article.md

kai-nguyen-aligent:
Fix a minor typo at line 181

### DIFF
--- a/1-js/05-data-types/12-json/article.md
+++ b/1-js/05-data-types/12-json/article.md
@@ -178,7 +178,7 @@ Here, the conversion fails, because of circular reference: `room.occupiedBy` ref
 The full syntax of `JSON.stringify` is:
 
 ```js
-let json = JSON.stringify(value[, replacer, space])
+let json = JSON.stringify(value, [replacer, space])
 ```
 
 value


joaquinelio:
not a typo

but a comment can be added:  
... // the  [ ]  means optional here

kai-nguyen-aligent:
I understand that the parameters in side [ ] are optional. However, the first comma should be out side of the square bracket as it's used for separate the parameters. Therefore, I still think this is a typo. I took references from this Wiki: https://en.wikipedia.org/wiki/Usage_message#:~:text=To%20indicate%20optional%20arguments%2C%20Square,grouping%20conventions%20as%20square%20brackets.

My apology for the unclear pull request description.

joaquinelio:
The comma is part of the optional part, if you dont use the 2nd arg, you dont need it.

mm...
I was comfortable with  [, optional ] for nearly half of century
but today  you can pass array params this way, it may be confusing
I can understand why MDN avoid it

kai-nguyen-aligent:
In Function binding/Partial functions (https://javascript.info/bind#partial-functions), I saw a slightly different way to interpret optional parameters (which look clearer for me):

`let bound = func.bind(context, [arg1], [arg2], ...);`

joaquinelio:
right !
looks cleaner. Maybe you are right

it's a paradox, it looks cleaner,
but now if you omit the param leaving the comma,
not an error but looks dirty 
just kidding

iliakan:
Maybe we need to officially decide on a single style to use everywhere =)

PR #3139, opened 2022-08-08
Correct the English and made minor improvements

peterhauke:
This is almost the minimal number of changes that I would make. 

I would still recommend changing "play" in the first line to "work".


joaquinelio:
👍 
Being non-native, I can understand only some of the corrections.
I also understand the lack of articles and pronouns, optional in other languages sounds odd in English. 

I guess 
"allows to create"  
can be fixed with "allows us"

iliakan:
Thanks!

PR #3148, opened 2022-08-14
Using the right word that obeys English concord

ChidubemAkpu:
I changed 'properties' on line 2 (after the main heading) of the article- 'property getters and setters', from 'properties' to'property' , so that it would be correct grammatically. The new modification obeys the English concord rule of singular verb for singular noun.


iliakan:
Maybe "There are two kinds of AN object property." ?

ChidubemAkpu:
Guess it should be 'The first kind is data property'

PR #3149, opened 2022-08-16
semántic

joaquinelio:
I think
to assign is to put (something under responsibility of, for example)

so you assign an object to a variable,
not a variable to an object 

"storing" is redundant but I don't care
"holding"  an object?
storing twice is stronger

### DIFF
--- a/1-js/04-object-basics/02-object-copy/article.md
+++ b/1-js/04-object-basics/02-object-copy/article.md
@@ -21,7 +21,7 @@ Quite an obvious result, right?
 
 Objects are not like that.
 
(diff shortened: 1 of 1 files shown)


iliakan:
Are you sure we need to improve here? I can suggest other variants possible also, if you think the current wording is bad =)

joaquinelio:
NO
im not sure.

but spanish 'asignar' has similar meaning and I find it confusing there too. 
not too bad, it s undrstable anyway


well, you need native reviewers, not me.

iliakan:
Sorry, the current wording seems better :/

Issue #3150, opened 2022-08-16
No assign importance rating to this task: https://github.com/javascript-tutorial/en.javascript.info/blob/master/1-js/02-first-steps/17-arrow-functions-basics/1-rewrite-arrow/task.md


SUPERustam:
No assign importance rating to this task:

https://github.com/javascript-tutorial/en.javascript.info/blob/master/1-js/02-first-steps/17-arrow-functions-basics/1-rewrite-arrow/task.md

PR #3152, opened 2022-08-16
Solved #3150. Completely full code block with arrow functions in section https://javascript.info/arrow-functions-basics

SUPERustam:
- Completely full code block with arrow functions in section https://javascript.info/arrow-functions-basics

- Solved #3150.
In my opinion, this task should have importance rating 5, because in similar task https://javascript.info/task/rewrite-switch-if-else  has also importance rating 5


iliakan:
You rewrite the Function Declaration also =) Not as requested.

PR #3164, opened 2022-08-26
Update solution.md

khldon123:
the "str" variable in the solution doesn't have the same value as the task, so i edited the "str" value for the solution and based on that i added the `\b` so we can get only the correct colors
Cont2


khldon123:
update solution file

iliakan:
explained below in the very same solution already =)

PR #3165, opened 2022-08-27
Mention of the `BigInt` data type in the article "Basic operators, maths"

Alexandre887:


iliakan:
Move this to Bigint instead, please.

We shouldn't mention BigInt in this chapter.

Alexandre887:
> We shouldn't mention BigInt in this chapter.

Since the article describes how a unary plus works, why don't we write about its feature concerning the BigInt data type? All the more, the statement that unary plus does the same thing as `Number(...)` is incorrect in this case and might mislead people reading this article. In my opinion, a little insertion in the article about `BigInt` shouldn't hurt.

> Move this to Bigint instead, please.

The article `BigInt' already has a short description of this feature:
image
Do you mean that I should just add more information to that part (as in my PR)?

iliakan:
If `BigInt` already has that info, then just don't add it to Numbers.

The article is about numbers, not bigints. And the behavior of "+" with numbers.

Also, because `BigInt` is a rarely used feature, we'd better not confuse readers with the info they won't need.

Alexandre887:
> If `BigInt` already has that info, then just don't add it to Numbers.
> 
> The article is about numbers, not bigints. And the behavior of "+" with numbers.

I was going to add this information to "Basic operators, maths" instead of "Number", but whatever.
Okay, in that case, I'll close this PR.

PR #3175, opened 2022-09-06
Update article.md

davidshq:
Minor changes to grammar

### DIFF
--- a/2-ui/2-events/01-introduction-browser-events/article.md
+++ b/2-ui/2-events/01-introduction-browser-events/article.md
@@ -195,7 +195,7 @@ Assign a handler to `elem.onclick`, not `elem.ONCLICK`, because DOM properties a
 
 ## addEventListener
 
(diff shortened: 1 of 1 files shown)


iliakan:
Please explain, is the current grammar wrong or bad?

Also, mentioning "the problem" is a bit clearer than mentioning "constraints", isn't it?

davidshq:
@iliakan - It's a balance between readability and clarity. Problem is a simpler word, constraint more clearly states the type of problem. That said, I don't have a strong opinion on this. If you want to retain problem, I'd change the wording to:

"which do not the same problems as event handlers."

"They are free of such a problem" isn't clear as to what problem exactly is present, the above clarifies while retaining "problems" to maintain readability.

iliakan:
Merged, replaced the last phrase with *Developers of web standards understood that long ago and suggested an alternative way of managing handlers using the special methods `addEventListener` and `removeEventListener` which aren't bound by such constraint.`*

Thanks!

PR #3179, opened 2022-09-08
Fix syntax error in object-copy article

SpliceReborn:
The square brackets in 'Object.assign(dest, src1[, src2, src3...])' 
results in a syntax error. The correct syntax should be  
'Object.assign(dest, src1, src2, src3, ...)', but I've added a srcN, changing it 
to 'Object.assign(dest, src1, src2, src3, ..., srcN)' to make it more relevant 
to the explanation found underneath the original code snippet.

### DIFF
--- a/1-js/04-object-basics/02-object-copy/article.md
+++ b/1-js/04-object-basics/02-object-copy/article.md
@@ -160,7 +160,7 @@ We can also use the method [Object.assign](https://developer.mozilla.org/en-US/d
 The syntax is:
 
 ```js
-Object.assign(dest, src1[, src2, src3...])
+Object.assign(dest, src1, src2, src3, ..., srcN)
 ```
 
 - The first argument `dest` is a target object.


iliakan:
How shall we tell people that `[...]` is a special "cross-language" syntax for optional arguments?

joaquinelio:
[ ] ... 
play with size, font, color,
add as md.
color wont do well as editors highlight syntax that way,
but, maybe, grey and a little oversize will tell the brain there is smthing out of place

or, 
do submit and use mdn way

iliakan:
You mean `Object.assign(obj, ...sources)`? May be a better choice, indeed! ;)

joaquinelio:
oh!
It's an idea, a font that means is meta code
far easier to implement.

I meant individual format 
**`Object.assign(dest, src1`** [ **`, src2, src3`**...] **`)`**
(well, I tricked it, the format should be in the brackets graying them, not the code)
**Object.assign( dest, src1** `[` **, src2, src3**`...] ` **)**

iliakan:
even if we make the brackets look different, people who don't know what they mean will still be confused.

joaquinelio:
yep
more and more I'm convincing myself the [ ] is not a friendly tool for newcomers

PR #3180, opened 2022-09-09
Update article.md

gleachkr:
Fixed a tiny typo.

### DIFF
--- a/1-js/05-data-types/07-map-set/article.md
+++ b/1-js/05-data-types/07-map-set/article.md
@@ -291,7 +291,7 @@ set.forEach((value, valueAgain, set) => {
 
 Note the funny thing. The callback function passed in `forEach` has 3 arguments: a `value`, then *the same value* `valueAgain`, and then the target object. Indeed, the same value appears in the arguments twice.
 
-That's for compatibility with `Map` where the callback passed `forEach` has three arguments. Looks a bit strange, for sure. But may help to replace `Map` with `Set` in certain cases with ease, and vice versa.
+That's for compatibility with `Map` where the callback passed `forEach` has three arguments. Looks a bit strange, for sure. But this convention may help to replace `Map` with `Set` in certain cases with ease, and vice versa.
 
 The same methods `Map` has for iterators are also supported:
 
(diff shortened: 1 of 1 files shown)


joaquinelio:
👍

gleachkr:
Yeah, "this convention" might be a little stuffy. What about a slightly bigger edit:

> But following this pattern helps us replace `Map` with `Set` in certain cases with ease, and vice versa.

joaquinelio:
It's ok anyway, it doesn't need to be perfect.

mm...
if we overthink about it we can lose the original idea (it may sound dubious on purpose, like making excuses). so I prefer the first fix.

But now I am overthinking.

joaquinelio:
I'm not native English, still learning.
Without the fix, it sounds like a suggestion; am I right?

gleachkr:
Hm... it doesn't quite sound like a suggestion to me. It's just saying that the callback having three arguments is a design choice that could have been made differently, but it has good reasons behind it. 

But if putting it this way makes it hard to understand, then again, maybe that's not the best formulation. Just dropping the original edit to 

> That's for compatibility with `Map` where the callback passed `forEach` has three arguments. Looks a bit strange, for sure. But this may help to replace `Map` with `Set` in certain cases with ease, and vice versa.

 would be OK too.

There's also maybe another tiny grammar nitpick for this paragraph. I think "the callback passed `forEach`" should be "the callback passed to `forEach`". I can do a separate PR or fold it into this one, if you agree with the edit.

joaquinelio:
I like the added "this" alone, 
`But this may help to replace`


sorry for
"Without the fix, it sounds like a suggestion; am I right?"
nothing to do with the repo, 
tldr
it was my interpretation of the original, not what the author intended to say.
`But may help to replace (...)` 
sounded to me like
`But replacing (...) may help`
I wasnt sure if was bad grammar or just a wrong way to put it.

gleachkr:
> `But may help to replace (...)`
> sounded to me like
> `But replacing (...) may help`
> I wasnt sure if was bad grammar or just a wrong way to put it.

Oh, I see! Yeah, to me, without the fix, the original just feels like a typo. It's definitely ungrammatical as is.

joaquinelio:
I'm not the one to say 
sorry, maybe the first fix was ok

"passed to" is needed too I think

And thanks 👍 for the explanation.

PR #3183, opened 2022-09-09
Update "Import *" section

Hasan-Mir:
Despite of what you said about `import * as <obj>`, if we use `import * as <obj>` when that 3rd-party library uses ES modules, Webpack is still doing tree-shaking and I think the first reason is redundant and can be removed because when that 3rd-party library uses commonJS module system, using `import {item1, item2} from module` will load the whole module and won't be making the build smaller.

https://github.com/webpack/webpack/tree/main/examples/harmony-unused#examplejs


iliakan:
Indeed, thank you! Merged with minor changes.

PR #3189, opened 2022-09-18
MOD: Use strict equality comparison (line 16)

rmarasigan:
if (password == "rockstar") ok() -> if (password === "rockstar") ok()


iliakan:
Is there a practical difference?

rmarasigan:
The "===" uses a strict comparison where it does not do any conversion. It does the same job, the only difference is that the "==" converts it first before performing the comparison.

iliakan:
Closing, as there seems to be no practical difference *in this case*.

PR #3191, opened 2022-09-19
docs: change `let` to `const`

YangFong:


joaquinelio:
mmm...
I didnt test it, but you should... :)

YangFong:
There's no secondary assignment, so it's not a problem.

It looks like I've changed all the `let`, but for `for...i`, I found that can not be modified, and adjusted back.

joaquinelio:
wow
my bad
it works
`for (const i = 1; `
fails, but
`for (const i of k ` 
initializees i on every loop.
spec compliant i guess

still, confusing.

joaquinelio:
wow
my bad, it works
`for ( const i , i < nn `
fails, but 
`for ( const i of kk`  
initializes i every loop.
spec compliant I guess

still, confusing

YangFong:
If you delve into the matter, you really don't know why it's enforceable. 

> The following is just my personal understanding of the logic.

I might think that the whole process, each iteration is performed in a different block scope: 

```js
// for..of
{
  {  
    const item = next();
    // ...
  }
  {
    const item = next();
    // ...
  }
}
```

Why is this so? Because `for...of` is not like `for... i`, each update, are in the original basis changes. Instead, it goes directly from one value to the other, without any connection, or correlation, between the two.

joaquinelio:
Ok, 
forloop special treatment.
Nextables dont need the context env an i++ requires

thanks

joaquinelio:
but I dont like it.
🤣

joaquinelio:
But agreed, use const wherever possible

YangFong:
This tweak is not a bug fix, it's a programming style tweak, it's optional, I'm just suggesting a style I like.

iliakan:
We use `let` in most places of the tutorial, for brewity, even though `const` is more suitable.

PR #3196, opened 2022-09-24
Update article.md

Alexandre887:


joaquinelio:
mm. 
Unicode is more important than ###. 
In fact, It deserves its own article.

Alexandre887:
> mm. Unicode is more important than ###. In fact, It deserves its own article.
> 
> " ### " dont get listed on the left panel, do they? didnt check

There is a separate section called "Internals, Unicode" which appears in the left pane and is the header for such subsections as "Surrogate pairs", "Diacritical marks and normalization", and a **new subsection called "Unicode Characters"**. "Unicode Characters" should therefore be labeled ### — your suggestion for a new article is completely irrelevant here

> In fact, It deserves its own article.

If the author of the textbook or other enthusiasts decide to send a PR adding a new article, it will appear in the textbook.

joaquinelio:
right!  
so, internals is the section, ok 👍 

I make irrelevant comments all the time, sorry for that

I'm translating the new "strings" and I'm a little tired,
a separate misc/ unicode+internals would look cleaner, the comment was a subtle suggestion to Ilya

Alexandre887:
@joaquinelio 

It's okay, I understand you. I am the maintainer of the Russian version of the tutorial, and I also constantly have to do a full review of Pull Requests, which add edits from the English version to the Russian version (new articles or edits to existing articles). Sometimes I even have to translate the content myself.

> a separate misc/ unicode+internals would look cleaner

I agree with you here. In general, in my opinion, the article "strings" needs a little revision, because after all, the tutorial is not a documentation, and a full description of some complex language features in the usual articles (outside the "miscellaneous" section) in some places is not quite appropriate.

 In my opinion, most of the sections in the articles labeled "Advanced knowledge" should be moved to separate articles, under "miscellaneous" (as was already done, for example, with "reference data type" from the article "Object references and copying")

iliakan:
Would you suggest to move the whole Unicode section to Miscellaneous?

Alexandre887:
> Would you suggest to move the whole Unicode section to Miscellaneous?

Yes, in principle, it can be done. There is enough content for a separate article. After that you can just specify the link to the new article, for example, in "Summary".

iliakan:
Should that be done? Tell your opinion ;)

romanstetsyk:
In my opinion, it's a good idea to move Unicode to a separate article.

iliakan:
Unicode is a separate article now.

Please take a look.

Alexandre887:
👍

PR #3198, opened 2022-09-24
fix small typo in web-components slots chapter.

skromez:
I don't see anyone using vice versa with "-" in between, so decided to remove it as well.


skromez:
also in this chapter it's "vice-versa" 
image
https://javascript.info/var

PR #3207, opened 2022-09-29
ECMA broken link

joaquinelio:


joaquinelio:
the original has a broken pdf link, I put the page
https://www.ecma-international.org/publications-and-standards/standards/ecma-402/

MAYBE the page link is safer than the new 
https://www.ecma-international.org/wp-content/uploads/ECMA-402_9th_edition_june_2022.pdf

the page has a big DOWNLOAD PDF button, so...

joaquinelio:
news:
the broken pdf link was due the lack of httpS: 
it could be fixed
https://www.ecma-international.org/ecma-402/1.0/ECMA-402.pdf 

still...

PR #3209, opened 2022-09-29
https link

joaquinelio:
most http links just redirect to https
others like this one dont, the browser shows ugly "not secure"

should we change them?


iliakan:
Yeah, no place for http in the modern world 👍

PR #3215, opened 2022-10-04
Add .at( ) to strings summary

joaquinelio:
1) thanks, I liked  at( ).

2) `at` alone seems odd but ` .at( ) ` doesnt fit with the neighbors `slice` `indexOf ` ...


joaquinelio:
I dont know what s different in line 522
I dont remember having edited it

PR #3216, opened 2022-10-04
Update dot notation description in article.md

Complexlity:


shallow-beach (approved):
helpful addition mirroring previous example

PR #3220, opened 2022-10-05
Unicode art, grammar suggestions

joaquinelio:


joaquinelio:
"chinese Hieroglyphic" ?
it seems it's specific for the ancient "sacred" Egyptian writting. 

maybe "logogram"

joaquinelio:
sorry, it should be 2 PRs.

I dont know if the examples were Han, (ideogram? logogram?), so I put the safe "character".
Chinese "Hieroglyphic" is wrong, it's specific for the ancient "sacred" Egyptian writting.

I added " ... (4 hex digits)"
because you start talking about hex, then jumped to bytes unit.

iliakan:
The wikipedia says: In a [written language](https://en.wikipedia.org/wiki/Written_language), a logogram, logograph, or lexigraph is a [written character](https://en.wikipedia.org/wiki/Written_character) that represents a [word](https://en.wikipedia.org/wiki/Word) or [morpheme](https://en.wikipedia.org/wiki/Morpheme). Chinese characters (pronounced [hanzi](https://en.wikipedia.org/wiki/Hanzi) in Mandarin, [kanji](https://en.wikipedia.org/wiki/Kanji) in Japanese, [hanja](https://en.wikipedia.org/wiki/Hanja) in Korean) are generally logograms, as are many [hieroglyphic](https://en.wikipedia.org/wiki/Egyptian_hieroglyphs) and [cuneiform](https://en.wikipedia.org/wiki/Cuneiform_script) characters.

So the word "hieroglyphic" is used here, maybe keep it? Much more person will understand this word, rather than the scientific "logographic" =)

joaquinelio:
yes, and they will understand wrong 
because they all are wrong

but...
logogram is odd.
 I'd use ideograph...
wtheck, you are right, the first jeroglifo stays. 
but the chinese hieroglyph shouldnt

iliakan:
Sorry I don't get it =)

Why `𩷶'` can't be called "a rare chinese hieroglyph"?

iliakan:
Oh, I see, that's the wrong word to describe them?

So we should say "logogram"...  Or character.

joaquinelio:
correct, hieroglyphics is ancient egyptian.

 but I liked to leave the first mention to it
everyone loves hieroglyphics

PR #3222, opened 2022-10-05
Add note for non-strict mode and change wording

romanstetsyk:
Checked grammar with Grammarly


romanstetsyk:
I'm not a native speaker; my level of English is C1 (advanced)

iliakan:
Thanks for removing "the", and hesitating about the other fix.

romanstetsyk:
You used "in strict mode" and "in non-strict mode" without "the" throughout all other articles. I tried to make it uniform

iliakan:
There should be no "the", thanks for that. =)

PR #3224, opened 2022-10-07
Update article.md

bruteF04C3:
Had a quick look at YDKJS, and some other resources on youtube and over the internet in general. Open for suggestions / discussion

Thanks !

### DIFF
--- a/1-js/01-getting-started/1-intro/article.md
+++ b/1-js/01-getting-started/1-intro/article.md
@@ -8,9 +8,11 @@ Let's see what's so special about JavaScript, what we can achieve with it, and w
 
 The programs in this language are called *scripts*. They can be written right in a web page's HTML and run automatically as the page loads.
 
-Scripts are provided and executed as plain text. They don't need special preparation or compilation to run.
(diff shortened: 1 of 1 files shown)


joaquinelio:
I dont know...
Point of view.

For me, a programmer,
compilation is the process I fire to obtain a code or pseudocode file.
And the effect that it refuses to compile if it encounters fatal errors is not minor.

One could say an interpreted one compiles commands as they appear,
but I dont think the philosophy help the learner to understand the lang better.

bruteF04C3:
> I dont know... Point of view.
> 
> For me, a programmer, compilation is the process I fire to obtain a code or pseudocode file. And the effect that it refuses to compile if it encounters fatal errors is not minor.
> 
> One could say an interpreted one compiles commands as they appear, but I dont think the philosophy help the learner to understand the lang better.

May not help the learner directly, but at a medium with such reach, this can definitely help reduce confusion.

iliakan:
The description is clear enough: a script doesn't need a preparation/compilation.

What engines do internally is their own business.

PR #3226, opened 2022-10-08
unreal keys

joaquinelio:
image

[Cmd] +[Opt] + [J]  instead of
[Cmd +Opt + J],  prettier and accurate

image
(individual keys, I didnt test darkmode) 
...
over 100  "key:" 
but (?=.`key:)(?=.*\+.`) only 20, changed manually.

It was fun but the pr looks silly.
drop it if you like


iliakan:
We have a unified way of showing keys combinations. They are formatted by our script.

Let me know if there are grounds to change it =)

joaquinelio:
nah

I like the format of single keys, it looks like an actual key.
The combination loses that feeling 
but it's very clear what it means anyway
silly PR

> image
>

PR #3227, opened 2022-10-08
add links

Rnbsov:


joaquinelio:
mdn: markdn worked nicely years ago,
now, if mdn is unstranslated, leads to page not found.
i prefer to leave full English path, easier for translators (me) to check & change

iliakan:
@joaquinelio I guess we should ditch `mdn://`, replace it everywhere?

joaquinelio:
Not pretty

image

well... "GOOD NEWS!" The user gets the link.
Not that enthusiastic news.

joaquinelio:
Yes.

1. I cant do an easy replace mdn: ---> http...
With Full path, I can check just inside the editor or github site

2. Not that bad if you leave Eng 
image

3. The best: Once inside MDN,  with the previous scroll youll leave your frustration to MDN.

Rnbsov:
Sooo, should I replace every mdn: with long ones?

iliakan:
> 3. The best: Once inside MDN,  with the previous scroll youll leave your frustration to MDN.

What's the #3? ;)

iliakan:
ping @joaquinelio

joaquinelio:
sorry... grammar again
and the mad old man I'm growing inside me was just kidding
they weren't variants, they were the enumeration of advantages full path gives when 

no MDN transl is available 

3)   !!!!!!!!!!!!!!!
MDN: User is frustrated because `javascript.info` leads to 404.  
FULL:  User is frustrated because `MDN` doesn't show his/her language. 

so, don't take me always too seriously.
Still useful anyway.

iliakan:
Guess, the only clean way is to replace mdn: with full urls.

PR #3228, opened 2022-10-09
Update article.md

jivendra:
Minor mistake in the article.


iliakan:
The original order has another meaning.

PR #3229, opened 2022-10-10
Improve refences to non-Latin letters

danshil:
The original article speaks of cyrillic [sic] letters and hieroglyphs, but uses Cyrillic letters and Hanzi as its example. Changes create consistency between variables and reference terminology.


joaquinelio:
Uppercase Cyrillic

joaquinelio:
use Cyrillic letters
odd, I know. It seems like an adjective, 
but from Grammarly com:
My mother is British, and my father is Dutch.

PR #3231, opened 2022-10-11
Content-Length is now a CORS-safelisted response header

wdscxsj:
See [MDN](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header) and [whatwg/fetch](https://github.com/whatwg/fetch/pull/626).


iliakan:
Hoorray!!!!

PR #3236, opened 2022-10-13
Refactor function subscribe()

wdscxsj:
Refactor function `subscribe()`; improve description of status 502, and mention status 408. `browser.js` could be edited accordingly.

### DIFF
--- a/5-network/10-long-polling/article.md
+++ b/5-network/10-long-polling/article.md
@@ -38,31 +38,31 @@ If the connection is lost, because of, say, a network error, the browser immedia
 A sketch of client-side `subscribe` function that makes long requests:
 
 ```js
-async function subscribe() {
-  let response = await fetch("/subscribe");
+async function subscribe(url) {
+  let response = await fetch(url);
 
-  if (response.status == 502) {
-    // Status 502 is a connection timeout error,
+  if(!response.ok) {
+    // Status 502 is a proxy error,
     // may happen when the connection was pending for too long,
-    // and the remote server or a proxy closed it
-    // let's reconnect
-    await subscribe();
-  } else if (response.status != 200) {
-    // An error - let's show it
-    showMessage(response.statusText);
-    // Reconnect in one second
-    await new Promise(resolve => setTimeout(resolve, 1000));
-    await subscribe();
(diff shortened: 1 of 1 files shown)


5-network/10-long-polling/article.md
```diff
@@ -38,31 +38,31 @@ If the connection is lost, because of, say, a network error, the browser immedia
 A sketch of client-side `subscribe` function that makes long requests:
 
 ```js
-async function subscribe() {
-  let response = await fetch("/subscribe");
+async function subscribe(url) {
+  let response = await fetch(url);
```
auchynnikau, line 42: Since we don't changing `response` it below, let's use `const`:

```suggestion
  const response = await fetch(url);
```

wdscxsj: Thanks! I suppose in a small scope like this, we don't have to be "const-correct" like what C++ taught me quite a few years ago. Other parts of the Tutorial seem to use the same strategy. The "When to Use const" note in David Flanagan's JavaScript: The Definitive Guide (7th edition, p. 54) is also a good read.

iliakan:
Thank you, but the `subscribe` implementation here is for demo purposes.

It will get massively refactored if someone needs to use it in a real app, because it needs to take into account various statuses, errors, etc.

E.g. 408 is quite rare. Most devs will never come across it, so does it make sense mentioning?

Other than that, I see you refactored out the `url` parameter. As said for demo purposes we can leave it simple. In practice, there'll be something else here anyway.

Is there anything else in this PR ? Actual bugixes, maybe?

wdscxsj:
No bug fixes, just that the original comment "Status 502 is a connection timeout error" (and the associated logic) could be improved.

AFAIK, [502](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) means the current server reports that it gets an invalid response from the upstream server, due to timeout or anything else. I propose to mention [408](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408) because it's the current server that determines a timeout, semantically as informative as 502.

But in the real world, the server may choose to return a generic error in these cases.

The code could be easier to follow if it's restructured like `if(!ok) ... else ...`, without mentioning any error codes.

PR #3238, opened 2022-10-14
http to https checked links

joaquinelio:


joaquinelio:
I didn't bother with wiki links as they redirect fast, 
I didn't touch any example code (they should, especially authentication examples)

Custom events article
`www.w3.org/TR/dom/#event` 
redirects (slowly) to whatwg.org,  
  
searching elem dom article
https://www.whatwg.org/specs/web-apps/current-work/#dom-window-nameditem
redirs to
https://html.spec.whatwg.org/#dom-window-nameditem
then hangs...
I took the current anchors there

PR #3239, opened 2022-10-14
Update article.md

bruteF04C3:
undefined and "not defined" are two different things, updated for the sake of clarity

### DIFF
--- a/1-js/02-first-steps/12-nullish-coalescing-operator/article.md
+++ b/1-js/02-first-steps/12-nullish-coalescing-operator/article.md
@@ -29,7 +29,7 @@ For example, here we show `user` if its value isn't `null/undefined`, otherwise 
 ```js run
 let user;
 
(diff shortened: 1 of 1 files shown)


joaquinelio:
I like the original
Not defined by the coder, so js assigns undefined 

Is  the only purpose of undefined type I hate

bruteF04C3:
> I like the original Not defined by the coder, so js assigns undefined
> 
> Is the only purpose of undefined type I hate

Don't get me wrong but I did not completely understand what you said. 

Since "Not defined" is a reference error as well,
here this term might create confusion for some people. Also, undefined is clearly the suitable choice here.

But again, I am open to any debate over the same

Thanks !

joaquinelio:
oh, got it, "not defined" is a common error msg
yes, it can be rewritten

sorry I have to work on my english grammar, I meant
1 I liked "not defined" as "the programmer didn't specify "user" 
2 The  "I hate undefined type"  is just garbage I usually drop.

PR #3246, opened 2022-10-20
Proper Polish language inflection

f6p:
Hello world! properly inflects as Witaj świecie! in Polish.

Cheers
🍷


iliakan:
Thank you!

PR #3247, opened 2022-10-20
Add .at( ) to strings summary

joaquinelio:


joaquinelio:
w t heck again

edited inside github, I didnt change last line
maybe it lacked eol or something

PR #3256, opened 2022-11-01
Edit article.md (127)

victorglowin:


Alexandre887:
No, in that case `John` will be displayed. 

> If a same-named variable is declared inside the function then it *shadows* the outer one. The outer one is ignored.

Please close this Pull Request.

PR #3258, opened 2022-11-04
Irrelevant info about popup blocking

Alexandre887:
# [Popups and window methods — Popup blocking](https://javascript.info/popup-windows#popup-blocking)

image

^^ This information is not relevant at this time. Starting from version **95.0**, Firefox no longer blocks popups that open with a delay of 3000 milliseconds. I personally tested it using BrowserStack.


iliakan:
Looks like they removed it.

PR #3265, opened 2022-11-14
map.delete removes the pair key/value

joaquinelio:


joaquinelio:
just in case.
One may think it leaves the key untouched.

PR #3268, opened 2022-11-18
Optional chaining edits

Abidi-S:
Fixed minor typos, and some phrasing changes to improve readability :)

### DIFF
--- a/1-js/04-object-basics/07-optional-chaining/article.md
+++ b/1-js/04-object-basics/07-optional-chaining/article.md
@@ -7,7 +7,7 @@ The optional chaining `?.` is a safe way to access nested object properties, eve
 
 ## The "non-existing property" problem
 
-If you've just started to read the tutorial and learn JavaScript, maybe the problem hasn't touched you yet, but it's quite common.
+If you've just started to learn JavaScript, maybe you haven't encountered this problem yet, but it's quite common.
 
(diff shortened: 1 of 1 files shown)


joaquinelio (commented):
NON NATIVE speaking
you can safely ignore me  :) Im here only to learn.

I understand many changes due grammar faults (most annoying the lack of pronouns/articles/to + infinitive )
or due to things that sound somewhat odd or uncommon 

Is anything wrong in line 10? or just style?

1-js/04-object-basics/07-optional-chaining/article.md
```diff
@@ -32,7 +32,7 @@ In many practical cases we'd prefer to get `undefined` instead of an error here
 let html = document.querySelector('.elem').innerHTML; // error if it's null
 ```
 
-Once again, if the element doesn't exist, we'll get an error accessing `.innerHTML` property of `null`. And in some cases, when the absence of the element is normal, we'd like to avoid the error and just accept `html = null` as the result.
+Once again, if the element doesn't exist, we'll get an error accessing `.innerHTML` property of `null`. And in some cases, where the absence of the element is normal, we'd like to avoid the error and just accept `html = null` as the result.
```
shallow-beach, line 35: disagree, Line 16 uses 'when'

PR #3271, opened 2022-11-19
Correct a small mistake

shc0743:
Correct a small mistake: Because javascript.info is https, the actual reason for the failure of fetching the http address is "Mixed Content". As the console says, "Mixed Content: The page at 'https://zh.javascript.info/fetch-crossorigin' was loaded over HTTPS, but requested an insecure resource 'http://example.com/'. This request has been blocked; the content must be served over HTTPS."
Change http to https to reflect the impact of cross-origin requests.


joaquinelio:
👍 👍 👍 
block by CORS
good catch( ) !!!

PR #3273, opened 2022-11-21
Added a section about pass-by-value and pass-by-reference in the func…

jeffkim1118:
Added a section in the function basic folder readme file to cover the concepts of both pass-by-value and pass-by-reference in JavaScript function.


joaquinelio:
the work is good
but
topic is already covered.

jeffkim1118:
Oh, I didn't know.

jeffkim1118:
Cool.

PR #3276, opened 2022-11-23
Fixed grammar

bogdanbacosca:
There should be no , before the word and


bogdanbacosca:
> mandatory comma when connecting 2 complete thoughts

Agreed. My impression was that this line was 1 single idea

bogdanbacosca:
Closed by accident.
Had to change the source branch of this PR.
Refer to #3305

PR #3279, opened 2022-11-26
minor code formatting for 1-js/02-first-steps/10-ifelse/article.md

bogdanbacosca:
The dots look weird in commit 657e389.
Also the extra ' is not needed from commit 45a2d14.


iliakan:
What's weird in … compared to ... ?

bogdanbacosca:
@iliakan might look different on my markdown previewer

iliakan:
Please look at the website instead ;)

bogdanbacosca:
@iliakan I double checked the website. It looks just a tiny bit different at the beginning of the Boolean conversion section. I don't think it's a big problem maybe just for consistency's sake. Same goes for the other commit.

iliakan:
Please leave dots as is, they're not a problem at all.

bogdanbacosca:
Done. Reverted that commit.

PR #3281, opened 2022-11-27
fix: condition in whilt-for chapter

leviding:
["breaking" when no number is entered] is described in line 224, but it will breaking too when enter number 0. So add "value !== 0" in condation.

Related issue https://github.com/javascript-tutorial/zh.javascript.info/issues/1109

### DIFF
--- a/1-js/02-first-steps/13-while-for/article.md
+++ b/1-js/02-first-steps/13-while-for/article.md
@@ -231,7 +231,7 @@ while (true) {
   let value = +prompt("Enter a number", '');
 
 *!*
-  if (!value) break; // (*)
+  if (value !== 0 && !value) break; // (*)
 */!*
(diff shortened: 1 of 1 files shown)


Blackmegg1:
> ["breaking" when no number is entered] is described in line 224, but it will breaking too when enter number 0. So add "value !== 0" in condation.
> 
> Related issue [javascript-tutorial/zh.javascript.info#1109](https://github.com/javascript-tutorial/zh.javascript.info/issues/1109)

It is also incorrect to add only this condition.
If the user cancels the input, then return "nul". 
`+null !== 0 // false`
It leads to another error.

joaquinelio:
Maybe it's simple on purpose.

Other examples have more complex prompts, when the topic need it

PR #3284, opened 2022-11-28
Async 01-callbacks

bogdanbacosca:
update to line 51:
> Added "inside it" to make the text a little more readable, doesn't change the context.

update to line 65:
> Replaced "loads" with "finished loading" to improve the description of the `load event`. Could also be replaced with "has completely loaded".

### DIFF
--- a/1-js/11-async/01-callbacks/article.md
+++ b/1-js/11-async/01-callbacks/article.md
@@ -62,7 +62,7 @@ newFunction(); // no such function!
 
 Naturally, the browser probably didn't have time to load the script. As of now, the `loadScript` function doesn't provide a way to track the load completion. The script loads and eventually runs, that's all. But we'd like to know when it happens, to use new functions and variables from that script.
 
-Let's add a `callback` function as a second argument to `loadScript` that should execute when the script loads:
+Let's add a `callback` function as a second argument to `loadScript` that should execute when the script finished loading:
 
 ```js
 function loadScript(src, *!*callback*/!*) {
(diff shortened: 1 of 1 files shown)


bogdanbacosca:
Suggestion for line 63:
> Provided additional reasoning as to why the script didn't have time to load, even if line 94 shows a lodash script coming from a CDN.

Maybe we could also provide additional reasoning for line 53?

1-js/11-async/01-callbacks/article.md
```diff
@@ -60,9 +60,9 @@ newFunction(); // no such function!
 */!*
 ```
 
-Naturally, the browser probably didn't have time to load the script. As of now, the `loadScript` function doesn't provide a way to track the load completion. The script loads and eventually runs, that's all. But we'd like to know when it happens, to use new functions and variables from that script.
+Naturally, the browser probably didn't have time to load the script because of some external http request. As of now, the `loadScript` function doesn't provide a way to track the load completion. The script loads and eventually runs, that's all. But we'd like to know when it happens, to use new functions and variables from that script.
```
shallow-beach, line 63: i don't think this is correct (there is no external http request involved, and has more to do with loadScript being asynchronous)

1-js/11-async/01-callbacks/article.md
```diff
@@ -60,9 +60,9 @@ newFunction(); // no such function!
...
 
-Let's add a `callback` function as a second argument to `loadScript` that should execute when the script loads:
+Let's add a `callback` function as a second argument to `loadScript` that should execute when the script finished loading:
```
shallow-beach, line 65: suggest:

>Let's add a `callback` function as a second argument to `loadScript` that should execute when the script finishes loading:

PR #3301, opened 2022-12-12
Update article.md

mishrasur7:
Providing better word internet as I think it makes more sense.


joaquinelio:
may be an intranet.
may be a local environment too

just my opinion, Im nobody here

PR #3302, opened 2022-12-12
grammar & legibility (to check)

joaquinelio:
non-native warning
The problem: I fixed it the spanish way, maybe wrong.

> The situation when the browser sent a request and has a pending connection with the server, is standard for this method.

I think the comma before "is" should be removed. 
Workaround: 2 commas.

### DIFF
--- a/5-network/10-long-polling/article.md
+++ b/5-network/10-long-polling/article.md
@@ -29,7 +29,7 @@ The flow:
 3. When a message appears - the server responds to the request with it.
 4. The browser makes a new request immediately.
 
-The situation when the browser sent a request and has a pending connection with the server, is standard for this method. Only when a message is delivered, the connection is reestablished.
+This situation, where the browser has sent a request and keeps a pending connection with the server, is standard for this method. Only when a message is delivered, the connection is closed and reestablished.
 
 ![](long-polling.svg)
 


p2635:
> The situation when the browser sent a request and has a pending connection with the server, is standard for this method. Only when a message is delivered, the connection is reestablished.

This is a horrible sentence in general. As a newbie, I have to read this 5 times, I don't even know what the main point is here. Is it just 'This is a standard method in most cases'? The rest of the sentence is repeating what the article already says:

> 1. A request is sent to the server.
> 2. The server doesn’t close the connection until it has a message to send.
> 3. When a message appears – the server responds to the request with it.
> 4. The browser makes a new request immediately.

p2635:
Also, I am confused by this:

> Only when a message is delivered, the connection is reestablished.

This is conflicting with this:

> 2. The server doesn’t close the connection until it has a message to send.

How can connection be reestablished? It was never closed. Maybe the author was meant to say 'Only when a message is delivered, a new request is made'.

If I get time, I am trying to review all the content for these problems #3335.

p2635:
A small recommendation, change the title of this PR to something more descriptive. For example: 'Check grammar/legibility for 'long polling' article' instead of 'grammar & legibility (to check)'.

joaquinelio:
Well, @p2635 , I tried to fix both issues here.  
Any review/advice from a native Eng speaker would be nice...

p2635:
I am a native English speaker, but I still need the original author to answer my questions above. I'm not a maintainer of the project , so I don't get to make the final decisions.

PR #3309, opened 2022-12-20
we are suggesting some changes, after translating into Romanian

AurelianVirgil:
Hi,

We would like to propose a few grammatical and lexical suggestions.

Thanks,

### DIFF
--- a/1-js/02-first-steps/03-strict-mode/article.md
+++ b/1-js/02-first-steps/03-strict-mode/article.md
@@ -1,6 +1,6 @@
 # The modern mode, "use strict"
 
-For a long time, JavaScript evolved without compatibility issues. New features were added to the language while old functionality didn't change.
+For a long time, JavaScript evolved without compatibility issues. New features were added to the language, while old functionality didn't change.
 
 That had the benefit of never breaking existing code. But the downside was that any mistake or an imperfect decision made by JavaScript's creators got stuck in the language forever.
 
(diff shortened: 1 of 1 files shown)


1-js/02-first-steps/03-strict-mode/article.md
```diff
@@ -39,7 +39,7 @@ Only comments may appear above `"use strict"`.
 ````
 
 ```warn header="There's no way to cancel `use strict`"
-There is no directive like `"no use strict"` that reverts the engine to old behavior.
+There is no directive like `"no use strict"` that reverts the JavaScript engine to old behavior.
```
joaquinelio, line 42: to  the old
maybe?

AurelianVirgil: Hi,

Yes, you are right. I overlooked it.

AurelianVirgil: Screenshot 2022-12-21 113411

AurelianVirgil: I have attached the discussion, it is in Romanian, but you can read the proposed version in English.
It seems I was very tired when I edited it, sorry.

The proposed version was :  ”There is no directive like "no use strict" that reverts the javascript engine used by browsers to the old behavior.” 

I removed ”used by browsers” but I forgot to put ”the”.

What do you think?

PR #3311, opened 2022-12-25
Update commens in few code snippets

Suyash-Purwar:
In the Non-writable and Non-configurable section, the comments in a few of the code snippets says 'Error' but that is only the case when the 'strict' mode is enabled. I understand that it is also mentioned but it should be reflected in the code snippet as well so that the reader is absolutely clear about the concept.


JeraldVin:
Probably applicable in a lot of places within the tutorial, but 'strict' mode is assumed everywhere as given here https://javascript.info/strict-mode

`All examples in this tutorial assume strict mode unless (very rarely) specified otherwise.`

PR #3312, opened 2022-12-26
👾 add code formatting

Rnbsov:


1-js/09-classes/05-extend-natives/article.md
```diff
@@ -32,7 +32,7 @@ When `arr.filter()` is called, it internally creates the new array of results us
 
 Even more, we can customize that behavior.
 
-We can add a special static getter `Symbol.species` to the class. If it exists, it should return the constructor that JavaScript will use internally to create new entities in `map`, `filter` and so on.
+We can add a special static getter [`Symbol.species`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/species) to the class. If it exists, it should return the constructor that JavaScript will use internally to create new entities in `map`, `filter` and so on.
```
bogdanbacosca, line 35:
```suggestion
We can add a special static getter [Symbol.species](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/species) to the class. If it exists, it should return the constructor that JavaScript will use internally to create new entities in `map`, `filter` and so on.
```

I don't think you really need the code format for the link description.

PR #3317, opened 2023-01-03
CSS-animation

bogdanbacosca:
A few tweaks and improvements:
line 3: more verbose language
line 50: `shorthand` is a better description, also added `in the same line` for clarity
line 89: grammar fix
line 91: grammar fix
line 144: added `whereas` for clarity
line 161: minor tweak
line 165: replaced `that` with `this` because it could make more sense
lines 171,195,233: minor tweaks
line 202: grammar fix
line 225: replaced `huge ones` with `huge values` to be more specific
line 251: minor tweak
line 253: formatting: added spaces
line 304: improvement suggestion: switch action at the beginning
line 324: used the same language as in line 312 for consistency.
line 347: considered that the two sentences belong together, making the idea be clearer
line 441: improved flow
line 444: formatting: removed space
task 2: formatting: added spaces
task 3: description suggestion

### DIFF
--- a/7-animation/2-css-animations/article.md
+++ b/7-animation/2-css-animations/article.md
@@ -248,7 +248,7 @@ Why it happens is pretty obvious if we look at the graph of the given Bezier cur
 
 ![](bezier-train-over.svg)
 
-We moved the `y` coordinate of the 2nd point below zero, and for the 3rd point we made it over `1`, so the curve goes out of the "regular" quadrant. The `y` is out of the "standard" range `0..1`.
+We moved the `y` coordinate of the 2nd point below zero and the 3rd point over `1`, so the curve goes out of the "regular" quadrant. The `y` is out of the "standard" range `0..1`.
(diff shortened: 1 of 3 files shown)


7-animation/2-css-animations/article.md
```diff
@@ -344,7 +344,7 @@ These values are rarely used, as they represent not a real animation, but rather
 
 When the CSS animation finishes, the `transitionend` event triggers.
 
-It is widely used to do an action after the animation is done. Also we can join animations.
+It is widely used to do an action after the animation is done and to join multiple animations.
```
joaquinelio, line 347: Needs to be fixed, but I prefer the original structure

```suggestion
It is widely used to do an action after the animation is done.  We can also join multiple animations.
```

joaquinelio: or 
"Also, we can..."

PR #3325, opened 2023-01-18
{} wrapped in ()

sagarpanchal:


Alexandre887 (commented):
`()` are not required here. 

If you pass `{}.toString.call(user)` as a function parameter (for example, 
`alert( {}.toString.call(user) )` ), everything will work fine.

sagarpanchal:
image

PR #3328, opened 2023-01-21
Confusing info

Alexandre887:
> Or, in other words, **multiplies `x` by itself `n` times** and returns the result.

E.g. 
x = 2, n = 3

In that case, it appears that **2** must be multiplied by itself **3** times (`2 * 2 * 2 * 2`). The result would be `16`, although 2<sup>3</sup> would be `8`.
Such an expression in the task condition may confuse the readers, so I suggest it be removed.

### DIFF
--- a/1-js/02-first-steps/15-function-basics/4-pow/task.md
+++ b/1-js/02-first-steps/15-function-basics/4-pow/task.md
@@ -4,7 +4,7 @@ importance: 4
 
 # Function pow(x,n)
 
-Write a function `pow(x,n)` that returns `x` in power `n`. Or, in other words, multiplies `x` by itself `n` times and returns the result.
+Write a function `pow(x,n)` that returns `x` in power `n`.
 
 ```js
 pow(3, 2) = 3 * 3 = 9


1-js/02-first-steps/15-function-basics/4-pow/task.md
```diff
@@ -4,7 +4,7 @@ importance: 4
 
 # Function pow(x,n)
 
-Write a function `pow(x,n)` that returns `x` in power `n`. Or, in other words, multiplies `x` by itself `n` times and returns the result.
+Write a function `pow(x,n)` that returns `x` in power `n`.
```
bogdanbacosca, line 7: My suggestion:
```suggestion
Write a function `pow(x,n)` that returns `x` to the power of `n`.
```

How about a wikipedia article for more details instead of `Or, in other words, multiplies `x` by itself `n` times and returns the result.` ?

Alexandre887: > How about a wikipedia article for more details instead of `Or, in other words, multiplies x by itselfn times and returns the result.` ?

I don't think a separate clarification, much less a wikipedia article, is needed in this case. Exponentiation is a simple mathematical operation known to anyone from comprehensive school. We don't resort to wikipedia articles when addition/subtraction operations are mentioned, right? 
In my opinion, it would be better to leave just a description of what needs to be done here:
> Write a function `pow(x,n)` that returns `x` to the power of `n`.

bogdanbacosca (approved):

p2635:
A small recommendation, change the title of this PR to something more descriptive. For example: 'Clarify exponent section for the functions article' instead of 'Confusing info'.

PR #3329, opened 2023-01-21
The solution to the task does not match its condition

Alexandre887:
__[Function pow(x,n)](https://javascript.info/function-basics#function-pow-x-n)__.

The task description says:
> Write a function pow(x,n) that returns x in power n. Or, in other words, multiplies x by itself n times and returns the result.
> . . .
> P.S. In this task the function should support **only natural values of n**: integers up from 1.

The current check for a natural number in the task solution does not take into account the fact that the value passed in can be a fractional value greater than 1. That is, if we pass in, for example, 1.1 as `n`, the function will execute without any problems, even though 1.1 is not a natural number.

This PR proposes an improved check that considers this nuance.

### DIFF
--- a/1-js/02-first-steps/15-function-basics/4-pow/solution.md
+++ b/1-js/02-first-steps/15-function-basics/4-pow/solution.md
@@ -13,9 +13,9 @@ function pow(x, n) {
 let x = prompt("x?", '');
 let n = prompt("n?", '');
 
-if (n < 1) {
-  alert(`Power ${n} is not supported, use a positive integer`);
-} else {
+if (n >= 1 && n % 1 == 0) {
   alert( pow(x, n) );
+} else {
+	alert(`Power ${n} is not supported, use a positive integer`);
 }
 ```


p2635:
Unless the point of the exercise is to exclude fractional values right? For someone like me, I stuck to whole numbers when I did the exercise. I think it's worth clarifying the requirements of this specific task.

Alexandre887:
> Unless the point of the exercise is to exclude fractional values right? For someone like me, I stuck to whole numbers when I did the exercise. I think it's worth clarifying the requirements of this specific task.

The description of the task must be consistent with the proposed solution. The author's solution does not take into account one nuance, which I described at the beginning of this PR.  With the current code, the function can accept nonnatural numbers, even though the task says that it should only accept natural numbers.

There are two solutions to this problem:

1. Accept my PR, which excludes the possibility of the function working with nonnatural numbers.
2. Paraphrasing the task description, creating a different condition.

p2635:
Agreed, I'm leaving that to the creator to decide on.

PR #3346, opened 2023-01-31
 Date and time

bogdanbacosca:
Line 34 - There is no variable named date prior to this mention. We could keep `.getTime( )` as is, or use the variable `now` or `Jan01_1970`.
line 57 - Though `date` is used in some docs, `day` makes more sense and I also saw day in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date). P.S this is just modified in one place: if agreed, will have to change this in multiple places.
line 61 - grammar fix
line 62 - replaced `is actually the day of` with `represents the day of`. I swapped date with day ahead of time, we can ignore that if line 57 does not change.
line 69 - typo
line 108 - replaced `shifted` with `offset`
line 167 - grammar fix
line 244 - took out `in JavaScript` because it is implicit and may improve readability.
line 377 - suggestion
line 390 - modified by the same logic(format) of line 388
line 402 - extra space
line 415 - took out `in JavaScript` because it is implicit and may improve readability.
line 416 - suggestion
line 424 - `returns` may be more accurate
Tasks 2 and 3 - formatting improvements
Tasks 7 and 8 - some improvements

### DIFF
--- a/1-js/05-data-types/11-date/7-get-seconds-to-tomorrow/solution.md
+++ b/1-js/05-data-types/11-date/7-get-seconds-to-tomorrow/solution.md
@@ -1,4 +1,4 @@
-To get the number of milliseconds till tomorrow, we can substract the current date from "tomorrow 00:00:00".
+To get the number of milliseconds till tomorrow, we can subtract the current date from "tomorrow 00:00:00".
 
 First, we generate that "tomorrow", and then do it:
 
--- a/1-js/05-data-types/11-date/7-get-seconds-to-tomorrow/solution.md
+++ b/1-js/05-data-types/11-date/7-get-seconds-to-tomorrow/solution.md
@@ -1,4 +1,4 @@
-To get the number of milliseconds till tomorrow, we can from "tomorrow 00:00:00" substract the current date.
+To get the number of milliseconds till tomorrow, we can substract the current date from "tomorrow 00:00:00".
 
 First, we generate that "tomorrow", and then do it:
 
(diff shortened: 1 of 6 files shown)


1-js/05-data-types/11-date/7-get-seconds-to-tomorrow/solution.md
```diff
@@ -1,4 +1,4 @@
-To get the number of milliseconds till tomorrow, we can from "tomorrow 00:00:00" substract the current date.
+To get the number of milliseconds till tomorrow, we can substract the current date from "tomorrow 00:00:00".
```
JeraldVin, line 1:
```suggestion
To get the number of milliseconds till tomorrow, we can subtract the current date from "tomorrow 00:00:00".
```
typo 😄

1-js/05-data-types/11-date/article.md
```diff
@@ -54,19 +54,19 @@ To create a new `Date` object call `new Date()` with one of the following argume
     // Wed Jan 25 2017 16:00:00 GMT-0800 (Pacific Standard Time)
     ```
 
-`new Date(year, month, date, hours, minutes, seconds, ms)`
+`new Date(year, month, day, hours, minutes, seconds, ms)`
```
JeraldVin, line 57: would date be more clear? Since new Date().getDay() returns the number of the weekday? 🤔

joaquinelio: agreed.
Using date aligns with getDate() and avoids confusion with getDay().

oh... 2023?

PR #3352, opened 2023-02-03
Return promise result after initialization.

kareemgamal1:
Or have 'use strict' at the beginning of the script.


shallow-beach:
disagree, inconsistent with the rest of the [section's examples](https://javascript.info/async-await#error-handling)

PR #3359, opened 2023-02-07
Fix Typos

detinsley1s:
I changed "parenthesis" to "parentheses" since "parenthesis" is singular and "parentheses" is plural and the sentence uses the plural version of the word. I changed "layed" to "laid" since "layed" is not the correct way to spell the past tense of "lay." Finally, I changed "setup" to "set up" since "setup" is a noun and "set up" is a verb and the verb should be used here.


1-js/05-data-types/03-string/article.md
```diff
@@ -254,7 +254,7 @@ while (true) {
 }
 ```
 
-The same algorithm can be layed out shorter:
+The same algorithm can be laid out shorter:
```
JeraldVin, line 257:
```suggestion
The same algorithm can be made shorter:
```
would 'made' be more appropriate/simple?

shallow-beach (approved):

PR #3363, opened 2023-02-09
Improve docs

rahulpandey70:
Bit of improvement in docs which i believe, that will help to developer to understand clearly.

### DIFF
--- a/1-js/02-first-steps/09-comparison/article.md
+++ b/1-js/02-first-steps/09-comparison/article.md
@@ -126,18 +126,18 @@ The same thing happens with an empty string:
 alert( '' == false ); // true
 ```
 
-This happens because operands of different types are converted to numbers by the equality operator `==`. An empty string, just like `false`, becomes a zero.
+This happens because operands of different data types are converted to numbers by the equality operator `==`. An empty string, just like `false`, becomes a zero.
 
 What to do if we'd like to differentiate `0` from `false`?
 
 **A strict equality operator `===` checks the equality without type conversion.**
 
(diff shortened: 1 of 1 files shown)


1-js/02-first-steps/09-comparison/article.md
```diff
@@ -126,18 +126,18 @@ The same thing happens with an empty string:
 alert( '' == false ); // true
 ```
 
-This happens because operands of different types are converted to numbers by the equality operator `==`. An empty string, just like `false`, becomes a zero.
+This happens because operands of different data types are converted to numbers by the equality operator `==`. An empty string, just like `false`, becomes a zero.
```
JeraldVin, line 129: Most of the articles in the tutorial do use 'types' instead of 'data types'. This is true even for MDN docs and others(eg:- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures) where mostly 'types' is used. So not sure if this is necessary - if it is, then probably needs changes in other articles as well

rahulpandey70: I think this is necessary, because when developers (beginners, non-tech) read this article, it makes hard to understand.
This is my personal view if you think this isn't necessary then it's ok.

JeraldVin: Sure, I just said that I wasn't sure :smile:

rahulpandey70: Haha, Let's wait what maintainer's think.

shallow-beach (approved):
approve but only for this early article, disagree with revising later articles for consistency

PR #3326, opened 2023-01-18
Typo/grammar corrections in import-export article

sagarpanchal:

PR #3213, opened 2022-10-03
typo "optimzed

joaquinelio:

PR #3365, opened 2023-02-09
Update article.md

konstantin-agafonov:


1-js/13-modules/02-import-export/article.md
```diff
@@ -99,7 +99,7 @@ Well, there are few reasons.
 ```smart header="Don't be afraid to import too much"
 Modern build tools, such as [webpack](https://webpack.js.org/) and others, bundle modules together and optimize them to speedup loading. They also removed unused imports.
 
-For instance, if you `import * as library` from a huge code library, and then use only few methods, then unused ones [will not be included](https://github.com/webpack/webpack/tree/main/examples/harmony-unused#examplejs) into the optimzed bundle.
+For instance, if you `import * as library` from a huge code library, and then use only few methods, then unused ones [will not be included](https://github.com/webpack/webpack/tree/main/examples/harmony-unused#examplejs) into the optimized bundle.
```
JeraldVin, line 102: https://github.com/javascript-tutorial/en.javascript.info/pull/3326

joaquinelio: yeah, twice
#3213

PR #3366, opened 2023-02-09
Update article.md

konstantin-agafonov:


1-js/13-modules/02-import-export/article.md
```diff
@@ -294,7 +294,7 @@ import func from '/path/to/func.js';
 ...
 ```
 
-Still, some teams consider it a serious drawback of default exports. So they prefer to always use named exports. Even if only a single thing is exported, it's still exported under a name, without `default`.
+Still, some teams consider default exports to be a serious drawback. So they prefer to always use named exports. Even if only a single thing is exported, it's still exported under a name, without `default`.
```
JeraldVin, line 297: without the word 'it', the context of the 'serious drawback' is removed

PR #3369, opened 2023-02-12
Improve the text

detinsley1s:
I made numerous additions and edits to make the page read more fluently. I also fixed a misspelling of "parentheses."

### DIFF
--- a/1-js/06-advanced-functions/04-var/article.md
+++ b/1-js/06-advanced-functions/04-var/article.md
@@ -4,7 +4,7 @@
 ```smart header="This article is for understanding old scripts"
 The information in this article is useful for understanding old scripts.
 
-That's not how we write new code.
+This is not how we write new code.
 ```
 
 In the very first chapter about [variables](info:variables), we mentioned three ways of variable declaration:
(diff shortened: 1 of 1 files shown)


manohargeekologix (approved):

1-js/06-advanced-functions/04-var/article.md
```diff
@@ -160,9 +160,9 @@ function sayHi() {
 sayHi();
 ```
 
-People also call such behavior "hoisting" (raising), because all `var` are "hoisted" (raised) to the top of the function.
+People also call such behavior "hoisting" (raising), because all `var`s are "hoisted" (raised) to the top of the function.
```
shallow-beach, line 163: Maybe:

>People also call such behavior "hoisting" (raising), because all `var` declarations are "hoisted" (raised) to the top of the function.

1-js/06-advanced-functions/04-var/article.md
```diff
@@ -205,7 +205,7 @@ sayHi();
 
 Because all `var` declarations are processed at the function start, we can reference them at any place. But variables are undefined until the assignments.
 
-In both examples above, `alert` runs without an error, because the variable `phrase` exists. But its value is not yet assigned, so it shows `undefined`.
+In both examples above, `alert` runs without an error, because the variable `phrase` exists, but its value is not yet assigned, so it shows `undefined`.
```
shallow-beach, line 208: I think punctuation should be removed:

>In both examples above, `alert` runs without an error because the variable `phrase` exists but its value is not yet assigned, so it shows `undefined`.

shallow-beach (changes requested):

shallow-beach (approved):

PR #3370, opened 2023-02-14
feat: replace old term "Lexical Environment" with "Environment Record"

andrew1931:
Here are links to [ECMAScript2020](https://262.ecma-international.org/11.0/#sec-lexical-environments) and [ECMAScript2023](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-environment-records).

There is no such term as "Lexical Environment" anymore (only "LexicalEnvironment" component as a state of an [execution context](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-execution-contexts)).

All "Lexical Environment" terms were replaced with "Environment Record" all over the "Executable Code and Execution Contexts" article.


iliakan:
I'm glad they finally did it in the standard.

It'll be much easier to describe. Need to rewrite the article and pics =)

shallow-beach:
[this article](https://javascript.info/closure#lexical-environment) is definitely a prime target

PR #3373, opened 2023-02-18
Update article.md

ImVietnam:


bogdanbacosca:
`key` is there for formatting. The word `key` never actually shows on the website. From inspecting the HTML, it looks they get a `<kbd>` tag which I think is good. It is a feature of the javascript.info website builder.

PR #3406, opened 2023-03-16
Added small detail for inderstanding

mukulbindal:
In unstrict mode, no error is thrown


1-js/09-classes/04-private-protected-properties-methods/article.md
```diff
@@ -148,7 +148,7 @@ let coffeeMachine = new CoffeeMachine(100);
 
 alert(`Power is: ${coffeeMachine.power}W`); // Power is: 100W
 
-coffeeMachine.power = 25; // Error (no setter)
+coffeeMachine.power = 25; // Error (no setter) in strict mode
```
shallow-beach, line 151: "All examples in this tutorial assume strict mode unless (very rarely) specified otherwise."
https://javascript.info/strict-mode#should-we-use-strict

PR #3421, opened 2023-03-31
Fix equality check

Ryuukishi:


shallow-beach:
appears to already have been solved with `==`, vote to close

PR #3430, opened 2023-04-12
Add comma for better readability

KrappRamiro:
I think a comma (,) after every "In the code below" makes the text more readable. For example, there is a line that says:

In the code below `prompt` asks for `width`, but not for `title`:

It can be mistakenly interpreted as "look for the code below `prompt()` " , instead of "look how `prompt` works in the code below" Have a good day :)


joaquinelio:
👍  I think the comma Is mandatory there,
but Im not a native speaker.

joachimklug (approved):

shallow-beach (approved):

PR #3436, opened 2023-04-24
Use of 3 dots spread operator can be confusing for newcomers

mysteryjeans:


joachimklug (changes requested):
Hi @mysteryjeans 
this is as described in the ECMA standard and should not be changed:
https://262.ecma-international.org/13.0/#sec-object.assign

Having only two dots is not a valid instruction

PR #3475, opened 2023-06-10
Update article.mdUpdate: Added push, pop, shift, unshift to list of methods that modify original Array

rijans:
Added push, pop, shift, unshift to list of methods in summary, that modify original Array

### DIFF
--- a/1-js/05-data-types/05-array-methods/article.md
+++ b/1-js/05-data-types/05-array-methods/article.md
@@ -767,7 +767,7 @@ A cheat sheet of array methods:
 - Additionally:
   - `Array.isArray(value)` checks `value` for being an array, if so returns `true`, otherwise `false`.
 
-Please note that methods `sort`, `reverse` and `splice` modify the array itself.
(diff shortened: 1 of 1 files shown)


Alexandre887 (approved):

1-js/05-data-types/05-array-methods/article.md
```diff
@@ -767,7 +767,7 @@ A cheat sheet of array methods:
 - Additionally:
   - `Array.isArray(value)` checks `value` for being an array, if so returns `true`, otherwise `false`.
 
-Please note that methods `sort`, `reverse` and `splice` modify the array itself.
+Please note that methods `push`, `pop`, `shift`, `unshift`, `sort`, `reverse` and `splice` modify the array itself.
```
shallow-beach, line 770: don't think these are needed

joaquinelio: I agree. This addition would be distracting. The text is a warning for methods that modify the original array as a maybe unwanted side effect.

PR #3477, opened 2023-06-12
Update article.md

anujbarochia:
The Hot Key for the multiple line comment is not true in the javascript.info document. The new hot key is `Shift + Alt + A` ; instead of `Ctrl + Shift + /`


Alexandre887 (commented):
For me, only <kbd>Ctrl + /</kbd> and <kbd>Ctrl + Shift + /</kbd> works.

Google Chrome `115.0.5790.170`

<hr>

<ins>https://developer.chrome.com/docs/devtools/shortcuts/</ins>:

image

PR #3478, opened 2023-06-13
change mdn links

ImVietnam:


Alexandre887 (commented):
`mdn:js` is a special predefined value that is used to write the links more briefly. This value is valid all over javascript.info, your PR is irrelevant

PR #3513, opened 2023-07-27
Correction to "Format the relative date" task

CottonCandyZ:
Change 'second' to 'millisecond'.


Alexandre887 (commented):
The condition of the task is correct, your edit is irrelevant. 

```javascript
alert( formatDate(new Date(new Date - 1)) ); // "right now" 
// 1 millisecond ago, which is less than 1 second.

PR #3537, opened 2023-08-13
remove "one-liner" statement

Sammeeey:
- removed the statement "As you can see, that's almost one-liner", because it doesn't make sense - neither grammatically nor logically. These 20 lines of code aren't even close to a one-liner.


shallow-beach (approved):
fair enough. even just the `<script>` is 13 lines.

PR #3539, opened 2023-08-13
A typo in the article

Alexandre887:
# [Numbers](https://javascript.info/number)

There is six zeroes to the left from $`\large1`$ in $`\large0.000001`$, not five.

### DIFF
--- a/1-js/05-data-types/02-number/article.md
+++ b/1-js/05-data-types/02-number/article.md
@@ -50,7 +50,7 @@ let mсs = 0.000001;
 Just like before, using `"e"` can help. If we'd like to avoid writing the zeroes explicitly, we could write the same as:
 
 ```js
-let mcs = 1e-6; // five zeroes to the left from 1
+let mcs = 1e-6; // six zeroes to the left from 1
 ```
 
 If we count the zeroes in `0.000001`, there are 6 of them. So naturally it's `1e-6`.


1-js/05-data-types/02-number/article.md
```diff
@@ -50,7 +50,7 @@ let mсs = 0.000001;
 Just like before, using `"e"` can help. If we'd like to avoid writing the zeroes explicitly, we could write the same as:
 
 ```js
-let mcs = 1e-6; // five zeroes to the left from 1
+let mcs = 1e-6; // six zeroes to the left from 1
```
shallow-beach, line 53: not helpful. perhaps:

`let mcs = 1e-6; // five zeroes to the left of 1 (to the right of the decimal point)`

Alexandre887: There is no such “up to the decimal point” condition here. Besides, literally after this sentence comes the following: 

> If we count the zeroes in `0.000001`, there are **6** of them. So naturally it's `1e-6`.

shallow-beach: then fix that:

> > If we count the zeroes in `0.000001`, there are **5** of them after the decimal point. So naturally it's `1e-6`.

shallow-beach:
same confusion as #3541

joaquinelio:
I never liked the "zeroes count" explanation.
It doesn't help to build intuition.

We have decimal notation. It comes with a key property:
Multiplying by 10, 100, 1000, moves the decimal point 1, 2 , 3 places to the right.
Dividing, to the left.

1.23e-9 moves the decimal point 9 places to the left. "Naturally" for real, no need for extra steps

I couldn't figure out how to change the text without ruining it.

shallow-beach:
> I never liked the "zeroes count" explanation. It doesn't help to build intuition.
> 
> We have decimal notation. It comes with a key property: Multiplying by 10, 100, 1000, moves the decimal point 1, 2 , 3 places to the right. Dividing, to the left.
> 
> 1.23e-9 moves the decimal point 9 places to the left. "Naturally" for real, no need for extra steps
> 
> I couldn't figure out how to change the text without ruining it.

zeros count only makes sense/builds intuition when applied to $\frac{1}{10^x}$. trying to apply it directly to a result is confusing - e.g, if we have `5555e-4`, that results in `0.5555`, which does *not* have 'usefully' countable zeros.

joaquinelio:
> > I never liked the "zeroes count" explanation. It doesn't help to build intuition.
> > We have decimal notation. It comes with a key property: Multiplying by 10, 100, 1000, moves the decimal point 1, 2 , 3 places to the right. Dividing, to the left.
> > 1.23e-9 moves the decimal point 9 places to the left. "Naturally" for real, no need for extra steps
> > I couldn't figure out how to change the text without ruining it.
> 
> zeros count makes sense in that e X ≡ 10 X (`X` is zero count of multiplier (either of numerator's zeros ( X ≥ 0 , with denominator=1) or of denominator's zeros ( X < 0 , with numerator=1)), which coincides with count of decimal point moves). trying to apply it directly to a result is confusing - e.g, if we have `5555e-4`, that results in 5555 ∗ 10 − 4 = `0.5555`, which does _not_ have 'usefully' countable zeros.

It's easier to think that way, don't you think?applying directly 
e-4,
5555. --> 0.5555
4 spaces to the left
Where each space is /power of 10

PR #3540, opened 2023-08-13
Fix redundancy

salpelter:
"Lowest" already conveys the idea of being at the bottom level, adding "very" before it creates redundancy.


Alexandre887 (approved):

smith558:
These are some good suggestions, we will merge them if you can open a new pull request now. 🙂

PR #3543, opened 2023-08-14
Update task.md to specify strict mode.

phalgunv:
The solution provided to this question assumes function is executed in "strict" mode. The code has been updated to mention

### DIFF
--- a/1-js/04-object-basics/04-object-methods/4-object-property-this/task.md
+++ b/1-js/04-object-basics/04-object-methods/4-object-property-this/task.md
@@ -9,6 +9,7 @@ Here the function `makeUser` returns an object.
 What is the result of accessing its `ref`? Why?
 
 ```js
+"use strict";
 function makeUser() {
   return {
     name: "John",


Alexandre887 (commented):
From [javascript.info/strict-mode](https://javascript.info/strict-mode):

image

Your PR is irrelevant.

phalgunv:
For the code snippet in question, the results differ in strict mode. The JS engine would not read that line you highlighted, leaving new developers baffled.

The code snippets should be self contained without having to back to documentation to figure things out.

Alexandre887:
> The code snippets should be self contained without having to back to documentation to figure things out.

If you think so, submit a Pull Request that adds `"use strict"` to all the code samples in the tutorial. By the way, there are **171 articles** in the tutorial, and each article contains **10-15** such examples. Do some simple mathematical calculations in your mind and imagine how many examples you will have to fix.

The article about strict mode says that absolutely all of the tutorial examples involve using "use strict". In case the behavior with and without ``use strict'` is different, it should be mentioned in the article.

The most you can do is add a small clarification in the text before this example, like "(note, we are using strict mode in this task)".

phalgunv:
> If you think so, submit a Pull Request that adds "use strict" to all the code samples in the tutorial.

 I'm not sure why it you recommend adding it to all examples. There would be snippets of code which may not be affected by use of strict mode. For example, see https://github.com/javascript-tutorial/en.javascript.info/blob/285083fc71ee3a7cf55fd8acac9c91ac6f62105c/1-js/02-first-steps/04-variables/article.md?plain=1#L33
This code adding "use strict" here is redundant as it does not affect the output.
Also, there may be some other examples where strict mode is enforced by javascript as in case of classes/modules; the article you mentions rightly mentions so. 

It makes sense to add "use strict" only to the code which would have a different result in non strict mode. That makes the code samples robust when copied and executed elsewhere (for e.g. console tab of Chrome Dev Tools). 

>The most you can do is add a small clarification in the text before this example, like "(note, we are using strict mode in this task)"
I disagree, the most you could do is update the code to have "use strict". Also, it requires fewer characters to fix and reduces cognitive load for the developer.

It should be noted that many developers new to Javascript might be Googling for a specific topic and be directed to a specific article in javascript.info. Hence, it may not be correct to assume that they'd go back to the article to read the line you highlighted.

As I've benefitted from javascript.info, this is a small attempt to make it better and save time for others. However, I'll leave it to the maintainers of this repo to decide the path forward for this PR.

Alexandre887:
For sure, it is up to the maintainer of the English version of the tutorial to accept your edit or not. I am just sharing my experience, as I am also a maintainer of one of the javascript.info translations. In practice, I have never yet come across an example from any of the articles that explicitly states "use strict" in the code. At most, as I mentioned, there are explanations like "we will use strict here". 

This is the writing style of a tutorial articles, if an example requires the use of strict mode to get a specific result, it is stated in the article - it seems quite logical to me. 

Regarding Googling, there are sometimes quite complex examples in articles that take 3-4 additional code blocks to explain. In this case, if "use strict" is required for a particular example to work correctly, based on your comment, we should add "use strict" to the beginning of each of the 5 code blocks? This is not reasonable. We could have added a small explanation before these examples, and would have focused the reader's attention on the explanation of the content, not on the repetitive "use strict" directive. 

Besides, if a developer already uses strict mode and decides to copy several code samples from the article at once, it will make copying a bit more complicated, because you will have to copy the code without the first line all the time.

shallow-beach:
Perhaps the `use strict` qualifier (just the short sentence) should be copied somewhere more prominent, like [the main page](https://javascript.info/)? Seems a common enough PR to address it.

PR #3544, opened 2023-08-14
Added clarification about absolute path

Alexandre887:


shallow-beach:
Could you give an example of a site operating condition where absolute paths are specified with respect to the root of the disk?

PR #3554, opened 2023-08-18
Update solution.md

omkarbate22:


Alexandre887 (commented):
We're talking about a variable `shoppingCart`:

```js
let shoppingCart = fruits;
```

Why are you changing `shoppingCart` to `shopping-cart` in the text? Besides, in JavaScript, dashes `-` **cannot** be used in variable names.

PR #3555, opened 2023-08-19
correct func parameter in solution

Sammeeey:
The task was to create a function `sortByAge(users)` (using `users` as parameter). Now also the solution reflects that.

### DIFF
--- a/1-js/05-data-types/05-array-methods/8-sort-objects/solution.md
+++ b/1-js/05-data-types/05-array-methods/8-sort-objects/solution.md
@@ -1,5 +1,5 @@
 ```js run no-beautify
-function sortByAge(arr) {
+function sortByAge(users) {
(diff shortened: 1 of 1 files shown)


1-js/05-data-types/05-array-methods/8-sort-objects/solution.md
```diff
@@ -1,5 +1,5 @@
 ```js run no-beautify
-function sortByAge(arr) {
+function sortByAge(users) {
   arr.sort((a, b) => a.age - b.age);
```
Alexandre887, line 3:
```suggestion
  users.sort((a, b) => a.age - b.age);
```

Alexandre887 (changes requested):
I'd also suggest to rename `arr` to `users` here:

image

PR #3568, opened 2023-08-28
Add space between "while" and parenthesis

salpelter:
Putting space between for/if/while and the parenthesis is recommended to avoid confusion with function calls.
javascript.info follows this recommendation, but I found and fixed an exception.


smith558:
@salpelter These are some good suggestions, we will merge them if you can open a new pull request now. 🙂

salpelter:
@smith558 the repository that I used to submit this PR was deleted at some point, can you please advise me what to do because I don't see a way to reopen this specific commit, I only know how to open a new one. same applies to this pr #3540

PR #3596, opened 2023-10-18
Fix: invalid url

inblossoms:
Fixed invalid url caused by redirection after the repository name was changed.


Alexandre887:
Duplicate of #3592

PR #3614, opened 2023-11-13
Unnecessary call to write function on SSE server.js

hamidjafari:
## Description
Remove unnecessary call to `write` function in SSE server.js as `setInterval` will start calling the function


joaquinelio:
No need to wait for the 1st 1000 ms

```
hey, give me a brick, every 15 secs
hey, starting now
```

PR #3615, opened 2023-11-17
Inaccuracy

Alexandre887:


smith558:
This may need further clarification to include even other cases such as:
```js
const method = { foo() {} }.foo;
const arrowFunction = () => {};
async function asyncFunction() {}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/prototype

PR #3617, opened 2023-11-24
Improve awkward sentence structure

smith558:
The original sentence was a bit awkward to understand and also not grammatically correct.

### DIFF
--- a/1-js/05-data-types/10-destructuring-assignment/article.md
+++ b/1-js/05-data-types/10-destructuring-assignment/article.md
@@ -40,7 +40,7 @@ alert(firstName); // John
 alert(surname);  // Smith
 ```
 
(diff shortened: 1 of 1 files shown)


joaquinelio:
👍 
Maybe the ugliest.

There are many pending grammar PR.
I don't agree with them all,  but I dont do the review because I m not a native English speaker.

joaquinelio:
Why 
43 Let's  Let’s
Careful, ’ is a formating char

smith558:
> Why 43 Let's Let’s Careful, ’ is a formating char

Yes, I just noticed that. For some reason, my autocorrect fixed that for me and I accidentally pushed it.

smith558:
> Why 43 Let's Let’s Careful, ’ is a formating char

Fixed.

PR #3622, opened 2023-12-08
Add new toSorted, toSpliced and toReversed methods to "Array methods" article

Hasan-Mir:
This pull request introduces additional notes to the documentation of Array methods. The aim is to inform readers about new methods that perform similar operations to `splice`, `sort`, and `reverse`, but do not mutate the original array. These methods are `toSpliced`, `toSorted`, and `toReversed`.

### DIFF
--- a/1-js/05-data-types/05-array-methods/article.md
+++ b/1-js/05-data-types/05-array-methods/article.md
@@ -96,6 +96,18 @@ arr.splice(2, 0, "complex", "language");
 
 alert( arr ); // "I", "study", "complex", "language", "JavaScript"
 ```
+````smart header="toSpliced"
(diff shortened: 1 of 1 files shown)


damonholden:
yeah lets get this updated. Also worth mentioning the `.with()` Array method aswell. Also some of the tasks like "copy and sort" in this section need to be updated to include the new apis.

smith558:
> Also worth mentioning the `.with()` Array method aswell. Also some of the tasks like "copy and sort" in this section need to be updated to include the new apis.

Would you be interested in opening a PR for this after this one gets merged (will be soon)? 🙂

damonholden:
> > Also worth mentioning the `.with()` Array method aswell. Also some of the tasks like "copy and sort" in this section need to be updated to include the new apis.
> 
> Would you be interested in opening a PR for this after this one gets merged (will be soon)? 🙂

Sounds good. Once merged, I'll work on the PR to get the other methods in and adjust the tasks as well. Will keep an eye out.

PR #3625, opened 2023-12-13
Update article.md

briandubya:
Removed paragraph texts that were indented within code blocks.  Added admonitions (note and warnings)


mohitgedar (commented):
- **change of line 63** is not necessary because it don't impact the original website . but its good that you did it 
- I think **changes from line 92-125** are unnecessary , I have read from the website, the warning and note should be in their separate blocks , which they are already.

PR #3628, opened 2023-12-22
Fix grammar and typos

nakhodkin:


smith558 (approved):

smith558:

joaquinelio:
about 
"Agree on terms within your team and in your own mind."
It's subtle.
If you erase "own", the implicit warning is no more.
The warning, people working alone tend to relax rules.
It is evident when reading the aricle, 
not so when correcting out of context sentences.

smith558:
> your own

"your team" implies, in fact, semantically requires that the team is his/her _own_. Therefore "your own" should not be necessary. ([source](https://ell.stackexchange.com/questions/33806/your-something-vs-your-own-something)) It is in _most_ cases used purely for emphasis. Do you feel it's necessary here?

In the case of "quick glance", "really know", "clearly communicate" and likewise. I would say in this article the modifiers in these phrases do not serve a clear purpose and mostly clutter the sentences. Let me know what you think! 🙂

joaquinelio:
It's ok the lint command,
but our languages don't require minifiers 😄 

I liked 
"Buy your food"
"Buy your own food" 
the last implies some reproval.
You could avoid "own", but if you say it aloud, you would say
"Buy YOUR food" with maybe an accusatory finger.
Expressiveness is the key.

I don't thing they clutter sentences.
Wrong grammar does.

joaquinelio:
Sorry
Too much debate for a style topic

I would say
focus on real grammar issues, there are many of them.

PR #3634, opened 2023-12-31
Fix grammar and typos

nakhodkin:


smith558 (approved):

smith558:
@nakhodkin Thank you for your contribution! 🙂

joaquinelio:
I don't review grammar because I'm not native speaker

But I think the fix on line 258 is incorrect, the comma is needed.

`If we want to check if `item` exists in the array, and don't need the index, then `arr.includes` is preferred.`

smith558:
> I don't review grammar because I'm not native speaker
> 
> But I think the fix on line 258 is incorrect, the comma is needed.
> 
> `If we want to check if `item`exists in the array, and don't need the index, then`arr.includes` is preferred.`

A comma before "and" is almost always optional. I belive here it's the case as well.

PR #3644, opened 2024-01-12
Updated RFC spec Safe Methods URL in Cookies chapter

alexandermirzoyan:
The old URL was just redirecting to the main RFC spec webpage, so the user was had to to find the needed Content from Table of Contents.

Within this change URL has the right section's ID.


smith558:

PR #3648, opened 2024-01-18
Automated testing,  advanced topic warn

joaquinelio:


joaquinelio:
Why:
Just read Disqus complaining comments !!!
.
Vets like/love the artcle,
but beginners feel very frustrated to find this article so early in the tutorial. 
So the warning. 
It's not that hard, I bet the warning will change the mood to "open mind"
I copied it from the last line.
.
Line 27, rewritten by chatgpt. Maybe can be dropped.
Hard to explain.
I think the artcle starts abstract and beginners fail to understand that they dont need to struggle hard with two codes at the same time.
Chatgpt:
_Starting with a more definitive statement can enhance clarity. (...) This provides a clear and concise explanation of automated testing._
.
Maybe redundant last line can now be reduced.

smith558:
@iliakan this may require an input from you

PR #3653, opened 2024-01-23
Update solution.md to accept arguments

AbhinasRegmi:
Original Code caused error if tried to pass args to debounce decorated function.


AbhinasRegmi:
My bad it works.

PR #3661, opened 2024-01-28
Updated the explanation about difference between pageX/Y and clientX/Y coordinates

Amine970:
I think pageX/Y change when the page is scrolled and clientX/Y don't change so I updated the article


smith558:

PR #3664, opened 2024-02-02
Update article so 'samesite=lax' is the default behavior

nepikn:
[MDN Set-Cookie: samesite=lax](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#lax)


smith558:

PR #3679, opened 2024-03-20
Remove the square brackets "[" and "]" from the syntax

nintendoit:
In the syntax for array splice method, the square bracket should not be present.

### DIFF
--- a/1-js/05-data-types/05-array-methods/article.md
+++ b/1-js/05-data-types/05-array-methods/article.md
@@ -41,7 +41,7 @@ The [arr.splice](mdn:js/Array/splice) method is a swiss army knife for arrays. I
 The syntax is:
 
 ```js
-arr.splice(start[, deleteCount, elem1, ..., elemN])
+arr.splice(start, deleteCount, elem1, ..., elemN)
 ```
 
 It modifies `arr` starting from the index `start`: removes `deleteCount` elements and then inserts `elem1, ..., elemN` at their place. Returns the array of removed elements.


nintendoit:
**Another question:**

for slice syntax -> `arr.slice([start], [end])`
Do square bracket serve any special purpose here?

joaquinelio:
> **Another question:**
> 
> for slice syntax -> `arr.slice([start], [end])` Do square bracket serve any special purpose here?

It's an old convention for "optional".

Mmm...
Should be:
`arr.slice( [start] [,end] )`

nintendoit:
Ok, then the splice syntax is correct.But it do look confusing.

Should I 
- close this PR 
or 
- revert that change and make change to the slice syntax?

smith558:
Hi all, it should actually be `arr.splice(start[, deleteCount, elem1, ..., elemN])`, as it is. All parameters optional except of the first one. It's a GNU syntax of method signatures for documentations. It's a bit obscure and not very beginner friendly, so I was already thinking whether we should adopt a different style such as the one MDN DOcs are using. Let's keep this open.

joaquinelio:
I like the brackets, the single line is more clear when you know the brackets purpose.
I suggested a font change just for the brackets. If you paint them gray and slightly bigger, your brain will recognize they are not part of the syntax.
Maybe the dots should be changed too.

But
Consider adopting  MDN style.
MDN has influence in the community, a single convention would be user friendly.

smith558:
It may be worth creating a separate article on this syntax so that people can be introduced to it, or at least provide a link. Unfortunately there isn't much resources on this online, there seems to be no standard. The only relevant discussion about this I can find is https://stackoverflow.com/questions/10925478/how-to-interpret-function-parameters-in-software-and-language-documentation.

joaquinelio:
I'd stay with brackets.
Imagine that stackoverflow link example in mdn style.

After "manual and specifications"
Title
"Syntax conventions"
Body
"Square brackets mean optional"
Odd.

PR #3681, opened 2024-03-31
Updated result visualization

Filin3:
## Description

Updated the visualization of the result from a number to a string


smith558:

PR #3696, opened 2024-05-29
Update article.md

mariaedwards:
There was a typo in the article around the recursion depth being 100k vs presumably intended 10k


joaquinelio:
From the article:
 10 000, "some eng allow more" 
100 000  "out of limit"

I think the article is ok as it is now

mariaedwards:
misread the out of limit part, sorry

PR #3697, opened 2024-05-31
Inappropriate code run button in the task description

Alexandre887:
In the “[Function in if](https://javascript.info/task/function-in-if)” task, there is a button for running the code, which allows to find out the answer to the task at once, even without going to the section with the solution.

image


FloffyGarlic:
The reader can already check the solution by just one button press: the "solution" button. Besides, seeing the engine actually return the solution described in the "solution" part can make the explanation feel proven. I don't think this is harmful.

PR #3699, opened 2024-06-05
Add a solution for that enables the direct use of atomic groups

slevithan:
The catastrophic backtracking solution described where atomic groups are emulated using lookahead and backreferences is useful but can be tricky to use and error prone (e.g. when quantifying the result, or in longer patterns that rely on multiple atomic groups). So this adds a link to an easy to use solution that enables the direct use of atomic groups via `(?>…)` in native JS regexes.


CyberAnilpatil (approved):

slevithan:
I updated the PR to make some additional tweaks:

- Atomic groups don't capture, so I changed "atomic capturing groups" to "atomic groups" (their correct name).
- The emulation trick (with a backreference to a capturing group inside a lookahead) emulates an atomic group, not a possessive quantifier, so I corrected this. This distinction is subtle since possessive quantifiers are sugar for some common uses of atomic groups, but it is important. For example, the lookahead/capture trick accurately emulates the behavior of an atomic group like `(?>…|…)` which is not followed by a quantifier at all.

slevithan:
Ping @Alexandre887 for review. This PR makes this (already quite informative) article even better for readers.

slevithan:
Hey @smith558, any problem with landing this PR as is?

Alexandre887:
Hey @slevithan, just want to make it clear that PRs are reviewed by volunteers here, so there is no strict timetable for when a particular PR will be merged or anything like that — it's directly dependent on the availability of free time and willingness to review a particular PR. So if there is no response for a long time, it doesn't mean that your PR has been forgotten about, or that it has some flaws (otherwise someone would have said you about it). Thanks for your patience.

PR #3700, opened 2024-06-08
change example element of multidimensional array

sneeed:
When teaching this subject I found that an example where you can see for which array the first and the second value in the [] after matrix stand, helps understanding better.

### DIFF
--- a/1-js/05-data-types/04-array/article.md
+++ b/1-js/05-data-types/04-array/article.md
@@ -426,7 +426,7 @@ let matrix = [
   [7, 8, 9]
 ];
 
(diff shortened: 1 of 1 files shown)


mohitgedar (approved):
## I agree with the statement that 
### "an example where you can see ,for which array the first and the second value in the [ ] after matrix stand for, helps understanding better." 
# The changes are also OK to go.

smith558 (approved):

smith558:
Thank you for your contribution! 🙂 @sneeed And the review! @mohitgedar

PR #3709, opened 2024-07-03
Remove BigInt IE incompatibility part

FloffyGarlic:
As Internet Explorer was retired years ago, this paragraph is not needed anymore, so I deleted that line.

### DIFF
--- a/1-js/02-first-steps/05-types/article.md
+++ b/1-js/02-first-steps/05-types/article.md
@@ -94,8 +94,6 @@ const bigInt = 1234567890123456789012345678901234567890n;
 
 As `BigInt` numbers are rarely needed, we don't cover them here, but devoted them a separate chapter <info:bigint>. Read it when you need such big numbers.
 
-You can check [*MDN* BigInt compatibility table](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#Browser_compatibility) to know which versions of a browser are supported.
-
 ## String
 
 A string in JavaScript must be surrounded by quotes.
--- a/1-js/02-first-steps/05-types/article.md
+++ b/1-js/02-first-steps/05-types/article.md
@@ -94,11 +94,6 @@ const bigInt = 1234567890123456789012345678901234567890n;
 
 As `BigInt` numbers are rarely needed, we don't cover them here, but devoted them a separate chapter <info:bigint>. Read it when you need such big numbers.
 
-
-```smart header="Compatibility issues"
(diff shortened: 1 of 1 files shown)


mohitgedar (commented):
Today BigInt is supported by edge , the new IE , but I thing the link to MDN will be useful in the content . I think deleting the whole part is not required , just modify it to give a reference to MDN to check compatibilites . and also remove that BigInt is not supported in edge , you understand what I am trying to say ?

joaquinelio:
The link, in this context (" You can check..."), is useless, it's all checked and IE is missing.
I would Remove it..

Should the ref to IE be removed?

Many references to IE compatibility exist in the tutorial. I once submitted a PR removing all of them, but I was told that IE is still used in corporate environments that have not yet migrated. 
Maybe is time, I don't know.

Also
the tutorial includes explanations for old and no longer recommended functions and practices because students may encounter old code using them
so maybe ie refs can stay.
Nah, in this particular case I would remove it too.

FloffyGarlic:
We could mention something along the lines of "Although virtually all browsers support BigInts today, they still count as a relatively new feature. You can check [insert MDN compatibility table] to check which version of each browser is compatible with BigInt". However, I still think it's redundant. Sure, knowing about the MDN reference is important, but since the type is compatible with virtually every browser, the MDN compatibility reference is not necessary anymore when it comes to BigInts themselves. Thus, it probably doesn't deserve a place in the BigInts paragraph. It's good to be concise.

FloffyGarlic:
> The link, in this context (" You can check..."), is useless, it's all checked and IE is missing. I would Remove it..
> 
> Should the ref to IE be removed?
> 
> Many references to IE compatibility exist in the tutorial. I once submitted a PR removing all of them, but I was told that IE is still used in corporate environments that have not yet migrated. Maybe is time, I don't know.
> 
> Also the tutorial includes explanations for old and no longer recommended functions and practices because students may encounter old code using them so maybe ie refs can stay. Nah, in this particular case I would remove it too.

May I ask for a *review* if you approve of this change?

joaquinelio (approved):
Agreed 

The link is waste of time, it shows all checked today

The IE incompatibility ref doesn't add anything.

joaquinelio:
>@joaquinelio May I ask for a review if you approve of this change?

@FloffyGarlic 
It's not matter of approval, the eepo is having little activity

About IE, even retired, the tutorial still shows it because of the old code (corporate code in this case) you may still encounter. Here I just dont think it is useful

smith558 (approved):

smith558:
Thank you for your contribution! 🙂 @FloffyGarlic And the review! @joaquinelio

PR #3711, opened 2024-07-04
Polyfill.io has been flagged as a server malware.

CSKishan:
Supply chain attack noticed in over 100,000 websites using polyfill.io.

Replaced with a safer alternative (polyfill-fastly.io)

### DIFF
--- a/1-js/03-code-quality/06-polyfills/article.md
+++ b/1-js/03-code-quality/06-polyfills/article.md
@@ -1,4 +1,3 @@
-
 # Polyfills and transpilers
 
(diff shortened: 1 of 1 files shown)


kricsleo:
Same here.

1-js/03-code-quality/06-polyfills/article.md
```diff
@@ -33,7 +32,7 @@ A transpiler would analyze our code and rewrite `height ?? 100` into `(height !=
 height = height ?? 100;
 
 // after running the transpiler
-height = (height !== undefined && height !== null) ? height : 100;
```
FloffyGarlic, line 36: Sure, these two lines do the exact same thing, but since the old one is much easier to understand, it's better. If some parentheses make code easier to read, they should stay. I don't approve of this change.

PR #3721, opened 2024-07-21
Clarify some things in Objects intro

TurekBot:
I noticed a few typos or unclear things while I was reading, so I figured I'd leave things better than I found them. I left an explanation for each change in its commit.

Any suggestions are welcome!

### DIFF
--- a/1-js/04-object-basics/01-object/article.md
+++ b/1-js/04-object-basics/01-object/article.md
@@ -7,7 +7,7 @@ In contrast, objects are used to store keyed collections of various data and mor
 
 An object can be created with figure brackets `{…}` with an optional list of *properties*. A property is a "key: value" pair, where `key` is a string (also called a "property name"), and `value` can be anything.
 
-We can imagine an object as a cabinet with signed files. Every piece of data is stored in its file by the key. It's easy to find a file by its name or add/remove a file.
+We can imagine an object as a cabinet with labeled files. Every piece of data is stored in its file by the key. It's easy to find a file by its name or add/remove a file.
 
 ![](object.svg)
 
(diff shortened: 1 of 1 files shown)


1-js/04-object-basics/01-object/article.md
```diff
@@ -161,7 +161,7 @@ alert( user.key ) // undefined
 
 ### Computed properties
 
-We can use square brackets in an object literal, when creating an object. That's called *computed properties*.
+If you want to be flexible about naming your properties, you can use something called *computed properties*.
```
shallow-beach, line 164: disagree, Line 164 is clarifying that *square brackets*  are computed properties

1-js/04-object-basics/01-object/article.md
```diff
@@ -101,7 +101,7 @@ For multiword properties, the dot access doesn't work:
 user.likes birds = true
 ```
 
-JavaScript doesn't understand that. It thinks that we address `user.likes`, and then gives a syntax error when comes across unexpected `birds`.
+JavaScript doesn't understand that. It thinks that we're referring to `user.likes`, and then gives a syntax error when it comes across the unexpected `birds`.
```
shallow-beach, line 104: How about 'accessing' (used in [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Property_accessors)):

>JavaScript doesn't understand that. It thinks that we're accessing `user.likes`, and then gives a syntax error when it comes across the unexpected `birds`.

1-js/04-object-basics/01-object/article.md
```diff
@@ -7,7 +7,7 @@ In contrast, objects are used to store keyed collections of various data and mor
 
 An object can be created with figure brackets `{…}` with an optional list of *properties*. A property is a "key: value" pair, where `key` is a string (also called a "property name"), and `value` can be anything.
 
-We can imagine an object as a cabinet with signed files. Every piece of data is stored in its file by the key. It's easy to find a file by its name or add/remove a file.
+We can imagine an object as a cabinet with labeled folders. Every piece of data is stored in its folder by the key. It's easy to find a folder by its name or add/remove a folder.
```
shallow-beach, line 10: file is clear IMO, see [File cabinet wiki](https://en.wikipedia.org/wiki/Filing_cabinet)

shallow-beach (commented):
a few thoughts!

PR #3722, opened 2024-07-21
Improving "Object references and copying" article

TurekBot:
I noticed a few things that could be better, so I went ahead and fixed them. Explanations for each change are the corresponding commit message.

Suggestions are welcome!


joaquinelio (commented):
Hyphens "--" are there on purpose, is metadata, result in a dash " – " on the view

Grammar
I am not a native english speaker.
- I thought it was 
"then (we do) replicate..."
"then (the code) replicateS" sound better o me.
- "and then, if..." sounds better too, about where the action is. 
Again, a native should review.

Minor changes look ok
Not a fan when they are not strictly errors, as they pile up and become a pain for translated repos.

PR #3723, opened 2024-07-21
Improving "Garbage collection" article

TurekBot:
I noticed a thing or two that could be better, so I went ahead and fixed them myself. 

For changes that seem less obvious, I explain in the commit message. However, maybe something that was obvious to me is not obvious to you, so feel free to ask for further explanation.

Suggestions are welcome!


joaquinelio (commented):
The tutorial have many grammar issues.
I'd focus there, actual errors
Still,
There are many good PR piling up...
 
Lack of reviewers

...

Backticks are there on purpose. They format the view (usually reserved words),resuling in a highlighted text, a quoted "user" in this case.

Why not dashes?
They are perfect. Allowed and clearer.
I love them so much that I leave them even in Spanish where they are not allowed. Take that, Real Academia Española.

"anymore", maybe, I dont know if mandatory

"the"  and "-s" added,
some simplifications,
Look like real fixes.

PR #3731, opened 2024-07-30
Update article.md

shallow-beach:
specify article subject


shallow-beach:
not a satisfactory clarification, so closing

PR #3734, opened 2024-08-01
Update article.md

shallow-beach:
clarify jump behavior


shallow-beach:
doesn't conform to article chaining style

shallow-beach:
fixed in #3736

PR #3736, opened 2024-08-01
Update article.md

shallow-beach:
clarify jump behavior

### DIFF
--- a/1-js/11-async/04-promise-error-handling/article.md
+++ b/1-js/11-async/04-promise-error-handling/article.md
@@ -51,6 +51,8 @@ new Promise((resolve, reject) => {
 *!*
   throw new Error("Whoops!");
 */!*
+}).then(() => {
+  alert('jump') // line skipped upon error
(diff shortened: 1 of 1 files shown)


iliakan:
It it unclear without it? Instead of 'jump', it'd be better to put something like "doesn't execute" or "jumped over" perhaps.

shallow-beach:
> It it unclear without it? Instead of 'jump', it'd be better to put something like "doesn't execute" or "jumped over" perhaps.

I am a newbie, and personally I needed to confirm jump behavior in a sandbox so I thought I might include it earlier (although technically it is clarified already in the URIError example in the 'Rethrowing' section). I'm not too picky about language :)

shallow-beach:
> It it unclear without it? Instead of 'jump', it'd be better to put something like "doesn't execute" or "jumped over" perhaps.

[resolution commit](./3736/commits/bfc8a7f4777e5ebb3fa69ae354199575229b37ad)

PR #3740, opened 2024-08-04
Update article.md

shallow-beach:
mention array-like Symbol, discussed previously, for comparison


shallow-beach:
merged pr #3739

PR #3742, opened 2024-08-05
Update recursion/article.md

shallow-beach:
use JS syntax rather than C-pointer-esque notation

I think many beginners would try and interpret `next -> list` as some convoluted arrow function.


shallow-beach:
merged into pr #3741

PR #3760, opened 2024-09-14
Update article.md

mann-codes2014:
Add **let** before **key** in **for (key in object)** in section https://javascript.info/object#forin


joaquinelio:
I dont know...

Not part of the sintax, I consider this sintax description
`for (key in object) {`
The examples below that have the strongly suggested let.

Key can even be declared lines before tthis,
but maybe Im overthinking this.

PR #3761, opened 2024-09-19
Convert list of polyfills into a sentence

dlarocque:
polyfills.io was removed a few months ago and now only core-js is listed, so this should be converted from a list to a single sentence.


dlarocque:
I will not sign the CLA, so I'm closing this PR and just going to report it as an issue instead.

PR #3770, opened 2024-10-15
Update article.md | changes under Comma subheading

errwnd:
Shouldn't "a"  be assigned the value of the later operation(3+4) rather than (1+2).


joaquinelio (commented):
No
 Read it, 
precedence is the point of the smart header

let a = (1 + 2, 3 + 4)    // a = (7)

let a =  1 + 2, 3 + 4     // (a = 3), 7

errwnd:
`let a = 1 + 2, 3 + 4 // (a = 3), 7`
Shall we have a code representation like:
`let a;
a =1+2, 3+4; 
alert(a); // returns 3
`

PR #3808, opened 2025-01-28
Still a Promise

joaquinelio:
Hello @iliakan 

This may be seen as a trivial task at this point.
And the point I want to make is explained later in a Fetch task.

But a couple of times I found myself trying to defend the power of async to beginers and even the sync-veterans that love the beauty of result = dbSqlSelect() 

they see await as a relief of pain
that's why I put it as a await task, and its name   

You may do whatever you want with it, put it in another place, delete sillly jokes and comments, drop it entirely and make fun of it.
It was fun to write.


joaquinelio:
>One will use .any, the other one .all ?

easy, done.
image


But if I finish the killing of the horses in favor of fetch, it'll be a new branch.
the fetch sketch branch looks itchy. What should I fetch() ?

iliakan:
Sorry for the delay, I just saw that the task needs a more serous consideration.

Surely we don't need duplicate tasks.

PR #3811, opened 2025-02-04
fix(json): fix invalid JSON syntax

aayushsingh7:
- (value[, replacer, space]) → (value, [replacer, space])
- (str[, reviver]) → (str, [reviver])


smith558:

PR #3813, opened 2025-02-10
Replacing line163 of article.md with a more accurate line

Jaskiranjot-Kaur:


joaquinelio:
I don't think that's true.
They both are technical limitations,
not rules to follow.

Watch the redundancy too.

smith558:
I indeed think the original phrasing is better here too. @Jaskiranjot-Kaur

smith558:
Anyway, thank you for your contribution! 🙂

PR #3820, opened 2025-02-22
The answer is out there

joaquinelio:
line 469 is as smart question,
needs to be answered

we dont manage everything oourselves 
like we dont put x y window coordinates on the screen anymore


joaquinelio:
I didnt chage the last line, 
I don0t know why it's marked

PR #3825, opened 2025-03-19
Update indexeddb/article.md - add warning about code rollback

catnipan:


joaquinelio (commented):
Not sure about this.

I think it's already implicit in the warn header (preceding line)

The new warn is a general version management problem and proper testing shoud catch it,
the original explains a really hard to catch issue you shoud be aware of

PR #3827, opened 2025-03-24
fetch() for local resources

joaquinelio:
I coudnt figure out a short explanation why sometimes (like, instead of shorter code with "use" for svg)  is better without adding too much text

At least it  informs local fetchs is allowed and can be handy


joaquinelio:
I asked chatGPT
it suggested: if short, shorter:

* … or even load local assets like SVGs, HTML fragments, or configurations dynamically

after 
* Submit an order,
* Load user information,
* Receive latest updates from the server,

mm...  maybe

PR #3831, opened 2025-04-01
Fix misunderstanding about for..of with plain objects in iterables article

krakowdeveloper:
Description:
This PR corrects a fundamental misunderstanding in the article about how for..of works with plain objects. The current explanation could lead readers to believe that:

Plain objects work with for..of by default (they don't - they must implement Symbol.iterator)

The iteration protocol is unclear before introducing Symbol.iterator

- [ ]  Improvements:

Clarifies that plain objects are NOT iterable by default
Better explains the relationship between objects, iterables, and iterators
Provides accurate examples showing the TypeError that occurs without proper implementation
Maintains all existing correct information while fixing the misconceptions

Before diving into implementation details, we now properly establish what doesn't work and why.

### DIFF
--- a/1-js/05-data-types/06-iterable/article.md
+++ b/1-js/05-data-types/06-iterable/article.md
@@ -5,16 +5,14 @@
 
 Of course, Arrays are iterable. But there are many other built-in objects, that are iterable as well. For instance, strings are also iterable.
 
-If an object isn't technically an array, but represents a collection (list, set) of something, then `for..of` is a great syntax to loop over it, so let's see how to make it work.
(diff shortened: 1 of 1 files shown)


krakowdeveloper:
how long takes a review?

iliakan:
I like the previous version more. It's differently phrased.

Maybe I'm biased toward my own text? :/ 
Open to suggestions.

joaquinelio:
IMHO 
I like the flow of the original 
and I dont see it suggests it's iterable by default

but there may be a point 
I would accept the change in line 8, it dissipates any doubt

PR #2495, opened 2021-02-04
needed 4 translated repos

joaquinelio:

PR #3856, opened 2025-06-19
explicit info link  loading-a-script

joaquinelio:
needed for translated repos to work


joaquinelio:
same issue that https://github.com/javascript-tutorial/en.javascript.info/pull/2495 fixed

and 
image

PR #3863, opened 2025-07-17
Fix minor typo

wjw020206:
Changed all incorrect uses of capital `O` in `instanceof` in the documentation to lowercase `o`.


joaquinelio (approved):
right!
it's an operator

PR #3869, opened 2025-08-22
Fix: Javascript's earlier names

vishal0902:
Initially, Javscript was called "Mocha", then changed to "LiveScript" before finally being named JavaScript.


vishal0902:
Yes you said it right because the focus is on "why JavaScript".

Actually I read about this on Wikipedia when I was going through an article
about Brenda Rich, the creator of JavaScript. I read this in career
section. Here's the link -
https://en.m.wikipedia.org/wiki/Brendan_Eich

By the way, javascript.info is the best resource I have ever found. Great
work.

vishal0902:
Yes you said it right because the focus is on "why JavaScript".

Actually I read about this on Wikipedia when I was going through an article
about Brendan Eich, the creator of JavaScript. I read this in career
section. Here's the link -
https://en.m.wikipedia.org/wiki/Brendan_Eich

By the way, javascript.info is the best resource I have ever found. Great
work.

PR #3901, opened 2025-11-08
Corrected hotkey for code comments on Mac

Lk-ux:


smith558:
@Lk-ux please check some of the comments

PR #3902, opened 2025-11-09
clarify wording about null and undefined equality in JS

Lk-ux:
Updated the explanation of JavaScript’s loose equality to accurately state that null and undefined are equal to themselves and to each other (==), but not equal to any other value. The previous wording was ambiguous about whether “each other” included self-equality.


smith558:

PR #3915, opened 2025-12-14
fix: update private-protected-properties-methods article.md en

front42:
### Private and protected properties and methods
Let's remove outdated information —
private properties and methods are part of the ECMAScript standard✅


front42:
For some reason, the en-version hasn't contribution tips with "pre-made" banners mention, but the ru-version [has](https://github.com/javascript-tutorial/ru.javascript.info/blob/master/CONTRIBUTING.md#как-добавить-блок-новая-возможность),
perhaps they're stored somewhere. `[recent browser="new"]` appears here five times, `[recent browser=Chrome]` [once](https://javascript.info/static-properties-methods#static-properties),
`[recent browser=none]` once too, so maybe shortcodе-marker links similar preformatted text, `none` hints not supported.

PR #3917, opened 2025-12-19
fix: correct definition of spread syntax in generators article.md en

front42:
### Generators
Spread syntax does not turn the iterable object into an array of items,
but in this article example, when used in a literal array declaration,
it adds the expanded list of generator values ​​to that array✅


joaquinelio (approved):
I agree 👍

PR #3939, opened 2026-02-22
fix: correct cross-window-communication article.md en

front42:
### Cross-window communication

While [editing](https://github.com/javascript-tutorial/ru.javascript.info/pull/2123) the ru-version, it turned out that it's more natural in some respects.
It avoids the word `tag` – DOM entities are involved here,
sometimes it's more accurately stated that we're talking about objects,
the possibility of [getting](https://learn.javascript.ru/cross-window-communication#dostup-k-soderzhimomu-ifreyma) `error` or `null` with `another origin` is indicated, and so on.
But both couldn't avoid confusing newcomers' minds by misunderstanding the terms [event](https://developer.mozilla.org/en-US/docs/Web/API/Window/load_event) and its handler in [DOM property](https://javascript.info/introduction-browser-events#dom-property)
or providing unconfirmed information about [short syntax](https://developer.mozilla.org/en-US/docs/Web/API/Window/message_event#syntax) not working, etc.
So let's try to combine the best of each of these versions✅

### DIFF
--- a/3-frames-and-windows/03-cross-window-communication/article.md
+++ b/3-frames-and-windows/03-cross-window-communication/article.md
@@ -51,7 +51,7 @@ For instance, let's try reading and writing to iframe from another origin:
     try {
       // ...but not to the document inside it
 *!*
-      let doc = iframe.contentDocument; // ERROR
+      let doc = iframe.contentDocument; // ...getting null
 */!*
     } catch(e) {
       alert(e); // Security Error (another origin)
--- a/3-frames-and-windows/03-cross-window-communication/article.md
+++ b/3-frames-and-windows/03-cross-window-communication/article.md
@@ -23,21 +23,21 @@ These ones do not:
 
 The "Same Origin" policy states that:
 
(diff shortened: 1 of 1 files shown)


3-frames-and-windows/03-cross-window-communication/article.md
```diff
@@ -23,21 +23,21 @@ These ones do not:
 
 The "Same Origin" policy states that:
 
-- if we have a reference to another window, e.g. a popup created by `window.open` or a window inside `<iframe>`, and that window comes from the same origin, then we have full access to that window.
```
ramgudimetla07-sketch, line 26: If we have a reference to another window (for example, a popup created using window.open() or a window inside an <iframe>), and that window belongs to the same origin, then we have full access to it.

Otherwise, if the window belongs to a different origin, we cannot access its contents such as variables, the DOM, or the document object. The only exception is the location object: we can change it (which allows us to redirect the user), but we cannot read it. This restriction prevents information leaks about the user’s current page or data.

ramgudimetla07-sketch (approved):

Issue #3943, opened 2026-03-07
In chapter Array mehtod's task section there is a typo which needs to be fixed


sajwan05:
In chapter [Array Methods](https://javascript.info/array-methods) 's section [tasks](https://javascript.info/array-methods#tasks) 's first exercise `camelize string` has a sentence `That is: removes all dashes, each word after dash becomes uppercased.`  

This sentence should be `That is: removes all dashes, each word after dash becomes capitalized.

Capitalized instead of uppercased.

[x] I would love to work on this issue.

<img width="852" height="493" alt="Image" src="https://github.com/user-attachments/assets/d7d155fa-e17b-4794-a152-85da86715b69" />

joaquinelio:
Good catch.

Feel free to open a PR if you'd like.
The maintainer merges PRs from time to time.

sajwan05:
Thanks @joaquinelio on it

PR #3948, opened 2026-03-22
typo- uppercased changed to capitalize

sajwan05:
In chapter [Array Methods](https://javascript.info/array-methods) 's section [Task](https://javascript.info/array-methods#tasks)'s first task `camelize` had a typo instead of `capitalize` it was using uppercase, so it is fixed with this PR.

**Issue** : #3943


joaquinelio (changes requested):
basically correct

you forgot a d 
capitalized

Issue #3940, opened 2026-02-28
[Typo]- In chapter Arrays in the section of loops we have to change `number` to `index`


sajwan05:
In chapter  [Arrays](https://javascript.info/array) 's section [Loops](https://javascript.info/array#loops) there in sentence `The for..of doesn’t give access to the **number** of the current element, just its value, but in most cases that’s enough. And it’s shorter.`  
Instead of **number** we have to use **index**. 
Because using number is not really telling us we're talking about index or number (which can be value)

<img width="1786" height="665" alt="Image" src="https://github.com/user-attachments/assets/deb442fa-3c33-40fd-b432-e7261488bbab" />

[X] I would love to work on this issue.

PR #3949, opened 2026-03-22
[Typo]- In chapter Arrays in the section of loops we have to changed number to index

sajwan05:
In chapter [Arrays](https://javascript.info/array) in the section [Loops](https://javascript.info/array#loops) we have `for...of` loop and there we changed number to index for more accurate language.

**Issue**: #3940


joaquinelio:
It looks like this branch was created from another branch, so it includes those commits as well.

No big deal.

The index fix is 👍
The "capitalized" fix still lacks the final d.

sajwan05:
Thanks for the suggestion but 
I think capitalize is right word here instead of capitalized.
Because capitalize is base form of verb, which can be used for some action which can happen now or in the future.
But if you still thinks "d" needs to be added let me know,  I will make the changes.

joaquinelio:
oh, I see.

I'm not a native English speaker, so I asked chatgpt :

«
I see what you meant: using capitalize as a sort of label, like saying “in uppercase.”
The problem is that English doesn’t use capitalize that way. If we want the neutral/label form, the idiomatic option is still capitalized.
So even though your intention makes sense, capitalized is the correct and natural choice here.
»

sajwan05:
> « I see what you meant: using capitalize as a sort of label, like saying “in uppercase.” The problem is that English doesn’t use capitalize that way. If we want the neutral/label form, the idiomatic option is still capitalized. So even though your intention makes sense, capitalized is the correct and natural choice here. »

Oh I see, thanks for this, I have changed capitalize to capitalized.

sajwan05:
@joaquinelio Hey I did the requested changes. Thanks.

joaquinelio (approved):
👍 

They are not cosmetic changes

PR #3965, opened 2026-05-22
Handle promise rejection in demoGithubUser call

AlexNT-maker:
Added a .catch(alert) safety net to the demoGithubUser() call to prevent an unhandled promise rejection in case the thrown error is not an HttpError.


smith558:
@codex review

PR #3972, opened 2026-06-19
Corrected the reasoning in the 'Step into' command explanation and made the `Making bound methods with class fields` section easier to understand.

singhal-shagun:
This pull request make the following improvements in the literature:

1. Corrected the reasoning as to why beginner-level Javascript learners can ignore the `3.1 Debugging in the browser --> “Step into”, hotkey F11` writeup.
2. The `9.1 Class basic syntax --> Making bound methods with class fields` writeup has been made easier to understand.

### DIFF
--- a/1-js/03-code-quality/01-debugging-chrome/article.md
+++ b/1-js/03-code-quality/01-debugging-chrome/article.md
@@ -143,7 +143,7 @@ There are buttons for it at the top of the right panel. Let's engage them.
     That's good if we're not interested to see what happens inside the function call.
 
 <span class="devtools" style="background-position:-4px -194px"></span> -- "Step into", hotkey `key:F11`.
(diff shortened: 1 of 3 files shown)


smith558:
@codex review

singhal-shagun:
Codex seems to have skipped the method vs function distinction in my writeup!

To clarify things, I clearly wrote the following in my writeup:
```
Making methods bound to object instances using class fields is a third solution. Such methods, called "bound methods" can be created with a very elegant syntax,
```
The intention is to convey to the reader that this third solution is applicable only to methods, not to functions in general.

Also, the original wirteup made no clarification as to what it meant by "bound methods". My writeup makes it explicit, making it easier for the reader to understand what the article intends to convey.

PR #3977, opened 2026-07-04
Convert prompt inputs to numbers with unary plus

Lk-ux:
The task description says the function should take natural **numbers** only, but prompt() returns String, which we pass to pow().
It accidentally works because we use *= operator which performs implicit conversion.
But it is best to convert explicitly to improve code readability and align better with the task requirement.


smith558:
@codex review

smith558 (approved):

smith558:

